PR #10, opened 2015-09-15
Added the anagram exercise.

kdar:
Added all the tests except for the unicode ones. I could add unicode support, but I'd probably have to include a small utf8 library that people could use.


kytrinyx:
I think it's fine to leave off the unicode support. If it seems like it's needed later, we can add it then.

Issue #4, opened 2014-10-20
Using rake to build C is extremely weird


LegalizeAdulthood:
My recommendation is to use CMake similar to what is done in the C++ track.

Using rake to build C code is just... weird.  I doubt anyone that is well steeped in C is going to be familiar with rake.

lpil:
:+1:

kytrinyx:
I didn't realize that we were doing that. I wholeheartedly agree that we should use CMake (or plain make) for this.

sunzenshen:
Don't mind either way. I chose Ceedling because it is one of the unit test frameworks recommended in *Test Driven Development for Embedded C*, and I wasn't sure if Boost was overkill for unit testing C code.

lpil:
You can use Unity (ceedling's test framework) without ceedling project management. :)

LegalizeAdulthood:
[cgreen](http://cgreen.sourceforge.net/) is a unit test framework designed with the needs of C in mind.  It does mocking via link-time substitution.  However, the framework may not work properly on Windows.  The last time I checked on it a couple years ago, it assumed unix and therefore had problems functioning properly on Windows.

I don't think you want to use a C++ oriented unit-test framework for testing C code.  I don't think it's reasonable to assume that C programmers know C++.  If they did, they wouldn't use C anymore :smile_cat:.

lpil:
> "Exception names should end with a trailing 'Error'."

Unity is ANSI C, so works everywhere. :)

> I don't think you want to use a C++ oriented unit-test framework for testing C code.

Unity is pure C :)

silentbicycle:
There's also [greatest](https://github.com/silentbicycle/greatest), which doesn't depend on anything and is actively un-opinionated about development style.

ryanplusplus:
It looks like everything is using `make` at this point.

PR #11, opened 2015-09-20
Hamming: Switch from Rake to Make

lpil:
Hello!

I've had a go at switching to `make` (#4), and was wondering if this looked like what we were after. If so, I'll have a go at the others.

Also, we've got these `unity.*` deps which will be common to all tests. Is there a way to avoid duplicating this `tests/vendor` dir?

Thanks,
Louis


lpil:
Thanks!

kytrinyx:
At the moment we don't have a way to distribute dependencies that are shared for each problem.

PR #18, opened 2015-12-07
Normalize YAML in Travis CI config

kytrinyx:
The YAML was not quite valid. Travis can read it
and wasn't complaining, but this ensures that the YAML
is clean.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
+---
 language: bash
-
 script:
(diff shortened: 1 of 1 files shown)


kytrinyx:
Looking at this a second time, the YAML was valid, but I ran it through a script which normalized it and I'm trying to make all the travis scripts consistent across all the language tracks.

PR #21, opened 2016-01-21
Made exercises directory and moved existing exercises

McFadge:
As per #20, moved existing exercises into an exercises directory


kytrinyx:
Thanks!

PR #23, opened 2016-01-30
leap: Corrected where test says it is running from

McFadge:
Very minor change so that while running the tests the correct test file is named.


kytrinyx:
Thanks!

PR #25, opened 2016-02-14
Add tests for buffer overflows

siebenschlaefer:
In exercises where the solution is written to a buffer, there should be a test for buffer overflows. Buffer overflows are a a common error source in C so we should help our users to avoid them.


kytrinyx:
Thanks!

PR #26, opened 2016-02-14
Explicitely state when functions don't take args.

siebenschlaefer:
With -Werror=strict-prototypes functions without arguments must be
declared as "return_type function_name(void)". -Werror=strict-prototypes
is in the CFLAGS variable of the make files, though unused.


kytrinyx:
Thanks!

PR #30, opened 2016-05-11
Update `config.json` with checklist_issue

joeltaylor:


fernandezpablo85:
What's the effect of this?

joeltaylor:
Whoops, sorry. I really should've given some context, but was opening ~13 of the same issues.

This is in reference to https://github.com/exercism/x-api/issues/109

PR #34, opened 2016-06-29
Recompile tests on any change to tests or code.

schmonz:
While practicing Hello World and Anagram, as I figured out what I was doing, I wanted to comment or uncomment tests. I expected that doing so would exclude or include those tests at runtime, but it didn't. Same problem as I was editing my solution. This commit tells `make` (for all exercises) that the compiled tests have dependencies on files in `src` and `test`, so that the tests get rebuilt when I expect.


kytrinyx:
Nice, thank you!

PR #36, opened 2016-08-08
99 Bottles of Beer on the Wall in c

divyaantonyjr:


kytrinyx:
Thanks @divyaantonyjr!

In order to merge this we need a couple of tweaks:

The directory name for 99 bottles of beer needs to be `beer-song`.

Check out how we set things up for the hello-world exercise:

```
exercises/hello-world/
├── makefile
├── src
│   └── example.h
└── test
    ├── test_hello_world.c
    └── vendor
        ├── unity.c
        ├── unity.h
        └── unity_internals.h
```

So to follow this structure for beer-song, you would need to

- create three directories: `src/`, `test/`, and `vendor/`
- rename the example solution to `src/example.h`
- rename the test file to `test/test_beer_song.c`
- copy the vendor directory from `hello-world` (we're working on a fix so that we don't have to include the vendor directory each time, but for now we need that)
- add a makefile that will run the test

Lastly, you'll need to add 'beer-song' to the config.json problems array (just stick it at the end of the array, I think).

@siebenschlaefer Would you mind reviewing the code for this one?

divyaantonyjr:
No problem

divyaantonyjr:
is this okay , it passed the travis-ci , But please check that the previous exercises are not disturbed before merging.... ,, It also compiles well in my pc (Arch Linux 64bit , gcc compiler)

divyaantonyjr:
Sorry commit d38fef319c7f87a84870906a9c6afded6b02db35 is exercise for binary to decimal conversion without using inbuit libs for conversion(first principles), i forgot to create a new WIP , so if you want you can use the code , i used inbuilt function only for stdout(printf) , if you think that must be avoided then i will use some inline assembly code , compiled and all test are passed(Arch Linux 64bit) ,

patricksjackson:
Hi @divyaantonyjr 

Don't submit two exercises in the same Pull Request. If you want to work on two at the same time you need to create different branches in your fork. The [contributing guide](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#branches) is a good place to start for learning how to do that.

I'll comment on your `beer-song` PR, as that's what this PR is for :)

1. The directory should be `beer-song` instead of `Beer-song`.
2. In `test_beer_song.c` you didn't actually write any tests. You have to check that the function is actually doing what you expect it to do, so that will require returning the string (instead of printing it) and comparing it to the expected output. Check out how the `bob` exercise does that using `TEST_ASSERT_EQUAL_STRING` -> [here](https://github.com/exercism/xc/blob/master/exercises/bob/test/test_bob.c)
3. I'd also strongly suggest that you look at some of the other implementations of this exercise to base yours off of. If you use [the cpp version of beer-song](https://github.com/exercism/xcpp/tree/master/beer-song), you can copy the format, tests, and some of the code for the solution. Doing the exercise broken up into multiple functions, `verse()` and `sing()`, makes it easier to test.

Issue #40, opened 2016-08-25
Choose, document, and use a style guide for provided code


ryanplusplus:
Currently there is some inconsistency in formatting/style for the tests and sample solutions. We should choose a style guide, document it in README.md, and ensure that it is applied consistently on the track.

ryanplusplus:
To get things rolling, how about the [Linux kernel style](https://www.kernel.org/doc/Documentation/CodingStyle) but with a more reasonable indent (3 spaces instead of 8)?

In other words, something like this:

```c
static int blah(int a)
   for (i = 0; i < num_elements; i++) {
      foo[i] = foo[i] + 42;
  
      if (foo[i] < 35) {
         printf ("Foo!");
         foo[i]--;
      } else {
         printf ("Bar!");
         foo[i]++;
      }
   }

   return a;
}
```

This is widely used, already documented, and can be auto-formatted using [GNU `indent`](https://www.gnu.org/software/indent/manual/indent.html) on any reasonable OS with the following settings:

```
-nbad -bap -nbc -bbo -hnl -br -brs -c33 -cd33 -ncdb -ce -ci4 
-cli0 -d0 -di1 -nfc1 -i8 -ip0 -l80 -lp -npcs -nprs -npsl -sai
-saf -saw -ncs -nsc -sob -nfca -cp33 -ss -ts3 -il1
```

We could even integrate a formatting check into our CI build to help us maintain consistent style.

patricksjackson:
I don't think you'll ever get 100% agreement on any coding style, but that looks like a sane set of choices to me, and the autoformatting idea is great :+1: 

I think you have the wrong settings listed for indent though. As far as I can tell you need to remove `-i8` and instead use `-i3 -nut` to get an indentation level of 3 without tabs.

ryanplusplus:
Thanks @patricksjackson. I've updated my post above and will test this out once I'm on my development machine.

palisaide:
I agree that the ability to use a simple program like `indent` is key to enforcing a formatting standard.  At least, it would immediately answer the question "is my code formatted correctly?"  Something that I have often asked myself on other language tracks.

ryanplusplus:
This should also include a preference for C99 features (https://github.com/exercism/xc/issues/53)

ryanplusplus:
This should also include a preference for fixed-width types (https://github.com/exercism/xc/issues/52)

patricksjackson:
[...]

PR #44, opened 2016-08-25
Remove explicitly passed string length in Hamming

### LINKED ISSUE #41: Hamming - use strlen instead of explicitly passing the length of each string
Especially since the tests pass the length as a magic number.

ryanplusplus:
I cleaned up the `compute` interface by removing explicitly passed (and error-prone) string length and calculated with `strlen` instead.

Closes #41

### DIFF
--- a/exercises/hamming/src/example.c
+++ b/exercises/hamming/src/example.c
@@ -2,12 +2,12 @@
 #include <string.h>
 #include "hamming.h"
 
-int compute(const char *lhs, const char *rhs)
+size_t compute(const char *lhs, const char *rhs)
 {
    size_t lhs_len = strlen(lhs);
    size_t rhs_len = strlen(rhs);
-   int count = 0;
-   int i = 0;
+   size_t count = 0;
+   size_t i = 0;
    for (i = 0; i < lhs_len && i < rhs_len; ++i) {
       if (lhs[i] != rhs[i]) {
          ++count;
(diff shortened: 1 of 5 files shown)


ryanplusplus:
I also took this opportunity to make the formatting more consistent with the provided example. If we're unhappy with this direction we can address it in #40.

palisaide:
Looks good.

krbullock:
It should probably have some concept of a max string length that it will handle, and use `strnlen()` to find the length in the reference solution instead.

ryanplusplus:
@krbullock do you mean to use that to prevent a denial of service attack with an incredibly long string? I can see how that would help someone learn about `strnlen`, but I'm not really seeing the utility in setting an arbitrary limit to the strings that can be compared.

krbullock:
Yeah, my thought is we should introduce people to _safe_ string handling in C from the beginning. Arbitrary length limits suck but are the reality in real C code in my (somewhat limited) experience.

Alternatively, maybe the caller could pass a single max length for both strings? Any safe code that would call `compute` would have to know the max length of the strings it was passing, whether they were static or user-provided.

ryanplusplus:
@krbullock the issue I have is that it's weird for this library to enforce length limits on strings that a _client_ provides. While it's a reality in C that you have to deal with arbitrary length limits, if the client is managing the memory then the client (and not the library) needs to deal with any weird length limitations.

palisaide:
I'm good with these changes.  Can we resolve conflicts and merge?

ryanplusplus:
Conflicts resolved and updated style applied.

PR #46, opened 2016-08-25
Use NULL instead of empty string when no name is provided

### LINKED ISSUE #33: [hello-world] name[0] == '\0' vs. name == NULL
Currently the tests as well as the example is implemented as checking the first character of `name` beeing `'\0'`, which makes it impossible to produce the (weird) output `"Hello, !"`.

Even if this output is weird it is valid in many other tracks, and since we are able to in C, I do also think we should resemble a behaviour which comes closer to most other tracks, and actually allow `hello(dest, size, "")` to fill `dest` with that weird string from above while `hello(dest, size, NULL)` will insert `"World"` as name.

Using `NULL` also does come close to other tracks that use some `Maybe` or `Option` type to achieve the behaviour described.

### LINKED ISSUE #43: Should we provide the solution's header and empty source file?
Including a (more or less) completed header will help set to ensure that solutions are const-correct and help to communicate all assumptions made about arguments in the tests.

Including an empty source file will help to ensure that the user can get _something_ compiling relatively quickly.

### LINKED ISSUE #56: Hello World -- limited buffer size seems like too much for an introduction exercise
The hello world exercise tests that the implementation is careful not to write out of bounds. While this is good practice, it seems excessive for a hello world. This may be someone's first exposure to C, to unit testing, and to Exercism, so it seems that we should save buffer overflows for a later topic.

I'd like to update the hello world exercise to use a comically over-sized buffer and save this topic for a later exercise. Thoughts?

ryanplusplus:
Closes #33 
Closes #43 
Closes #56

### DIFF
--- a/exercises/hello-world/src/example.c
+++ b/exercises/hello-world/src/example.c
@@ -1,0 +1,8 @@
+#include <stdio.h>
+#include "hello_world.h"
+
+void hello(char* buffer, unsigned int buffer_length, const char* name) {
+    if (name == NULL) name = "World";
+
+    snprintf(buffer, buffer_length, "Hello, %s!", name);
+}
--- a/exercises/hello-world/src/example.c
+++ b/exercises/hello-world/src/example.c
@@ -1,4 +1,5 @@
 #include <stdio.h>
+#include <stddef.h>
 #include "hello_world.h"
 
 void hello(char* buffer, unsigned int buffer_length, const char* name) {
--- a/exercises/hello-world/src/example.h
+++ b/exercises/hello-world/src/example.h
@@ -1,6 +1,0 @@
-#ifndef _HELLO_WORLD_H
-#define _HELLO_WORLD_H
-
-void hello(char* buffer, unsigned int buffer_length, const char* name);
-
-#endif
--- a/exercises/hello-world/src/example.h
+++ b/exercises/hello-world/src/example.h
@@ -1,8 +1,6 @@
-#include <stdio.h>
+#ifndef _HELLO_WORLD_H
+#define _HELLO_WORLD_H
 
-void hello(char* buffer, unsigned int buffer_length, const char* name){
-    
-    if (name[0] == '\0') name = "World";
-    
-    snprintf(buffer, buffer_length, "Hello, %s!", name);
-}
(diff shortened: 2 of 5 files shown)


exercises/hello-world/src/example.c
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
```
dpepoon: Is this required...
~~~
#include <stddef.h>
~~~

ryanplusplus: That's what provides the `NULL` constant.

ryanplusplus: I need to add that include to the test, though, since it uses `NULL` without explicitly including `stddef.h`.

ryanplusplus: Fixed

dpepoon: I was suggesting that it be added to the .c file.

ryanplusplus: Oh, my bad. You're right. Added.

ryanplusplus:
Please have another look. This now addresses #43 and #56 and should be a much gentler introduction to the track.

patricksjackson:
That looks good to me.

exercises/hello-world/src/example.c
```diff
@@ -2,8 +2,8 @@
 #include <stddef.h>
 #include "hello_world.h"
 
-void hello(char* buffer, unsigned int buffer_length, const char* name) {
+void hello(char *buffer, const char *name) {
     if (name == NULL) name = "World";
```
palisaide: Let's make this line the first thing we change after closing #40

palisaide:
👍

exercises/hello-world/src/example.c
ryanplusplus: Good deal

ryanplusplus:
Hey @patricksjackson -- I really appreciate your effort on reviews. Are you interested in becoming a co-maintainer for this track and getting committer rights? If you're just passing through that's fine too.

patricksjackson:
Sure. I'd been meaning to ask about that here or at one of the other tracks at some point :)

ryanplusplus:
Thanks, I've asked @kytrinyx to add you (I can't do so myself).

PR #47, opened 2016-08-25
Fixed include guard for Anagram.h

ryanplusplus:
Looks like a gremlin got left in when we changed this from `example.h` to `anagram.h`.


exercises/anagram/src/anagram.h
```diff
@@ -1,5 +1,5 @@
-#ifndef _EXAMPLE_H
-#define _EXAMPLE_H
+#ifndef _ANAGRAM_H
+#define _ANAGRAM_H
```
dpepoon: 👍

PR #48, opened 2016-08-28
Add Gigasecond exercise

krbullock:


config.json
```diff
@@ -8,6 +8,7 @@
     "hello-world",
     "anagram",
     "leap",
+    "gigasecond",
```
ryanplusplus: In addition to the `problems` list (which has been deprecated), there's an `exercises` list below. Can you add this exercise to that list as well? See https://github.com/exercism/xc/issues/37 for more information.

exercises/gigasecond/test/test_gigasecond.c
```diff
@@ -0,0 +1,67 @@
...
+ * (date, time, offset from UTC) into a time_t (seconds since Unix
+ * epoch).
+ */
+time_t parse_date(const char *const str) {
+  struct tm date;
+  strptime(str, "%F %T %z", &date);
+  return mktime(&date);
+}
+
+void test_date(void) {
```
ryanplusplus: This problem has a [common test list](https://github.com/exercism/x-common/blob/master/gigasecond.json). Can you update these tests to match the common test list so that we can keep this exercise in sync with other tracks?

ryanplusplus:
This looks great. Once `config.json` is updated I think that we can merge this.

config.json
krbullock: Ahh, thanks for the reference. I didn't see the `exercises` list in the contributing guide or in other tracks (Ruby, Rust). I'll add that in.

krbullock:
@ryanplusplus Updated.

palisaide:
Looks good!

ryanplusplus:
Thanks!

PR #50, opened 2016-08-30
Add beer-song exercise

### LINKED ISSUE #49: Add "Beer Song" exercise

palisaide:
Closes #49

### DIFF
--- a/config.json
+++ b/config.json
@@ -75,6 +75,13 @@
       "bitwise operations",
       "performance optimizations"
     ]
+  }, {
+    "difficulty": 2,
+    "slug": "beer-song",
+    "topics": [
+      "strings",
+      "control-flow (if-else statements)"
+    ]
   }],
   "deprecated": [
 
--- a/config.json
+++ b/config.json
@@ -12,7 +12,8 @@
     "hamming",
     "raindrops",
     "bob",
-    "grains"
+    "grains",
+    "beer-song"
   ],
   "exercises": [{
     "slug": "hello-world",
--- a/config.json
+++ b/config.json
@@ -80,7 +80,8 @@
     "slug": "beer-song",
     "topics": [
       "strings",
-      "control-flow (if-else statements)"
+      "control-flow (if-else statements)",
+      "control-flow (loops)"
     ]
   }],
   "deprecated": [
--- a/exercises/beer-song/makefile
+++ b/exercises/beer-song/makefile
@@ -21,6 +21,6 @@ test: tests.out
 clean:
 	rm -f *.o *.out
 
-tests.out: test/test_beer_song.c src/beer_song.c
+tests.out: test/test_beer_song.c src/beer_song.c src/beer_song.h
 	@echo Compiling $@
 	@cc src/beer_song.c test/vendor/unity.c test/test_beer_song.c -o tests.out
--- a/exercises/beer-song/makefile
+++ b/exercises/beer-song/makefile
@@ -1,0 +1,26 @@
+CFLAGS=-std=c99
+CFLAGS += -Wall
+CFLAGS += -Wextra
+CFLAGS += -Werror
+CFLAGS += -Wpointer-arith
+CFLAGS += -Wcast-align
+CFLAGS += -Wwrite-strings
+CFLAGS += -Wswitch-default
+CFLAGS += -Wunreachable-code
+CFLAGS += -Winit-self
+CFLAGS += -Wmissing-field-initializers
+CFLAGS += -Wno-unknown-pragmas
+CFLAGS += -Wstrict-prototypes
+CFLAGS += -Wundef
+CFLAGS += -Wold-style-definition
+
+
+test: tests.out
+	@./tests.out
+
+clean:
+	rm -f *.o *.out
+
+tests.out: test/test_beer_song.c src/beer_song.c
+	@echo Compiling $@
+	@cc src/beer_song.c test/vendor/unity.c test/test_beer_song.c -o tests.out
--- a/exercises/beer-song/src/example.c
+++ b/exercises/beer-song/src/example.c
@@ -2,18 +2,18 @@
 #include <stdio.h>
 #include <string.h>
 
-static int getVerse(char *buffer, int verseNumber) {
-  int bytesWritten = 0;
+static unsigned int getVerse(char *buffer, unsigned int verseNumber) {
+  unsigned int bytesWritten = 0;
 
   if(verseNumber == 0) {
     bytesWritten = sprintf(buffer, "No more bottles of beer on the wall, no more bottles of beer.\n"
       "Go to the store and buy some more, 99 bottles of beer on the wall.\n");
-  } else if(verseNumber == 2) {
-    bytesWritten = sprintf(buffer, "2 bottles of beer on the wall, 2 bottles of beer.\n"
-        "Take one down and pass it around, 1 bottle of beer on the wall.\n");
   } else if(verseNumber == 1) {
     bytesWritten = sprintf(buffer, "1 bottle of beer on the wall, 1 bottle of beer.\n"
       "Take it down and pass it around, no more bottles of beer on the wall.\n");
+  } else if(verseNumber == 2) {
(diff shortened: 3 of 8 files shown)


patricksjackson:
I tested this and it works fine, but it needs at least one small tweak:

The example solution needs to be written as if the filenames are actually beer_song.c/h. So the include guards and the `#include` need to reference `beer_song.c/h`. This makes it easy to just rename the files for testing.

The function `resetTest()` in the testing file doesn't appear to be necessary, although I don't know all that much about Unity. At least some of the other problems have the same stub though...

(Formatting will need to change after we close #40, but we're going to need to make a pass through all the problems anyway, so I think this is fine for now)

palisaide:
@patricksjackson, thanks for clarifying the example convention.  I saw some different approaches in the other exercises and I wasn't sure which was correct.  However, now I'm wondering if there's any reason to call the example `example.c` and not just `beer_song.c`.  I think it would be obvious that the files in `src` are just an example.  It would also make it easier to get the CI build to actually run the tests on the examples.

I just tested without the empty `resetTest()` and everything appears to be fine, so I guess Unity doesn't need it.

patricksjackson:
They need to be called `example.*` so that the exercism client doesn't download them when you fetch exercises. If they were named `beer_song.*` they'd be downloaded along with the rest of the exercise. Doing it that way lets the tracks provide skeleton files if they want to.

Everything looks good to me :) :+1:

config.json
```diff
@@ -74,6 +75,13 @@
       "bitwise operations",
       "performance optimizations"
     ]
+  }, {
+    "difficulty": 2,
+    "slug": "beer-song",
+    "topics": [
+      "strings",
+      "control-flow (if-else statements)"
```
ryanplusplus: How about `control-flow (loops)`?

exercises/beer-song/src/example.c
```diff
@@ -0,0 +1,40 @@
...
+      "Take it down and pass it around, no more bottles of beer on the wall.\n");
+  } else if(verseNumber <= 99) {
+    bytesWritten = sprintf(buffer, "%d bottles of beer on the wall, %d bottles of beer.\n"
+      "Take one down and pass it around, %d bottles of beer on the wall.\n",
+      verseNumber, verseNumber, verseNumber - 1);
+  }
+
+  return bytesWritten;
+}
+void verse(char *buffer, int verseNumber) {
```
ryanplusplus: Can you add a newline above this?

exercises/beer-song/test/test_beer_song.c
```diff
@@ -0,0 +1,93 @@
...
+    "Go to the store and buy some more, 99 bottles of beer on the wall.\n";
+
+  sing(response, 3, 0);
+  TEST_ASSERT_EQUAL_STRING(expected, response);
+}
+
+int main(void) {
+  UnityBegin("test/test_beer_song.c");
+
+  RUN_TEST(test_handles_arbirtary_verse);
```
ryanplusplus: `arbirtrary` => `arbitrary`

exercises/beer-song/test/test_beer_song.c
```diff
@@ -0,0 +1,93 @@
...
+    "6 bottles of beer on the wall, 6 bottles of beer.\n"
+    "Take one down and pass it around, 5 bottles of beer on the wall.\n";
+
+  sing(response, 8, 6);
+  TEST_ASSERT_EQUAL_STRING(expected, response);
+}
+
+void test_sings_the_last_4_verses(void) {
+  char response[BUFFER_SIZE * 4];
+  const char expected[BUFFER_SIZE * 4] =
```
ryanplusplus: Any reason to not just use the larger buffer always?

exercises/beer-song/src/example.c
```diff
@@ -0,0 +1,40 @@
...
+
+  if(verseNumber == 0) {
+    bytesWritten = sprintf(buffer, "No more bottles of beer on the wall, no more bottles of beer.\n"
+      "Go to the store and buy some more, 99 bottles of beer on the wall.\n");
+  } else if(verseNumber == 2) {
+    bytesWritten = sprintf(buffer, "2 bottles of beer on the wall, 2 bottles of beer.\n"
+        "Take one down and pass it around, 1 bottle of beer on the wall.\n");
+  } else if(verseNumber == 1) {
+    bytesWritten = sprintf(buffer, "1 bottle of beer on the wall, 1 bottle of beer.\n"
+      "Take it down and pass it around, no more bottles of beer on the wall.\n");
```
ryanplusplus: Pretty serious nitpick here: your cases go 0, 2, then 1. How about 0, 1, 2?

exercises/beer-song/src/example.c
```diff
@@ -0,0 +1,40 @@
...
+
+static int getVerse(char *buffer, int verseNumber) {
+  int bytesWritten = 0;
+
+  if(verseNumber == 0) {
+    bytesWritten = sprintf(buffer, "No more bottles of beer on the wall, no more bottles of beer.\n"
+      "Go to the store and buy some more, 99 bottles of beer on the wall.\n");
+  } else if(verseNumber == 2) {
+    bytesWritten = sprintf(buffer, "2 bottles of beer on the wall, 2 bottles of beer.\n"
+        "Take one down and pass it around, 1 bottle of beer on the wall.\n");
```
ryanplusplus: Indentation is a bit off here

exercises/beer-song/src/example.c
```diff
@@ -0,0 +1,40 @@
...
+      verseNumber, verseNumber, verseNumber - 1);
+  }
+
+  return bytesWritten;
+}
+void verse(char *buffer, int verseNumber) {
+  (void)getVerse(buffer, verseNumber);
+}
+
+void sing(char *buffer, int startingVerseNumber, int endingVerseNumber) {
```
ryanplusplus: Should we use an unsigned value for the verse numbers?

ryanplusplus:
Looks good to me. Almost ready.

exercises/beer-song/src/example.c
palisaide: Yes.

exercises/beer-song/src/example.c
palisaide: Yep, this bothers me as well.

exercises/beer-song/src/example.h
```diff
@@ -0,0 +1,7 @@
+#ifndef _BEER_SONG_H
+#define _BEER_SONG_H
+
+void verse(char *buffer, int verseNumber);
+void sing(char *buffer, int startingVerseNumber, int endingVerseNumber);
```
ryanplusplus: Looks like the prototypes didn't get the `unsigned` treatment

exercises/beer-song/src/example.c
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  return bytesWritten;
+}
+
+void verse(char *buffer, unsigned int verseNumber) {
+  (void)getVerse(buffer, verseNumber);
+}
+
+void sing(char *buffer, int startingVerseNumber, int endingVerseNumber) {
```
ryanplusplus: These guys are still signed

kytrinyx:
> The example solution needs to be written as if the filenames are actually 

@patricksjackson Would this be worth articulating in the README for the track?

patricksjackson:
@kytrinyx Yep, it's something to call out. The filenaming scheme is common for all the tracks (so it's in the x-common contributing guide) but C and Cpp have to actually have this conflict in submitted code.

I'll make sure it gets in.

patricksjackson:
LGTM

exercises/beer-song/makefile
```diff
@@ -0,0 +1,26 @@
...
+CFLAGS += -Wold-style-definition
+
+
+test: tests.out
+	@./tests.out
+
+clean:
+	rm -f *.o *.out
+
+tests.out: test/test_beer_song.c src/beer_song.c
```
ryanplusplus: Not to prolong this forever, but can you add `src/beer_song.h` to make sure this rebuilds if the header is changed?

ryanplusplus:
Can you re-run the indent script as well? I'm fairly certain this passes Travis only because those checks were added later.

PR #51, opened 2016-08-30
Add 'grains' exercise

palisaide:
This was modeled after the C++ implementation, which uses 1-based indexing (e.g. `square(1)` returns 1).  However, I'm wondering if this is the best approach.  It certainly makes the tests read better, but we might be teaching the wrong lesson.  Most experienced C developers would consider square 0 to be the first square.

### DIFF
--- a/config.json
+++ b/config.json
@@ -11,7 +11,8 @@
     "gigasecond",
     "hamming",
     "raindrops",
-    "bob"
+    "bob",
+    "grains"
   ],
   "exercises": [{
     "slug": "hello-world",
@@ -64,6 +65,12 @@
     "topics": [
       "strings",
       "control-flow (if-else statements)"
+    ]
+  }, {
+    "difficulty": 1,
+    "slug": "grains",
+    "topics": [
+      "control-flow (loops)"
     ]
   }],
   "deprecated": [
--- a/config.json
+++ b/config.json
@@ -70,7 +70,9 @@
     "difficulty": 1,
     "slug": "grains",
     "topics": [
-      "control-flow (loops)"
+      "control-flow (loops)",
+      "bitwise operations",
+      "performance optimizations"
     ]
   }],
   "deprecated": [
--- a/exercises/grains/src/example.c
+++ b/exercises/grains/src/example.c
@@ -1,4 +1,4 @@
-#include "example.h"
+#include "grains.h"
 
 #define NUMBER_OF_SQUARES   (64)
 
--- a/exercises/grains/src/example.c
+++ b/exercises/grains/src/example.c
@@ -1,0 +1,18 @@
+#include "example.h"
+
+#define NUMBER_OF_SQUARES   (64)
+
+unsigned long square(int index) {
+  return 1ul << (index - 1);
+}
+
+unsigned long total(void) {
+  unsigned long result = 0;
+  int i;
+
+  for(i = 1; i <= NUMBER_OF_SQUARES; i++) {
+    result += square(i);
+  }
+
+  return result;
+}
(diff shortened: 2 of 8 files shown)


config.json
```diff
@@ -65,6 +66,12 @@
       "strings",
       "control-flow (if-else statements)"
     ]
+  }, {
+    "difficulty": 1,
+    "slug": "grains",
+    "topics": [
+      "control-flow (loops)"
```
patricksjackson: You should add 'Bitwise operations' and possibly 'Performance' to the topics. The problem implies that you're supposed to attempt to optimize for performance.

exercises/grains/test/test_grains.c
```diff
@@ -0,0 +1,53 @@
+#include "vendor/unity.h"
+#include "../src/grains.h"
+
+void testReset(void) {
+}
```
patricksjackson: Unnecessary.

exercises/grains/src/example.h
```diff
@@ -0,0 +1,7 @@
+#ifndef _GRAINS_H
+#define _GRAINS_H
+
+unsigned long square(int sq);
+unsigned long total(void);
```
patricksjackson: Is an unsigned long guaranteed to hold 64 bits?

palisaide: I personally would prefer using fixed-width types, but all of the other exercises seem to favor  primitive types.  What's the consensus?  Should we pull in `stdint.h` and use `uint64_t`?

config.json
palisaide: 👍

exercises/grains/src/example.h
patricksjackson: Personally I would prefer pulling in uint64_t for things that require fixed-width types. It might be worthwhile to make an issue to discuss.

exercises/grains/test/test_grains.c
```diff
@@ -0,0 +1,50 @@
...
+void test_square_64(void) {
+  TEST_ASSERT_EQUAL(9223372036854775808ul, square(64));
+}
+
+void test_total(void) {
+  TEST_ASSERT_EQUAL(18446744073709551615ul, total());
+}
+
+int main(void) {
+  UnityBegin("test/test_hello_world.c");
```
palisaide: Oops...

patricksjackson:
Looking at the Unity source, it looks like TEST_ASSERT_EQUAL casts the inputs to signed integers to compare them. Those calls should be changed to TEST_ASSERT_EQUAL_UINT64.

exercises/grains/src/example.h
ryanplusplus: I agree with @patricksjackson -- let's use stdint types here.

palisaide: @ryanplusplus, I'm assuming you mean switch over to stdint types in the entire exercise, right?

exercises/grains/src/example.c
```diff
@@ -0,0 +1,18 @@
...
+
+unsigned long square(int index) {
+  return 1ul << (index - 1);
+}
+
+unsigned long total(void) {
+  unsigned long result = 0;
+  int i;
+
+  for(i = 1; i <= NUMBER_OF_SQUARES; i++) {
```
ryanplusplus: We should also decide what C standard we're targeting. I'm partial to C99, but I'll create an issue to discuss.

exercises/grains/src/example.h
ryanplusplus: @palisaide Yes

ryanplusplus:
I'm happy with this once we get fully switched over to fixed width types (ie: `uint64_t`).

ryanplusplus:
Looks good, thanks!

Issue #53, opened 2016-08-30
Should we settle on C99?


ryanplusplus:
This will give us (guaranteed) access to stdint types and is generally a more pleasant experience than C89. Shall we proceed to take advantage of C99 in all of our examples and tests where appropriate?

palisaide:
I think C99 is the way to go.  I can't think of any negatives.

ryanplusplus:
Okay, let's go for it.

Issue #52, opened 2016-08-30
Should we use fixed-width types (e.g. stdint) exclusively?


palisaide:
Should we use fixed-width types (e.g. stdint) exclusively or favor primitive types unless a size-guaranteed type is required for the particular exercise?

ryanplusplus:
I personally feel that fixed width types (almost) always make the most sense. So my 2 cents is yes, we should.

patricksjackson:
I agree. The only real downside is adding some confusion for people new to the language, but it will be mostly hidden from them anyways.

PR #54, opened 2016-08-31
Add pangram exercise

palisaide:
This will probably need to be updated depending on the outcome of #52 and #53

### DIFF
--- a/config.json
+++ b/config.json
@@ -79,7 +79,6 @@
     "difficulty": 3,
     "slug": "pangram",
     "topics": [
-      "filtering",
       "strings",
       "control-flow (if-else statements)",
       "control-flow (loops)"
--- a/config.json
+++ b/config.json
@@ -12,7 +12,8 @@
     "hamming",
     "raindrops",
     "bob",
-    "grains"
+    "grains",
+    "pangram"
   ],
   "exercises": [{
     "slug": "hello-world",
@@ -74,6 +75,15 @@
       "bitwise operations",
       "performance optimizations"
     ]
+  }, {
+    "difficulty": 3,
+    "slug": "pangram",
+    "topics": [
+      "filtering",
+      "strings",
+      "control-flow (if-else statements)",
+      "control-flow (loops)"
+  ]
   }],
   "deprecated": [
 
--- a/exercises/pangram/src/example.c
+++ b/exercises/pangram/src/example.c
@@ -1,26 +1,25 @@
 #include "pangram.h"
 #include <string.h>
+#include <ctype.h>
 
 #define NUMBER_OF_LETTERS (26)
-#define DOWNCASE(letter) ((letter >= 'A') && (letter <= 'Z')) ? letter + ('a' - 'A') : letter
 
 bool is_pangram(const char *sentence) {
-   int sentence_length = strlen(sentence);
-   char found_letters[NUMBER_OF_LETTERS];
-   int i;
+   size_t sentence_length = strlen(sentence);
+   bool found_letters[NUMBER_OF_LETTERS];
 
    memset(found_letters, 0, sizeof(found_letters));
 
-   for(i = 0; i < sentence_length; i++) {
-      char letter = DOWNCASE(sentence[i]);
+   for(size_t i = 0; i < sentence_length; i++) {
+      char letter = tolower(sentence[i]);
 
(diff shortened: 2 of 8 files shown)


exercises/pangram/src/example.c
```diff
@@ -0,0 +1,29 @@
+#include "pangram.h"
+#include <string.h>
+
+#define NUMBER_OF_LETTERS (26)
+#define DOWNCASE(letter) ((letter >= 'A') && (letter <= 'Z')) ? letter + ('a' - 'A') : letter
+
+bool is_pangram(const char *sentence) {
+   int sentence_length = strlen(sentence);
```
ryanplusplus: `strlen` actually returns a `size_t` so you should use that here and for `i`.

exercises/pangram/src/example.c
```diff
@@ -0,0 +1,29 @@
...
+   char found_letters[NUMBER_OF_LETTERS];
+   int i;
+
+   memset(found_letters, 0, sizeof(found_letters));
+
+   for(i = 0; i < sentence_length; i++) {
+      char letter = DOWNCASE(sentence[i]);
+
+      if((letter >= 'a') && letter <= 'z') {
```
ryanplusplus: Inconsistent use of parens (included on the first check, omitted on the second).

exercises/pangram/src/example.h
```diff
@@ -0,0 +1,7 @@
+#ifndef _PANGRAM_H
+#define _PANGRAM_H
+
+#include <stdbool.h>
+
+bool is_pangram(const char *sentence);
+#endif
```
ryanplusplus: Can you add a line of whitespace above to be consistent with the formatting of other exercises?

config.json
```diff
@@ -74,6 +75,15 @@
       "bitwise operations",
       "performance optimizations"
     ]
+  }, {
+    "difficulty": 3,
+    "slug": "pangram",
+    "topics": [
+      "filtering",
```
ryanplusplus: I'm not sure that this really covers filtering. I think this really refers to whittling down a collection to a smaller collection.

ryanplusplus:
Looks good overall, thanks

patricksjackson:
Apparently @ryanplusplus and I were going through this at the same time, there may be a bit of overlap.

On the testing:

1. Please add more of the test cases from the [common json](https://github.com/exercism/x-common/blob/master/pangram.json), at the very least use one with numbers to catch people accidentally testing for `isalnum()`.
2. For C we should also have a NULL test case.

Now the example solutions shouldn't need to be overly fancy, just something that passes all the tests, but it does need a couple changes:

1. `size_t i, sentence_length;`
2. Your solution would fail if an input was given that overflowed `char`.
3. Using `isalpha()` and `tolower()` may be better to introduce people to the library functions.

Actually, in all honesty I've never been clear on what requirements we should actually place on the example solutions, but those seem like good things to have.

ryanplusplus:
@patricksjackson good catches. I definitely agree on `tolower` and `isalpha`. I am fortunate to not have to do much text processing in C so I didn't even think of these.

As far as the requirements for the example solutions, I am of the mind that the sample solutions should be something that someone could learn from if they were to stumble upon this repo in search of better ways to complete the exercises. I strive to write "refrigerator code" (code you'd take home and put on your parents' refrigerator) for sample solutions, so I would encourage you to keep the nitpicks coming!

exercises/pangram/src/example.c
```diff
@@ -0,0 +1,29 @@
...
+         found_letters[letter - 'a']++;
```
ryanplusplus: How about using a `bool` array and changing this to:

```
found_letters[letter - 'a'] |= true;
```

patricksjackson: That would've been my suggestion as well (though you only need `=` and not `|=`)

ryanplusplus: Oh derp. Good call.

palisaide:
Thanks for the comments!  So, I totally forgot that there was a common json for most of these exercises; I was going through other languages trying to find test cases.

I agree with all suggestions.

ryanplusplus:
Changes look good to me. I'm going to give @patricksjackson a chance to review them and then I'll merge.

patricksjackson:
LGTM :+1:

PR #57, opened 2016-09-02
Add 'nth-prime' exercise

palisaide:


config.json
```diff
@@ -83,6 +84,14 @@
...
       "control-flow (loops)"
   ]
+  }, {
+    "difficulty": 4,
+    "slug": "nth-prime",
+    "topics": [
+      "control-flow (if-else statements)",
+      "control-flow (loops)",
+      "performance optimizations"
+  ]
```
ryanplusplus: Please indent this

exercises/nth-prime/src/example.c
```diff
@@ -0,0 +1,32 @@
+#include "nth_prime.h"
+#include <stdbool.h>
+#include <math.h>
+
+static bool is_prime(uint32_t n)
+{
+  for(uint32_t i = (uint32_t)sqrt(n); i > 1; i--) {
```
patricksjackson: This is a pretty inefficient way to do trial division. Low numbers factor into more numbers than higher numbers, so it's more efficient to start from the bottom. If you wrote it as `for(uint32_t i = 2; i * i < n: ++i)` you could get rid of the `math.h` include as well.

(Sorry for being pedantic on the algorithm, [I've done a lot of prime stuff lately](https://github.com/patricksjackson/primes-cpp))...

ryanplusplus: I agree, good call.

exercises/nth-prime/src/example.c
```diff
@@ -0,0 +1,32 @@
...
+    if(n % i == 0) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+uint32_t nth(uint32_t n)
+{
```
ryanplusplus: A couple of these leading brackets are on the next line instead of the same line. Can you make these consistent within this file?

exercises/nth-prime/src/example.c
ryanplusplus: If you wanted to get _really_ crazy, you could use C99 variable length arrays and do a real sieve. That should be a fair bit faster, but this is also a nice, simple solution.

patricksjackson: Sieving would be faster for finding a large range of primes, but this is just checking one. The fastest ways to check a single prime are either using the [probabilistic primality testers](https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test) with the correct set of witnesses to make the algorithms deterministic, or by trial division (like above). You _can_ speed up trial division by reducing the number of checks through wheel factorization, like I did [here](https://github.com/patricksjackson/primes-cpp/blob/master/primes.cpp#L288), but even compared to the above it is only a speed bump of ~4x.

ryanplusplus: Yeah my bad. I was thinking a sieve could be used to find N primes, but I now realize there's no good way to set the upper bound on the sieve in order to guarantee that you'll find N primes.

palisaide: @patricksjackson, good call.  I've updated the algorithm.

exercises/nth-prime/src/example.c
palisaide: Formatting should now be fixed according to #40

ryanplusplus: Yeah my bad on this. Didn't realize that the Linux kernel style has a bracket placement exception for functions.

ryanplusplus:
LGTM

exercises/nth-prime/src/example.c
palisaide: > Heretic people all over the world have claimed that this inconsistency
> is ...  well ...  inconsistent, but all right-thinking people know that
> (a) K&R are _right_ and (b) K&R are right.

exercises/nth-prime/src/example.c
```diff
@@ -0,0 +1,32 @@
+#include "nth_prime.h"
+#include <stdbool.h>
+#include <math.h>
```
patricksjackson: Math.h isn't necessary anymore.

Otherwise it LGTM.

PR #58, opened 2016-09-02
Configure Travis task to execute tests for all examples

palisaide:
My idea is just to have a shell script that will copy/rename the `example.c` and `example.h` files with the appropriate names and then execute `make` in each exercise directory.  I do not claim to know anything about bash scripts, so suggestions are greatly encouraged.

### DIFF
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -10,21 +10,19 @@ for D in ../exercises/*; do
         # Copy the examples with the correct name for the exercise
         if [ -e "${D}/src/example.c" ]
         then
-          echo Generating "${D}/src/${EXERCISE_NAME}.c"
           cp -n ${D}/src/example.c ${D}/src/${EXERCISE_NAME}.c
         fi
 
         if [ -e "${D}/src/example.h" ]
         then
-          echo Generating "${D}/src/${EXERCISE_NAME}.h"
           cp -n ${D}/src/example.h ${D}/src/${EXERCISE_NAME}.h
         fi
 
         # Make it!
-        ( cd ${D} ; make clean; make )
-
-        # Clean up
-        #rm -f ${D}/src/${EXERCISE_NAME}.c
-        #rm -f ${D}/src/${EXERCISE_NAME}.h
+        ( cd ${D};
+          echo "Running tests for ${EXERCISE_NAME}";
+          make clean >> /dev/null; # I'm just trying to suppress the output (I'm sure there's a much better way)
+          make | grep FAIL:  # Only output failures
+        )
     fi
 done
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -8,14 +8,23 @@ for D in ../exercises/*; do
         EXERCISE_NAME="${STRIPPED_OF_EXTENSION:5}"
 
         # Copy the examples with the correct name for the exercise
-        cp -f ${D}/src/example.c ${D}/src/${EXERCISE_NAME}.c
-        cp -f ${D}/src/example.h ${D}/src/${EXERCISE_NAME}.h
+        if [ -e "${D}/src/example.c" ]
+        then
+          echo Generating "${D}/src/${EXERCISE_NAME}.c"
+          cp -n ${D}/src/example.c ${D}/src/${EXERCISE_NAME}.c
+        fi
+
+        if [ -e "${D}/src/example.h" ]
+        then
+          echo Generating "${D}/src/${EXERCISE_NAME}.h"
+          cp -n ${D}/src/example.h ${D}/src/${EXERCISE_NAME}.h
+        fi
 
         # Make it!
         ( cd ${D} ; make clean; make )
 
         # Clean up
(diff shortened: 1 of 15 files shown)


palisaide:
The script appears to be working as it is correctly detecting a failed test in the anagram exercise.

palisaide:
The question now is: Do we leave the build failing and fix the actual failures in a separate ticket or fix them here so that the merge results in a passing build?

ryanplusplus:
Depends... how complicated is the problem? If it's small then fix it here. If it's large then fix it in another PR. We've been broken this long it won't be a big deal to wait for the other PR.

palisaide:
There are multiple problems with multiple exercises, so I created #62 to fix them.

bin/run-tests
```diff
@@ -0,0 +1,34 @@
...
+    if [ -d "${D}" ]; then
+        # Get the exercise name from the test file
+        TEST_FILE=$(basename $(ls ${D}/test/test*))
+        STRIPPED_OF_EXTENSION="${TEST_FILE%.*}"
+        EXERCISE_NAME="${STRIPPED_OF_EXTENSION:5}"
+
+        # Copy the examples with the correct name for the exercise
+        if [ -e "${D}/src/example.c" ]
+        then
+          cp -n ${D}/src/example.c ${D}/src/${EXERCISE_NAME}.c
```
patricksjackson: Remove the `-n` from both `cp` commands. We want them to override the files, like with our dummpy `hello_world.c`

bin/run-tests
```diff
@@ -0,0 +1,34 @@
...
+
+        if [ -e "${D}/src/example.h" ]
+        then
+          cp -n ${D}/src/example.h ${D}/src/${EXERCISE_NAME}.h
+        fi
+
+        # Make it!
+        { cd ${D};
+          echo "Running tests for ${EXERCISE_NAME}";
+          make clean >> /dev/null; # I'm just trying to suppress the output (I'm sure there's a much better way)
```
patricksjackson: Directing output to /dev/null works fine. This line shouldn't be necessary on Travis but it certainly doesn't hurt and makes using the script locally easier, so I say leave it in.

Maybe remove the comment though.

bin/run-tests
```diff
@@ -0,0 +1,34 @@
...
+          if make | grep FAIL: ; then
+            set -e
```
patricksjackson: Remove `set -e`. It does nothing in this script.

patricksjackson:
In working the anagram problem I figured out that you could setup a Travis image locally for testing the config. [Check it out here](https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally)

Travis is on Ubuntu 12.04 and GCC 4.6, so you'll have to make a change to all the makefiles since `-Wpedantic` didn't exist in that version of GCC. Just edit them with this commit so that we keep all the Travis setup requirements in one place. The sed command from the `exercism/xc` root is

>sed -i -- 's/-Wpedantic/-pedantic/g' exercises/*/makefile

Overall this script works great, just needs a few minor changes and it'll be good to go :)

ryanplusplus:
Wow, a local Travis image will help me immensely with #73 as I think I'm running into a similar issue with an older versions. Thanks for the link!

ryanplusplus:
@palisaide is this ready for another look?

palisaide:
Yep, should be good to go now

patricksjackson:
Oops, I meant to say that you can run that sed command to touch all the makefiles, not to use it in the script. Running sed every build could introduce issues if we need to retouch the build system.

'-pedantic' is identical to '-Wpedantic' on modern compilers AFAIK, so we should just change it everywhere.

Everything else looks good :)

palisaide:
Ahh, got it.  All of the makefiles should now be updated and the command removed from the script.

bin/run-tests
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+for D in exercises/*; do
+    CURRENT_DIR=$(pwd)
+
+    if [ -d "${D}" ]; then
+        # Get the exercise name from the test file
+        TEST_FILE=$(basename $(ls ${D}/test/test*))
```
ryanplusplus: Isn't the exercise name just `$D`?

ryanplusplus: Okay, I think that this is really for getting the underscored name for the exercise (ie: `hello_world` instead of `hello-world`). I guess this raises the question of why we don't simply name the source exactly the way the exercise is named.

ryanplusplus:
One comment, but I don't want to drag this out forever. If @patricksjackson is good with this then so am I.

bin/run-tests
palisaide: Correct.  And I've asked myself the same question.

patricksjackson:
Awesome :)

@ryanplusplus - If you want to talk about the naming conventions we can somewhere else, but my feeling is that they're fine the way they are.

ryanplusplus:
I'm happy to leave them as they are

PR #67, opened 2016-09-07
Make: rebuild when header files change

### LINKED ISSUE #64: Makefile builds do not track dependencies
If you change a depended-upon header and re-`make` it will not recompile C files that include that header. We could do this the complicated way and build dep files, or we could weasel out and just add header files as dependencies for the .o build targets. I think that I'm in favor of keeping it simple at this point. Thoughts?

ryanplusplus:
Closes #64

### DIFF
--- a/exercises/anagram/makefile
+++ b/exercises/anagram/makefile
@@ -1,7 +1,7 @@
 CFLAGS=-std=c99
 CFLAGS += -Wall
 CFLAGS += -Wextra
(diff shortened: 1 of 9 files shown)


ryanplusplus:
Updated to fix all exercises

patricksjackson:
LGTM

palisaide:
Looks good 👍

PR #68, opened 2016-09-08
Removing non-standard time functions.

### LINKED ISSUE #60: For gigasecond, is it more complicated than just adding 1e9 seconds?
Because simply adding 1e9 seconds works for the other languages, but here it only passes 3/5 tests.

### LINKED ISSUE #66: Gigasecond: strptime is not a C standard function
`strptime` is not available in vanilla C99 and is a Unix extension (https://www.gnu.org/software/libc/manual/html_node/Parsing-Date-and-Time.html). You need to define `_GNU_SOURCE` and `__USE_XOPEN` on Ubuntu to have `time.h` bring it in. We should remove this to improve compatibility across platforms.

patricksjackson:
Converting the gigasecond exercise to use `struct tm` and `mktime()`.

This forcibly turns off Daylight Savings Time, which may have been a contributing factor for the problems we were seeing. (If I let my system choose what to use, the tests would fail ~25% of the time. I still have no clue as to why.)

Closes #66 
Closes #60

### DIFF
--- a/exercises/gigasecond/src/example.c
+++ b/exercises/gigasecond/src/example.c
@@ -1,7 +1,4 @@
-#include <stdio.h>
-#include <math.h>
-
-#include "example.h"
+#include "gigasecond.h"
 
 time_t gigasecond_after(time_t start)
 {
--- a/exercises/gigasecond/test/test_gigasecond.c
+++ b/exercises/gigasecond/test/test_gigasecond.c
@@ -1,59 +1,60 @@
-#include <time.h>
-
 #include "vendor/unity.h"
 #include "../src/gigasecond.h"
 
-/* Convert a date string in the format "YYYY-MM-DD HH:MM:SS +NNNN"
- * (date, time, offset from UTC) into a time_t (seconds since Unix
- * epoch).
- */
-time_t parse_date(const char *const str)
+// Constructs a time_t type from the given date settings
+time_t construct_date(int year, int month, int day, int hour, int min, int sec)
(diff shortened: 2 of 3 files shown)


exercises/gigasecond/src/example.c
```diff
@@ -1,7 +1,4 @@
-#include <stdio.h>
-#include <math.h>
-
-#include "example.h"
+#include "gigasecond.h"
```
ryanplusplus: Excellent! This has been driving me crazy.

exercises/gigasecond/test/test_gigasecond.c
```diff
@@ -1,59 +1,60 @@
...
+   time_t actual = gigasecond_after(construct_date(2015, 1, 24, 23, 59, 59));
+   TEST_ASSERT(expected == actual);
 }
 
 /*
 void test_your_birthday(void) {
-  time_t expected = parse_date("");
-  time_t actual = gigasecond_after(parse_date(""));
-  TEST_ASSERT_EQUAL_INT(expected, actual);
+   time_t birthday = construct_date(1989,1,1,1,1,1);
```
ryanplusplus: Travis didn't catch this because it's in a comment, but the formatting is a bit off

ryanplusplus:
LGTM outside of a tiny nitpick

ryanplusplus:
LGTM

palisaide:
Looks good 👍

PR #69, opened 2016-09-08
Updating makefiles with new default flags.

### LINKED ISSUE #61: Default Makefile Flags
While looking into what was happening with #58, I noticed that the default Makefile being copied around doesn't actually use the long list of CFLAGs. (It may also be a reason why the Travis tests are failing in that PR, so we probably need to fix this first.)

In the exercises I checked, enabling it broke the build due to the `-Werror` flag. So rather than just enabling it everywhere, I think we should first agree on which flags we should actually include, and then update the exercises 1 by 1 in case they need changes.

Here are the current flags in the file with a few of my own comments.
```
CFLAGS = -std=c99
CFLAGS += -Wall
CFLAGS += -Wextra
CFLAGS += -Werror
CFLAGS += -Wpointer-arith   # Enabled by pedantic
CFLAGS += -Wcast-align  # Might be useful(?) I don't know
CFLAGS += -Wwrite-strings  # Helps for enforcing const char*
CFLAGS += -Wswitch-default  # Maybe, it breaks some of my own code
CFLAGS += -Wunreachable-code  # Removed from GCC
CFLAGS += -Winit-self  # Good (C++ enables this in Wall, but not C)
CFLAGS += -Wmissing-field-initializers  # In -Wextra
CFLAGS += -Wno-unknown-pragmas  # May be good for allowing other compiler pragmas
CFLAGS += -Wstrict-prototypes  # Forces people to use foo(void) instead of foo(), potentially confusing
CFLAGS += -Wundef  # Probably don't need that
CFLAGS += -Wold-style-definition  # In -Wextra
```

I think we should include the following:
```
CFLAGS = -std=c99
CFLAGS += -Wall
CFLAGS += -Wextra
CFLAGS += -Wpedantic
CFLAGS += -Werror
```

And it might be a good idea to include some (or all) of these.
```
CFLAGS += -Wwrite-strings  # Helps for enforcing const char*
CFLAGS += -Winit-self  # Good (C++ enables this for Wall, but not C)
CFLAGS += -Wstrict-prototypes  # Forces people to use foo(void) instead of foo(), potentially confusing
```

What do you all think? I don't usually mess around with warnings outside of `-Wall -Wextra -Wpedantic`.

If you need to look up the warning options, [here are GCC's](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)

patricksjackson:
Adding the new makefile to every test (and fixing warnings from bob and leap).

I was going to touch each one individually, but it turned out there were only 2 very minor fixes needed to conform.

Closes #61

### DIFF
--- a/exercises/anagram/makefile
+++ b/exercises/anagram/makefile
@@ -1,18 +1,8 @@
-CFLAGS=-std=c99
+CFLAGS  = -std=c99
 CFLAGS += -Wall
 CFLAGS += -Wextra
+CFLAGS += -Wpedantic
 CFLAGS += -Werror
-CFLAGS += -Wpointer-arith
-CFLAGS += -Wcast-align
-CFLAGS += -Wwrite-strings
-CFLAGS += -Wswitch-default
-CFLAGS += -Wunreachable-code
-CFLAGS += -Winit-self
-CFLAGS += -Wmissing-field-initializers
-CFLAGS += -Wno-unknown-pragmas
-CFLAGS += -Wstrict-prototypes
-CFLAGS += -Wundef
-CFLAGS += -Wold-style-definition
 
 
 test: tests.out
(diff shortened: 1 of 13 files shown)


ryanplusplus:
:+1:

PR #70, opened 2016-09-09
Switch to standard Linux style

ryanplusplus:
This removes the modifications we were carrying against the standard Linux style for `indent`. This will make formatting appear correctly in GitHub's web view.

### DIFF
--- a/exercises/leap/src/example.c
+++ b/exercises/leap/src/example.c
@@ -3,14 +3,13 @@
 bool is_leap_year(int year)
 {
 	if (year % 4 == 0) {
-
 		if (year % 400 == 0) {
 			return true;
 		}
 		if (year % 100 == 0) {
 			return false;
 		}
-
+		return true;
 	}
 	return false;
 }
--- a/exercises/leap/src/example.c
+++ b/exercises/leap/src/example.c
@@ -2,15 +2,15 @@
 
 bool is_leap_year(int year)
 {
-   if (year % 4 == 0) {
+	if (year % 4 == 0) {
 
-      if (year % 400 == 0) {
-	 return true;
-      }
-      if (year % 100 == 0) {
-	 return false;
-      }
+		if (year % 400 == 0) {
+			return true;
+		}
+		if (year % 100 == 0) {
+			return false;
+		}
 
-   }
-   return false;
+	}
+	return false;
 }
--- a/exercises/leap/src/example.c
+++ b/exercises/leap/src/example.c
@@ -5,10 +5,10 @@ bool is_leap_year(int year)
    if (year % 4 == 0) {
 
       if (year % 400 == 0) {
-         return true;
+	 return true;
       }
       if (year % 100 == 0) {
-         return false;
+	 return false;
       }
 
    }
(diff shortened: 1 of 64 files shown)


ryanplusplus:
Looks like the actual difference is the specification of a tab size (8) and `-nut` (don't use tabs). The tabbing now looks weird and I don't fully understand why it's doing some of the things it's doing. I am planning to dig in a bit more if I can understand why things are working the way that they are, but I'm about to just give in and use the standard Linux kernel formatting settings (ie: let the indent be 8). Any thoughts or insights are welcome.

exercises/leap/src/example.c
```diff
@@ -5,10 +5,10 @@ bool is_leap_year(int year)
    if (year % 4 == 0) {
 
       if (year % 400 == 0) {
-         return true;
+	 return true;
```
ryanplusplus: I am struggling to understand how this could be indented only 2 spaces from the `if`. My current attempt at fixing things is very clearly wrong.

patricksjackson: I don't know why it looks wrong, but that line begins with a tab and a space which should be the correct indentation.

patricksjackson:
What exactly is the problem?

If it's just around the indent level, I think the only relevant flags are `-i8 -ts8` from the linux config.

In the thread about style I suggested replacing them with `-i3 -nut` as that forces the indent level to 3 and to not use any tabs. We could instead use `-i3 -ts3` and you'd just have to make sure you set your editor tab size to be 3 (which I will certainly forget to do).

ryanplusplus:
Thanks @patricksjackson. My problem is that I had my editor set to 3 space tabs, but I was viewing the changes in GitHub (which uses 8 space tabs) and in `git diff` (which also uses large tabs). I think the real issue is simply that the recommended long form and the short form (`indent -linux -i3`) are inconsistent. I think the real fix here is to just change the short form to `indent -linux -i3 -nut`. Thanks for the reality check.

ryanplusplus:
So... given that GitHub and a lot of other editors default to 8 space tabs, should we just use standard Linux kernel formatting to avoid seemingly incorrect rendering in the GitHub web view?

patricksjackson:
As long as we're consistent, and we can enforce it with a script, I don't have strong objections to any sane style choices. Since modifying it didn't seem to work that well I'm ok with pushing everything to use the standard linux kernel format to play nice with Github.

Maybe do the update after merging #50 so that @palisaide isn't stuck updating it forever :)

ryanplusplus:
Sounds good to me, will do that :)

patricksjackson:
Is this ready to merge? It looks fine to me.

ryanplusplus:
Yes, it's ready to merge. I was trying to wait for a good time when nothing else was pending, but I don't think that time is ever going to come.

patricksjackson:
Ah, ok. If between this and the indent script you have to break the style checker on master briefly I wouldn't worry about it.

palisaide:
Shall we go ahead and merge this?

ryanplusplus:
👍

ryanplusplus:
I think #119 more or less takes care of this

PR #75, opened 2016-09-14
Fix indent checking script on Travis

### LINKED ISSUE #73: Travis failed to catch an indentation issue
Indentation was incorrect in https://github.com/exercism/xc/pull/71 but `bin/verify-indent` indicated that there was no issue. When run locally `bin/verify-indent` _did_ indicate that `indent.sh` needed to be run.

ryanplusplus:
Closes #73

### DIFF
--- a/bin/verify-indent
+++ b/bin/verify-indent
@@ -1,6 +1,8 @@
-#!/bin/sh
+#!/bin/bash
 
 sh indent.sh
+
+git diff --quiet
 
 if ! $(git diff --quiet)
 then
--- a/bin/verify-indent
+++ b/bin/verify-indent
@@ -2,6 +2,8 @@
 
 sh indent.sh
 
+# This script fails when run by Travis if the diff is not checked twice
+# Unfortunately no better solution was found
 git diff --quiet
 
 if ! $(git diff --quiet)
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ bin/configlet
 bin/configlet.exe
 **/*.out
 **/*.o
+.DS_Store
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,4 @@ script:
 - bin/fetch-configlet
 - bin/configlet .
 - bin/verify-indent
+- bin/run-tests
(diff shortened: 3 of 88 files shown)


ryanplusplus:
I'm thoroughly confused.

I followed the [instructions for running a Travis Docker instance locally](https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally) and this _does_ work there.

Additionally, this _did_ work before (see https://travis-ci.org/exercism/xc/builds/157757693). Comparing the previous failure and a recent non-failure there appear to be no differences between the Travis image or the worker.

So... any ideas?

bin/verify-indent
```diff
@@ -1,7 +1,9 @@
-#!/bin/sh
+#!/bin/bash
 
 sh indent.sh
 
+git diff --quiet
```
ryanplusplus: Finally found a fix. I have no idea why repeating the command here makes it work. Any guesses?

patricksjackson:
Wow, that's odd. When I switched the check to `git diff-index --quiet HEAD` it worked fine every time (without the repeated call to git diff).

As for why it doesn't work, I don't really know. It's not a race condition (using `sleep` instead of the extra `git diff` doesn't help, and it will fail even if the tree is dirty before verify-indent is run). It's not specific to those particular files either. And most confusingly the original version works sometimes but not all the time. Just call `bin/verify-indent` several times repeatedly and it'll work randomly.

So yea, I have no idea either.

ryanplusplus:
@patricksjackson I tried your simpler solution, but it identifies a failure even when there are no changed files. Although I hate the current hacky solution, I'm not sure how to proceed toward a simpler solution, particularly when I don't know of a way to consistently reproduce the issue locally.

ryanplusplus:
@patricksjackson @palisaide are you guys okay with this change?

patricksjackson:
Well, I don't really like it, but it seems to work.

I say add a short comment in the script to explain why they're both there just in case some future contributor forgets to look at the history. After that I'm ok to merge it.

ryanplusplus:
@patricksjackson I added a comment

patricksjackson:
LGTM

ryanplusplus:
Updated with latest master

Issue #65, opened 2016-09-07
Anagram tests include a tremendous amount of duplication


ryanplusplus:
This basic block is included in most of the tests:

``` c
   int outputs_len = sizeof(outputs) / MAX_STR_LEN;
   qsort(outputs, outputs_len, sizeof(char) * MAX_STR_LEN, stringcmp);

   struct Vector vout = anagrams_for("Orchestra", vin);
   qsort(vout.vec, vout.size, sizeof(char) * MAX_STR_LEN, stringcmp);

   TEST_ASSERT_EQUAL(outputs_len, vout.size);
   int x;
   for (x = 0; x < vout.size; x++) {
      TEST_ASSERT_EQUAL(strlen(outputs[x]), strlen(vout.vec[x]));
      TEST_ASSERT_EQUAL_MEMORY(outputs[x], vout.vec[x], strlen(outputs[x]));
   }
```

We should probably refactor these tests to remove all of this noise and make it much more clear what is actually being checked.

PR #76, opened 2016-09-14
anagram: fixing strlen error

patricksjackson:
I believe this is the problem with the current anagram solution. For now I only spot fixed this one issue, but the rest of the problem may need to be looked at for improvements (#65 at least). If you guys think we should fix it up now I can go ahead, but this should let us get Travis up and running.

What I _think_ happened was the original `malloc` and `memcpy` threw away the null terminator. It would occasionally work if the next byte of memory was a null terminator and not otherwise. I should've just remembered to use valgrind, since that catches the overflow immediately.


ryanplusplus:
Good catch. I'm going to merge so we can get Travis going.

PR #78, opened 2016-09-22
binary-search: Adding bsearch problem.

patricksjackson:
The only weird thing with this problem is returning an 'error flag' in the case where the value isn't found. I use the length of the array but that's not necessarily the best option, so I'm definitely open to suggestions.

### DIFF
--- a/config.json
+++ b/config.json
@@ -18,7 +18,8 @@
     "beer-song",
     "rna-transcription",
     "difference-of-squares",
-    "sum-of-multiples"
+    "sum-of-multiples",
+    "binary-search"
   ],
   "exercises": [{
     "slug": "hello-world",
@@ -124,6 +125,15 @@
     "topics": [
       "control-flow (case statements)",
       "control-flow (loops)"
+    ]
+  }, {
+    "difficulty": 2,
+    "slug": "binary-search",
+    "topics": [
+      "arrays",
+      "control-flow (if-else statements)",
+      "control-flow (loops)",
+      "searching"
     ]
   }],
   "deprecated": [
--- a/exercises/binary-search/test/test_binary_search.c
+++ b/exercises/binary-search/test/test_binary_search.c
@@ -1,0 +1,91 @@
+#include "vendor/unity.h"
+#include "../src/binary_search.h"
+
+void test_single_element(void)
+{
+   int arr[] = { 6 };
+   size_t length = sizeof(arr) / sizeof(arr[0]);
+   TEST_ASSERT_EQUAL_UINT64(0, binary_search(6, arr, length));
+}
+
+void test_value_in_middle(void)
+{
+   int arr[] = { 1, 3, 4, 6, 8, 9, 11 };;
+   size_t length = sizeof(arr) / sizeof(arr[0]);
+   TEST_ASSERT_EQUAL_UINT64(3, binary_search(6, arr, length));
+}
+
+void test_value_at_beginning(void)
+{
+   int arr[] = { 1, 3, 4, 6, 8, 9, 11 };;
+   size_t length = sizeof(arr) / sizeof(arr[0]);
+   TEST_ASSERT_EQUAL_UINT64(0, binary_search(1, arr, length));
+}
+
+void test_value_at_end(void)
+{
+   int arr[] = { 1, 3, 4, 6, 8, 9, 11 };;
+   size_t length = sizeof(arr) / sizeof(arr[0]);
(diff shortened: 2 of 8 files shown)


config.json
```diff
@@ -110,6 +111,15 @@
       "control-flow (case statements)",
       "control-flow (loops)"
     ]
+  }, {
+	"difficulty": 2,
```
palisaide: Looks like the indentation is off here.

palisaide:
I'm not necessarily opposed to returning the array length as the "error flag," although I could see it causing some confusion.  I'm thinking that maybe the return type should just be `int` instead of `size_t` because it's really returning an index, not a size.  In that case, we could just make the "error flag" be -1 like it is in the json.

ryanplusplus:
@palisaide `size_t` is actually the correct type for indexing. As it can represent the size of any object, it's guaranteed to be big enough to index an array of any size whereas `int` is not.

exercises/binary-search/test/test_binary_search.c
```diff
@@ -0,0 +1,91 @@
...
+void test_value_larger_than_everything(void)
+{
+   int arr[] = { 1, 3, 4, 6, 8, 9, 11 };;
+   size_t length = sizeof(arr) / sizeof(arr[0]);
+   TEST_ASSERT_EQUAL_UINT64(length, binary_search(13, arr, length));
+}
+
+void test_empty_array(void)
+{
+   int *arr = NULL;
```
ryanplusplus: By using `NULL` here instead of passing a valid pointer but indicating that the length is 0, do we muddy the waters about whether they should check for `NULL` instead of the length?

exercises/binary-search/test/test_binary_search.c
```diff
@@ -0,0 +1,91 @@
...
+void test_single_element(void)
+{
+   int arr[] = { 6 };
+   size_t length = sizeof(arr) / sizeof(arr[0]);
+   TEST_ASSERT_EQUAL_UINT64(0, binary_search(6, arr, length));
+}
+
+void test_value_in_middle(void)
+{
+   int arr[] = { 1, 3, 4, 6, 8, 9, 11 };;
```
ryanplusplus: Is the `;;` a typo? It's repeated in many of the tests.

ryanplusplus:
Just a few nitpicks. Regarding using `length` as an error flag, it doesn't really bother me. I think your other options are introducing an output parameter for the location and then returning a success code (or some similar permutation). Using an output parameter in this case seems more awkward than simply using a special value for the error case.

exercises/binary-search/test/test_binary_search.c
patricksjackson: Yep that's a typo, it's on most of the tests due to copy/pasting. I'll remove it.

config.json
patricksjackson: Whoever added the rna-transcription accidentally inserted a tab for that line (and I copied theirs). I'll fix them.

patricksjackson:
Yea, I don't want to use `int` as the return value. I was between using either `length` or `SIZE_MAX` as the return value, but both of them are not that good since they're one off the ends of the array. I'm reconsidering going with `SIZE_MAX` as the error value, as I think it's a little better than length. At least it doesn't change between calls.

Actually, I see that the stdlib `bsearch` returns a pointer into the array if it found something, and `NULL` otherwise. That seems like a much better way to do it, and it could get people more comfortable with using pointers.

Any objection to switching it to use pointers?

exercises/binary-search/test/test_binary_search.c
patricksjackson: The way I see it, `bsearch` has a few different scenarios regarding `NULL` and whether or not it's actually valid input.

1. `ptr = NULL, length = 0` - **Valid** - Returns not found.
2. `ptr =~NULL, length = 0` - **Valid** - Returns not found.
3. `ptr = NULL, length = ~0` - **Invalid** - Undefined behaviour.

The second scenario I should probably add a test for, but I think we can call the 3rd undefined behaviour and not handle it.

ryanplusplus: Agreed.

patricksjackson:
Alright, I'll switch it over when I get the chance and remove the WIP label when it's ready to be reviewed again.

ryanplusplus:
LGTM, thanks for the cleanup!

palisaide:
LGTM 👍

ryanplusplus:
There are conflicts, but I will merge once they are resolved.

PR #79, opened 2016-09-26
Add difference-of-squares exercise

palisaide:
Adds the 'difference-of-squares' exercise

### DIFF
--- a/exercises/difference-of-squares/src/example.h
+++ b/exercises/difference-of-squares/src/example.h
@@ -3,6 +3,6 @@
 
 unsigned int sum_of_squares(unsigned int number);
 unsigned int square_of_sum(unsigned int number);
-unsigned int difference(unsigned int number);
+unsigned int difference_of_squares(unsigned int number);
 
 #endif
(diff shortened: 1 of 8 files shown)


patricksjackson:
Everything looks good to me :+1:

exercises/difference-of-squares/src/example.h
```diff
@@ -0,0 +1,8 @@
+#ifndef _DIFFERENCE_OF_SQUARES_H
+#define _DIFFERENCE_OF_SQUARES_H
+
+unsigned int sum_of_squares(unsigned int number);
+unsigned int square_of_sum(unsigned int number);
+unsigned int difference(unsigned int number);
```
ryanplusplus: Given that the exercise is "difference of squares", why not name this function `difference_of_squares`?

ryanplusplus:
Just one nitpick, otherwise LGTM.

exercises/difference-of-squares/src/example.h
palisaide: I was trying to follow other tracks, most of which use `difference`.  But in our case, since we don't have a `Squares` class, I agree that `difference_of_squares` is better.

palisaide:
Ready for another look

PR #80, opened 2016-09-26
Add sum-of-multiples exercise

palisaide:
Adds 'sum-of-multiples' exercise

### DIFF
--- a/config.json
+++ b/config.json
@@ -16,7 +16,8 @@
     "pangram",
     "nth-prime",
     "beer-song",
-    "rna-transcription"
+    "rna-transcription",
+    "sum-of-multiples"
   ],
   "exercises": [{
     "slug": "hello-world",
@@ -101,12 +102,19 @@
       "strings",
       "control-flow (if-else statements)",
       "control-flow (loops)"
-	]
+    ]
   }, {
-	"difficulty": 2,
+    "difficulty": 2,
     "slug": "rna-transcription",
     "topics": [
       "strings",
+      "control-flow (case statements)",
+      "control-flow (loops)"
+    ]
+  }, {
+    "difficulty": 4,
+    "slug": "sum-of-multiples",
+    "topics": [
       "control-flow (case statements)",
       "control-flow (loops)"
     ]
--- a/exercises/sum-of-multiples/src/example.c
+++ b/exercises/sum-of-multiples/src/example.c
@@ -1,22 +1,16 @@
 #include "sum_of_multiples.h"
 #include <stdbool.h>
 
-unsigned int sum_of_multiples_to(const unsigned int *multiples, const size_t number_of_multiples, const int upper_bound)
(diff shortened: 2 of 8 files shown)


config.json
```diff
@@ -101,15 +102,22 @@
       "strings",
       "control-flow (if-else statements)",
       "control-flow (loops)"
-	]
```
palisaide: Some tabs were introduced here at some point.  I converted them to spaces.

exercises/sum-of-multiples/src/example.c
```diff
@@ -0,0 +1,24 @@
+#include "sum_of_multiples.h"
+#include <stdbool.h>
+
+unsigned int sum_of_multiples_to(const unsigned int *multiples, const size_t number_of_multiples, const int upper_bound)
```
patricksjackson: `upper_bound` and `*multiples` should have the same type. Same goes for `current` in the loop below.

exercises/sum-of-multiples/src/example.h
```diff
@@ -0,0 +1,8 @@
+#ifndef _SUM_OF_MULTIPLES_H
+#define _SUM_OF_MULTIPLES_H
+
+#include <stddef.h>
+
+unsigned int sum_of_multiples_to(const unsigned int *multiples, const size_t number_of_multiples, const int up_to);
```
patricksjackson: Is there a reason this isn't just named `sum_of_multiples`?

exercises/sum-of-multiples/src/example.c
```diff
@@ -0,0 +1,24 @@
...
+         if ((current % multiples[i]) == 0) {
+            is_multiple = true;
+            break;
+         }
+      }
+
+      if (is_multiple) {
+         sum += current;
+      }
+   }
```
ryanplusplus: Instead of using `is_multiple`, could you just move `sum += current` on line 13?

ryanplusplus:
Looks good outside of a few small nitpicks

exercises/sum-of-multiples/src/example.c
palisaide: Haha, yep.  Not sure why I didn't see that.

exercises/sum-of-multiples/src/example.c
palisaide: Right, I'll make them all unsigned.

exercises/sum-of-multiples/src/example.h
palisaide: I thought `sum_of_multples_to` would be more descriptive.  But now that I read it, I seems excessive.  I'll change it to `sum_of_multiples`

palisaide:
Ready for another look!

ryanplusplus:
LGTM (but needs to have master merged)

PR #81, opened 2016-10-01
Add largest-series-product

ryanplusplus:
See https://github.com/exercism/x-common/blob/master/exercises/largest-series-product/canonical-data.json


exercises/largest-series-product/src/example.c
```diff
@@ -0,0 +1,54 @@
...
+      return -1;
+
+   if (!span && !digit_count)
+      return 1;
+
+   for (size_t i = 0; i < digit_count; i++) {
+      if (!is_digit(digits[i]))
+         return -1;
+
+      int to_add = to_int(digits[i]);
```
palisaide: You could also use the stdlib `atoi()` here, which has the added benefit of returning 0 for any non-numeric char passed.

ryanplusplus: There are two reasons that I opted to go this route:
- `'0'` also returns `0` so you can't use this alone to determine whether an invalid character was provided.
- `atoi` wants a null-terminated string and not a single character. This means we'd have to construct a null-terminated string on the fly in order to use it.

palisaide:
LGTM 👍

exercises/largest-series-product/test/test_largest_series_product.c
```diff
@@ -0,0 +1,114 @@
...
+void test_can_get_the_largest_product_of_a_big_number_ii(void)
+{
+   TEST_ASSERT_EQUAL(28350,
+                     largest_series_product
+                     ("52677741234314237566414902593461595376319419139427", 6));
+}
+
+void test_can_get_the_largest_product_of_a_big_number_project_euler(void)
+{
+   TEST_ASSERT_EQUAL(23514624000,
```
patricksjackson: `TEST_ASSERT_EQUAL` is a macro define for `TEST_ASSERT_EQUAL_INT`, which may overflow here. It would still pass the tests since it just casts them both to an int. (I think it uses 64bit ints by default on systems that it detects have 64bit pointers, but I think it's safer to explicitly use 64bit types where we know we need it.)

exercises/largest-series-product/src/example.h
```diff
@@ -0,0 +1,8 @@
+#ifndef _LARGEST_SERIES_PRODUCT_H
+#define _LARGEST_SERIES_PRODUCT_H
+
+#include <stddef.h>
+
+long largest_series_product(char *digits, size_t span);
```
patricksjackson: Isn't `long` usually 32bit? Looks like we need to use `int64_t` here for the return type.

ryanplusplus: Good call. Replaced with `int64_t`.

exercises/largest-series-product/test/test_largest_series_product.c
ryanplusplus: Good catch. Replaced with `TEST_ASSERT_EQUAL_INT64`.

patricksjackson:
LGTM 👍

PR #82, opened 2016-10-06
Add 'roman-numerals' exercise

palisaide:
See https://github.com/exercism/x-common/tree/master/exercises/roman-numerals

### DIFF
--- a/config.json
+++ b/config.json
@@ -19,7 +19,8 @@
     "rna-transcription",
     "difference-of-squares",
     "sum-of-multiples",
-    "binary-search"
+    "binary-search",
+    "roman-numerals"
   ],
   "exercises": [{
     "slug": "hello-world",
@@ -134,6 +135,15 @@
       "control-flow (if-else statements)",
       "control-flow (loops)",
       "searching"
+    ]
+  }, {
+    "difficulty": 3,
+    "slug": "roman-numerals",
+    "topics": [
+      "arrays",
+      "memory management",
+      "structs",
+      "control-flow (loops)"
     ]
   }],
   "deprecated": [
--- a/config.json
+++ b/config.json
@@ -143,7 +143,8 @@
       "arrays",
       "memory management",
       "structs",
-      "control-flow (loops)"
+      "control-flow (loops)",
+      "strings"
     ]
   }],
   "deprecated": [
--- a/exercises/roman-numerals/src/example.c
+++ b/exercises/roman-numerals/src/example.c
@@ -1,0 +1,43 @@
+#include "roman_numerals.h"
+#include <stdlib.h>
+#include <string.h>
+
+#define NUM_OF_ELEMENTS(a)    (sizeof(a) / sizeof(a[0]))
+#define MAX_NUMERAL_LENGTH    (6)
+
+typedef struct
+{
+    char *numeral;
+    unsigned int value;
+} Numeral_Values_t;
+
+const Numeral_Values_t numeral_values[] =
+{
+    { "M", 1000 },
+    { "CM", 900 },
+    { "D",  500 },
+    { "CD", 400 },
+    { "C",  100 },
+    { "XC",  90 },
+    { "L",   50 },
+    { "XL",  40 },
+    { "X",   10 },
+    { "IX",   9 },
+    { "V",    5 },
+    { "IV",   4 },
+    { "I",    1 }
+};
+
+char * to_roman_numeral(unsigned int number)
+{
+   char *numerals = calloc(0, sizeof(char) * MAX_NUMERAL_LENGTH);
+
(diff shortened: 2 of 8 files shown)


config.json
```diff
@@ -135,6 +136,15 @@
...
       "searching"
     ]
+  }, {
+    "difficulty": 3,
+    "slug": "roman-numerals",
+    "topics": [
+      "arrays",
+      "memory management",
+      "structs",
+      "control-flow (loops)"
```
ryanplusplus: How about `"strings"`?

exercises/roman-numerals/src/example.c
```diff
@@ -0,0 +1,43 @@
...
+#include <string.h>
+
+#define NUM_OF_ELEMENTS(a)    (sizeof(a) / sizeof(a[0]))
+#define MAX_NUMERAL_LENGTH    (6)
+
+typedef struct
+{
+    char *numeral;
+    unsigned int value;
+} Numeral_Values_t;
```
ryanplusplus: Serious nitpick: how about `numeral_values_t` or `numeralvalues_t` to match the rest of the style conventions?

exercises/roman-numerals/src/example.c
```diff
@@ -0,0 +1,43 @@
...
+    { "I",    1 }
+};
+
+char * to_roman_numeral(unsigned int number)
+{
+   char *numerals = calloc(0, sizeof(char) * MAX_NUMERAL_LENGTH);
+
+   for (size_t i = 0; i < NUM_OF_ELEMENTS(numeral_values); i++) {
+      while (number >= numeral_values[i].value) {
+         strcat(numerals, numeral_values[i].numeral);
```
ryanplusplus: Given that you have a hard limit for the length, how about `strncat`?

exercises/roman-numerals/src/example.c
```diff
@@ -0,0 +1,43 @@
+#include "roman_numerals.h"
+#include <stdlib.h>
+#include <string.h>
+
+#define NUM_OF_ELEMENTS(a)    (sizeof(a) / sizeof(a[0]))
+#define MAX_NUMERAL_LENGTH    (6)
```
ryanplusplus: I think this needs to be 7 in order to account for the null terminator

ryanplusplus:
I tried to use the new review feature. Hopefully I didn't screw anything up.

exercises/roman-numerals/src/example.c
patricksjackson: Isn't it more than that? "MMMDCCCLXXXVIII" is the longest valid answer I can think of at the moment.

exercises/roman-numerals/src/example.c
patricksjackson: I don't think we've defined our own types in our style conventions. My usual strategy is to use Caps for user types to distinguish it from library types, so I have a slight preference for `Numeral_Values_t` or even `NumeralValues_t` in this case.

It obviously isn't really important to me either way.

exercises/roman-numerals/src/example.c
ryanplusplus: Right, the 6/7 only covers what is tested

config.json
palisaide: Good idea

exercises/roman-numerals/src/example.c
palisaide: Good call on the null terminator.  And yes, this is just to cover the test cases.

exercises/roman-numerals/src/example.c
palisaide: Changed to strncat although I'm questioning my calculation of the length.  Let me know what you guys think.

palisaide:
Ready for another look

exercises/roman-numerals/src/example.c
```diff
@@ -0,0 +1,43 @@
...
+    { "I",    1 }
+};
+
+char * to_roman_numeral(unsigned int number)
+{
+   char *numerals = calloc(0, sizeof(char) * MAX_NUMERAL_LENGTH);
+
+   for (size_t i = 0; i < NUM_OF_ELEMENTS(numeral_values); i++) {
+      while (number >= numeral_values[i].value) {
+         strncat(numerals, numeral_values[i].numeral, MAX_NUMERAL_LENGTH - strlen(numerals));
```
ryanplusplus: I'm fairly certain that this calculation is correct. `strlen(numerals)` will give you the number of characters prior to the null terminator and because `MAX_NUMERAL_LENGTH` _does_ include the null terminator then `strncat` will be able to insert all characters plus a null terminator.

ryanplusplus:
LGTM

PR #83, opened 2016-10-14
Add 'allergies' exercise

palisaide:
See https://github.com/exercism/x-common/tree/master/exercises/allergies

### DIFF
--- a/exercises/allergies/src/example.c
+++ b/exercises/allergies/src/example.c
@@ -1,0 +1,33 @@
+#include "allergies.h"
+#include <stdlib.h>
+
+static const int scores[] =
+{
+   1,
+   2,
+   4,
+   8,
+   16,
+   32,
+   64,
+   128
+};
+
+bool is_allergic_to(Allergen_t allergen, int score)
+{
+   return ((score & scores[allergen]) == scores[allergen]);
+}
+
+void get_allergens(int score, Allergen_List_t *list)
+{
+   Allergen_t *allergens = calloc(0, sizeof(Allergen_t) * Allergen_Last);
+   list->allergens = allergens;
+
+   for (Allergen_t allergen = Allergen_First; allergen <= Allergen_Last; allergen++) {
+      if (is_allergic_to(allergen, score)) {
+         score -= scores[allergen];
+         list->allergens[list->count] = allergen;
+         list->count++;
+      }
+   }
+}
--- a/exercises/allergies/src/example.c
+++ b/exercises/allergies/src/example.c
@@ -20,8 +20,7 @@ bool is_allergic_to(Allergen_t allergen, int score)
 
 void get_allergens(int score, Allergen_List_t *list)
 {
-   Allergen_t *allergens = calloc(0, sizeof(Allergen_t) * Allergen_Last);
-   list->allergens = allergens;
+   list->allergens = calloc(0, sizeof(Allergen_t) * Allergen_Last);
 
    for (Allergen_t allergen = Allergen_First; allergen <= Allergen_Last; allergen++) {
       if (is_allergic_to(allergen, score)) {
--- a/exercises/allergies/src/example.c
+++ b/exercises/allergies/src/example.c
@@ -17,7 +17,7 @@ bool is_allergic_to(Allergen_t allergen, unsigned int score)
    return ((score & scores[allergen]) == scores[allergen]);
 }
 
-void get_allergens(int score, Allergen_List_t * list)
+void get_allergens(unsigned int score, Allergen_List_t * list)
 {
    list->allergens = calloc(Allergen_Count, sizeof(Allergen_t));
    list->count = 0;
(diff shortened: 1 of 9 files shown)


palisaide:
Wow... I have no idea why every test is failing on the Travis build.  They all pass on my machine.

exercises/allergies/src/example.c
```diff
@@ -0,0 +1,32 @@
...
+void get_allergens(int score, Allergen_List_t * list)
+{
+   list->allergens = calloc(0, sizeof(Allergen_t) * Allergen_Last);
+
+   for (Allergen_t allergen = Allergen_First; allergen <= Allergen_Last;
+        allergen++) {
+      if (is_allergic_to(allergen, score)) {
+         score -= scores[allergen];
+         list->allergens[list->count] = allergen;
+         list->count++;
```
ryanplusplus: Looks like this guy is never initialized to 0

exercises/allergies/src/example.c
```diff
@@ -0,0 +1,32 @@
...
+};
+
+bool is_allergic_to(Allergen_t allergen, int score)
+{
+   return ((score & scores[allergen]) == scores[allergen]);
+}
+
+void get_allergens(int score, Allergen_List_t * list)
+{
+   list->allergens = calloc(0, sizeof(Allergen_t) * Allergen_Last);
```
ryanplusplus: `Allergen_Last` is 7, but there are actually 8 allergens.

I think it will make things a bit simpler if the allergen enum is rearranged like this:

```c
enum {
  Allergen_Eggs,
  Allergen_Peanuts,
  ...
  Allergen_Count
}
```

Then you can rearrange your `for` to use `allergen < Allergen_Count` as the termination condition and use `sizeof(Allergen_t) * Allergen_Count` as your allocation size.

exercises/allergies/src/example.h
```diff
@@ -0,0 +1,27 @@
...
+   Allergen_Tomatoes,
+   Allergen_Chocolate,
+   Allergen_Pollen,
+   Allergen_Cats,
+   Allergen_Last = Allergen_Cats
+} Allergen_t;
+
+typedef struct {
+   int count;
+   Allergen_t *allergens;
```
ryanplusplus: Given that we know exactly how many allergens there are prior to declaring this type, why wouldn't we simply create a properly sized array for `allergens` instead of requiring a memory allocation?

exercises/allergies/src/example.h
```diff
@@ -0,0 +1,27 @@
...
+} Allergen_List_t;
+
+bool is_allergic_to(Allergen_t allergen, int score);
```
ryanplusplus: Shouldn't the score be an unsigned value? If you're treating this as a bitfield it will be safer and less confusing to have it be an unsigned.

exercises/allergies/src/example.c
ryanplusplus: Additionally, the signature for `calloc` is `calloc(<number of items>, <item size>)` so I think this should be `calloc(Allergen_Last + 1, sizeof(Allergen_t))` (this is probably the reason that all of the tests fail on Travis).

exercises/allergies/src/example.h
ryanplusplus: I guess leaving this as an allocation means that the implementation could potentially allocate only as many as are actually necessary.

exercises/allergies/src/example.c
palisaide: Good call.  I'm sure that's why the tests are failing.

exercises/allergies/src/example.c
palisaide: Yep.  I originally had `get_allergens()` mallocing the entire `Allergen_List_t` and then switched to having it passed in.  Will fix.

exercises/allergies/src/example.h
palisaide: Yep.  Will fix.

exercises/allergies/src/example.h
palisaide: I could go either way.  I do like having the option allocate only as many as necessary.

palisaide:
Alright, tests are passing now.

exercises/allergies/src/example.h
```diff
@@ -0,0 +1,26 @@
...
+   Allergen_Count
+} Allergen_t;
+
+typedef struct {
+   int count;
+   Allergen_t *allergens;
+} Allergen_List_t;
+
+bool is_allergic_to(Allergen_t allergen, unsigned int score);
+void get_allergens(int score, Allergen_List_t * list);
```
ryanplusplus: Can you update this `score` to be unsigned as well?

ryanplusplus:
Looks good once `score` is updated to be an `unsigned`

palisaide:
Oops, missed that one... fixed

ryanplusplus:
LGTM

PR #86, opened 2016-10-28
Add 'phone-number' exercise

palisaide:
See the exercise description [here](https://github.com/exercism/x-common/blob/master/exercises/phone-number/description.md)
Test cases were copied from the [C++ track](https://github.com/exercism/xcpp/blob/master/phone-number/phone_number_test.cpp) as there is no canonical-data.json for this exercise.
- [x] Set up makefile, test file, and implementation files
- [x] Add tests
- [x] Add example

### DIFF
--- a/exercises/phone-number/src/example.c
+++ b/exercises/phone-number/src/example.c
@@ -1,0 +1,1 @@
+#include "phone_number.h"
--- a/exercises/phone-number/src/example.c
+++ b/exercises/phone-number/src/example.c
@@ -55,9 +55,12 @@ char *phone_number_get_area_code(const char *input)
 
 char *phone_number_format(const char *input)
 {
+   char *cleaned_input = phone_number_clean(input);
    char *output = calloc(FORMATTED_LENGTH, sizeof(char));
 
-   sprintf(output, "(%.3s) %.3s-%.4s", input, &input[AREA_CODE_LENGTH], &input[EXTENSION_OFFSET]);
+   sprintf(output, "(%.3s) %.3s-%.4s", cleaned_input, &cleaned_input[AREA_CODE_LENGTH], &cleaned_input[EXTENSION_OFFSET]);
+
+   free(cleaned_input);
 
    return output;
 }
--- a/exercises/phone-number/src/example.c
+++ b/exercises/phone-number/src/example.c
@@ -1,1 +1,70 @@
 #include "phone_number.h"
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+
+#define AREA_CODE_LENGTH         (3)
+#define EXCHANGE_LENGTH          (3)
+#define EXTENSION_LENGTH         (4)
+#define VALID_NUMBER_LENGTH      AREA_CODE_LENGTH + EXCHANGE_LENGTH + EXTENSION_LENGTH
+#define FORMATTED_LENGTH         VALID_NUMBER_LENGTH + 4
+
+#define INVALID_NUMBER_RESULT    "0000000000"
+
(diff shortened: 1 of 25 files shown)


palisaide:
I'm not particularly happy with the implementation of `phone_number_format()`, so I'm open to any suggestions on how to make that one cleaner.

exercises/phone-number/test/test_phone_number.c
```diff
@@ -0,0 +1,94 @@
+#include "vendor/unity.h"
+#include "../src/phone_number.h"
+#include <stdlib.h>
+
+void test_phone_number(const char *input, const char *expected)
```
ryanplusplus: Should this be `test_phone_number_clean`?

exercises/phone-number/src/example.c
```diff
@@ -0,0 +1,70 @@
...
+      }
+   }
+
+   return output;
+}
+
+char *phone_number_get_area_code(const char *input)
+{
+   char *output = calloc(AREA_CODE_LENGTH, sizeof(char));
+   strncpy(output, phone_number_clean(input), AREA_CODE_LENGTH);
```
ryanplusplus: A simple alternative would be to just set `phone_number_clean(input)[4] = '\0'` and return that. It uses a little bit of extra memory, but the client can still `free` it like normal.

exercises/phone-number/src/example.c
```diff
@@ -0,0 +1,70 @@
...
+   *cursor++ = ' ';
+   memcpy(cursor, &input[AREA_CODE_LENGTH], EXCHANGE_LENGTH);
+   cursor += EXCHANGE_LENGTH;
+   *cursor++ = '-';
+   memcpy(cursor, &input[AREA_CODE_LENGTH + EXCHANGE_LENGTH], EXTENSION_LENGTH);
+   cursor += EXTENSION_LENGTH;
+   *cursor++ = '\0';
+
+   return output;
+}
```
ryanplusplus: I'm fairly certain you can use `sprintf` for this:

```c
sprintf(output, "(%.3s) %.3s-%.4s", input, input[AREA_CODE_LENGTH], input[AREA_CODE_LENGTH + EXCHANGE_LENGTH);
```

ryanplusplus:
Sorry, I merged some of your other PRs and created a conflict

exercises/phone-number/src/example.c
```diff
@@ -0,0 +1,70 @@
...
+
+char *phone_number_get_area_code(const char *input)
+{
+   char *output = calloc(AREA_CODE_LENGTH, sizeof(char));
+   strncpy(output, phone_number_clean(input), AREA_CODE_LENGTH);
+
+   return output;
+}
+
+char *phone_number_format(const char *input)
```
ryanplusplus: `phone_number_get_area_code` uses `phone_number_clean` internally, but this doesn't. I think we should add a test that requires this for consistency.

patricksjackson:
I don't see anything other than what @ryanplusplus already noted. The rest LGTM 👍

exercises/phone-number/src/example.c
palisaide: Good idea!

exercises/phone-number/src/example.c
palisaide: Yes, much cleaner

exercises/phone-number/src/example.c
palisaide: I agree.  I'll add a test for this.

palisaide:
Ready for another look

ryanplusplus:
Changes look good to me 👍

PR #90, opened 2016-11-04
Clock Exercise

StevenRoot:
Added.Ds_Store to gitignore
Implemented the clock exercise.

Note the introduction of X Macro's to generate some of the test routines.  Only as an introduction.  It may be too complicated for a user to modify but I think it is a small enough introduction for users to appreciate the elimination of the duplication.  For these endless test cases -- it could have been applied more.

### DIFF
--- a/exercises/clock/src/clock.h
+++ b/exercises/clock/src/clock.h
@@ -1,0 +1,9 @@
+#ifndef _CLOCK_H
+#define _CLOCK_H
+
+#define MAX_STR_LEN 5
+
+void clock(char time_text[MAX_STR_LEN], int hour, int minute);
+void clock_add(char time_text[MAX_STR_LEN], int minute_offset);
+
+#endif
--- a/exercises/clock/src/clock.h
+++ b/exercises/clock/src/clock.h
@@ -1,7 +1,7 @@
 #ifndef _CLOCK_H
 #define _CLOCK_H
 
-#define MAX_STR_LEN (5 +1) // "##:##\0"
+#define MAX_STR_LEN (5 + 1) // "##:##\0"
 
 typedef char time_text_t[MAX_STR_LEN];
 
--- a/exercises/clock/src/clock.h
+++ b/exercises/clock/src/clock.h
@@ -6,6 +6,6 @@
 typedef char time_text_t[MAX_STR_LEN];
 
 void clock(time_text_t time_text, int hour, int minute);
-void clock_add(char time_text[MAX_STR_LEN], int minute_offset);
+void clock_add(time_text_t time_text, int minute_offset);
 
 #endif
--- a/exercises/clock/src/clock.h
+++ b/exercises/clock/src/clock.h
@@ -3,7 +3,9 @@
 
 #define MAX_STR_LEN 5
 
-void clock(char time_text[MAX_STR_LEN], int hour, int minute);
+typedef char time_text_t[MAX_STR_LEN];
+
+void clock(time_text_t time_text, int hour, int minute);
 void clock_add(char time_text[MAX_STR_LEN], int minute_offset);
 
 #endif
--- a/exercises/clock/src/clock.h
+++ b/exercises/clock/src/clock.h
@@ -1,7 +1,7 @@
 #ifndef _CLOCK_H
 #define _CLOCK_H
 
-#define MAX_STR_LEN 5
+#define MAX_STR_LEN (5 +1) // "##:##\0"
 
 typedef char time_text_t[MAX_STR_LEN];
 
--- a/exercises/clock/test/test_clock.c
+++ b/exercises/clock/test/test_clock.c
@@ -5,18 +5,18 @@
 #include "vendor/unity.h"
 #include "../src/clock.h"
 
-char actual_time_text[MAX_STR_LEN];
-char actual_time_text2[MAX_STR_LEN];
+time_text_t actual_time_text;
+time_text_t actual_time_text2;
 
-// general routines that tests expected result
-void test_expected_value(char *expected_clock_text, int hour, int minute)
+// general routines that test expected results
+void test_expected_value(const char *expected_clock_text, int hour, int minute)
 {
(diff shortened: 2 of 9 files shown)


exercises/clock/test/test_clock.c
```diff
@@ -0,0 +1,489 @@
...
+   if (expected_equality) {
+      TEST_ASSERT_EQUAL_STRING(actual_time_text, actual_time_text2);
+   } else {
+      bool result = (0 == strcmp(actual_time_text, actual_time_text2));
+
+      TEST_ASSERT_EQUAL(0, result);
+   }
+}
+
+// the Tests
```
ryanplusplus: inconsistent Capitalization

exercises/clock/test/test_clock.c
```diff
@@ -0,0 +1,489 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+#include "vendor/unity.h"
+#include "../src/clock.h"
+
+char actual_time_text[MAX_STR_LEN];
+char actual_time_text2[MAX_STR_LEN];
+
+// general routines that tests expected result
```
ryanplusplus: Small grammar error here

exercises/clock/test/test_clock.c
```diff
@@ -0,0 +1,489 @@
...
+void test_expected_value(char *expected_clock_text, int hour, int minute)
+{
+   memset(actual_time_text, '\0', sizeof(actual_time_text));
```
ryanplusplus: Should this be here? Shouldn't the implementation be responsible for null-terminating the string?

exercises/clock/test/test_clock.c
```diff
@@ -0,0 +1,489 @@
...
+#include <string.h>
+
+#include "vendor/unity.h"
+#include "../src/clock.h"
+
+char actual_time_text[MAX_STR_LEN];
+char actual_time_text2[MAX_STR_LEN];
+
+// general routines that tests expected result
+void test_expected_value(char *expected_clock_text, int hour, int minute)
```
ryanplusplus: How about dropping a `const` on `expected_clock_text` since it's passed by reference and you do not modify the value?

exercises/clock/test/test_clock.c
```diff
@@ -0,0 +1,489 @@
...
+   int minute_offset = -3000;
+   char *expected_clock_text = "00:20";
+
+   test_time_add_expected_value(expected_clock_text, hour, minute,
+                                minute_offset);
+}
+
+// these tests check equality.
+// Part of standard test collection
+// Included even though they don't make much sense in C
```
ryanplusplus: Agreed, these don't make much sense in C since a "clock" is just a string and I think we fairly conclusively proved these above.

However, they _could_ make sense if we were to allocate clock objects and provide a clock object comparison.

I'm okay if we don't refactor to add clock objects, but we should probably remove these if we don't since their inclusion is puzzling for the C implementation.

exercises/clock/src/clock.h
```diff
@@ -0,0 +1,9 @@
+#ifndef _CLOCK_H
+#define _CLOCK_H
+
+#define MAX_STR_LEN 5
+
+void clock(char time_text[MAX_STR_LEN], int hour, int minute);
```
ryanplusplus: Would it make sense to typedef `char time_text[MAX_STR_LEN]` to `clock_t` or similar?

exercises/clock/test/test_clock.c
```diff
@@ -0,0 +1,489 @@
...
+   test_time_expected_equality(hour1, minute1, hour2, minute2, expected_equality);\
+   }\
+
+// below expands to multiple routines...that are all designed to test the equality of 2 different
+// clock representations to see if they are the same or not
+// that 2 clocks are equal...
+// the fields are:
+//    test name, time1_hour, time1_minute, time2_hour, time2_minute, (time1 == time2)
+
+#define EQUALITY_TESTS(ENTRY) \
```
ryanplusplus: These guys are failing in the CI environment

StevenRoot: will comment them and see if CI passes...  then we can decide whether to leave them as tests or not...

exercises/clock/test/test_clock.c
```diff
@@ -0,0 +1,554 @@
...
+   RUN_TEST(test_add_more_than_two_days);
+   RUN_TEST(test_subtract_minutes);
+   RUN_TEST(test_subtract_to_previous_hour);
+   RUN_TEST(test_subtract_more_than_an_hour);
+   RUN_TEST(test_subtract_across_midnight);
+   RUN_TEST(test_subtract_more_than_two_hours);
+   RUN_TEST(test_subtract_more_than_two_hours_with_borrow);
+   RUN_TEST(test_subtract_more_than_one_day);
+   RUN_TEST(test_subtract_more_than_two_days);
+
```
StevenRoot: split out the tests we might get rid of...

exercises/clock/test/test_clock.c
```diff
@@ -0,0 +1,554 @@
...
+//   ENTRY(test_clocks_with_negative_minute_that_wraps,  4,  10,   5, -1490, true)\
+//   ENTRY(test_clocks_with_negative_minute_that_wraps_multiple_times,  6,  15,   6, -4305, true)\
+//   ENTRY(test_clocks_with_negative_hours_and_minutes,  7,  32,   -12, -268, true)\
+//   ENTRY(test_clocks_with_negative_hours_and_minutes_that_wrap,  18,  7,   -54, -11513, true)\
+//
+//
+//EQUALITY_TESTS(EXPAND_EQUALITY_TEST_AS_TEST)
+
+
+//Xmacro expanded...
```
StevenRoot: manually expanded all of the x macros and it works..

I have not tried the xmacros in ubuntu yet to see why it may not work in linux...

StevenRoot:
I think everything passes in both mac and linux enviro now.

palisaide:
LGTM 👍

PR #92, opened 2016-11-07
Make run-test script fail if compilation has errors

### LINKED ISSUE #91: run-tests script does not exit with failure if there's a compilation error
Currently, the `run-tests` script does not catch a compiler error, only test failures.  We need to make the script return non-zero if there is a compiler error so that Travis considers the build a failure.

palisaide:
Closes #91

### DIFF
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -26,6 +26,8 @@ for D in exercises/*; do
           make clean >> /dev/null;
           if make | grep FAIL: ; then
             exit 1
+          elif [ ${PIPESTATUS[0]} -ne 0 ]; then
+            exit 1
           fi
           cd ${CURRENT_DIR};
         }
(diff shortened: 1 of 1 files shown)


bin/run-tests
```diff
@@ -26,6 +26,8 @@ for D in exercises/*; do
           make clean >> /dev/null;
           if make | grep FAIL: ; then
             exit 1
+          elif [ ${PIPESTATUS[0]} -ne 0 ]; then
```
ryanplusplus: I assume this works, but couldn't you simply add `set -e` to the top of this script so it fails if anything goes wrong?

palisaide: I tried that a while ago and I could never get it to actually work.  I think I just figured out why (from the Bash Reference Manual):
> The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement ...

In this case, it's part of an `if` statement, so it doesn't exit.  Although it couldn't hurt to add `set -e` to catch any other unforeseen errors.

PR #93, opened 2016-11-15
sieve exercise

StevenRoot:
still providing sieve.h since I am allocating the results array.  Otherwise it is pretty straight forward

### DIFF
--- a/config.json
+++ b/config.json
@@ -172,7 +172,7 @@
       "control-flow (if-else statements)"
     ]
   }, {
-     "difficulty": 1,
+     "difficulty": 2,
      "slug": "clock",
      "topics": [
         "control-flow (if-statements)",
@@ -181,7 +181,7 @@
         "preprocessor (x-macros in test)"
      ]
   }, {
-     "difficulty": 1,
+     "difficulty": 2,
      "slug": "sieve",
      "topics": [
         "control-flow (if-statements)",
--- a/config.json
+++ b/config.json
@@ -185,7 +185,7 @@
      "slug": "sieve",
      "topics": [
         "control-flow (if-statements)",
-        "memory management",
+        "memory management"
      ]
   }],
   "deprecated": [
--- a/config.json
+++ b/config.json
@@ -180,6 +180,13 @@
         "text formatting",
         "preprocessor (x-macros in test)"
      ]
+  }, {
+     "difficulty": 1,
+     "slug": "sieve",
+     "topics": [
+        "control-flow (if-statements)",
+        "memory management",
+     ]
   }],
   "deprecated": [
 
--- a/exercises/sieve/src/example.c
+++ b/exercises/sieve/src/example.c
@@ -16,27 +16,28 @@ unsigned int sieve(const unsigned int limit, primesArray_t primes)
       numberIsPrime = malloc(limit + 1);
       memset(numberIsPrime, 1, limit + 1);
 
+      unsigned int maxFactor = sqrt(limit) + 1;
+
       // mark 0 and 1 as not prime
       for (unsigned int i = 0; i < 2; i++) {
          numberIsPrime[i] = 0;
       }
 
       // mark the remaining numbers in the array according to the algo.
-      for (unsigned int i = 2; i < sqrt(limit) + 1; i++) {
+      for (unsigned int index = 2; index < maxFactor;) {
          // mark all of multiples that can't be prime
-         for (unsigned int nonPrimeIndex = (2 * i); nonPrimeIndex < (limit + 1);
-              nonPrimeIndex += i) {
+         for (unsigned int nonPrimeIndex = (2 * index);
+              nonPrimeIndex < (limit + 1); nonPrimeIndex += index) {
             numberIsPrime[nonPrimeIndex] = 0;
          }
 
-         // adjust the index i - normally bad practice within a loop but ok here as always making equivalent or larger
-         for (unsigned int newStartIndex = i + 1;
-              newStartIndex < sqrt(limit) + 1; newStartIndex++) {
-            if (numberIsPrime[newStartIndex]) {
-               i = newStartIndex - 1;   // need to account for the ++ of the outer for loop
+         // adjust the index
+         do {
+            index++;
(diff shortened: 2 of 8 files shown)


StevenRoot:
ok how to I make sure the math library gets into the build -- no linker switches and I need a -lm right?

config.json
```diff
@@ -180,6 +180,13 @@
         "text formatting",
         "preprocessor (x-macros in test)"
      ]
+  }, {
+     "difficulty": 1,
```
ryanplusplus: Should this be a 1? Hello world is a 1 so I want to say this should be at least a 2 or 3.

exercises/sieve/src/sieve.h
```diff
@@ -0,0 +1,10 @@
+#ifndef _SIEVE_H
+#define _SIEVE_H
+
+#define MAX_LIMIT_TESTED (1000)
+
+typedef unsigned int primesArray_t[MAX_LIMIT_TESTED];
```
ryanplusplus: It would make the testing more annoying, but it seems like a callback solution would be more natural and efficient

exercises/sieve/src/example.c
```diff
@@ -0,0 +1,40 @@
...
+#include <math.h>
+#include "sieve.h"
+
+unsigned int sieve(const unsigned int limit, primesArray_t primes)
+{
+   unsigned int numberOfPrimes = 0;
+   unsigned char *numberIsPrime;
+
+   // clear the results
+   memset(primes, 0, MAX_LIMIT_TESTED);
```
ryanplusplus: How about `sizeof(primes)` instead of `MAX_LIMIT_TESTED`?

exercises/sieve/test/test_sieve.c
```diff
@@ -0,0 +1,114 @@
+#include <string.h>
+#include "vendor/unity.h"
+#include "../src/sieve.h"
+
+primesArray_t result_array;
```
ryanplusplus: Mark this fella `static`

exercises/sieve/test/test_sieve.c
```diff
@@ -0,0 +1,114 @@
...
+
+// Individual Tests
```
ryanplusplus: This seems out of place since there are no non-individual tests (I think...)

exercises/sieve/test/test_sieve.c
```diff
@@ -0,0 +1,114 @@
...
+void test_no_primes_under_two(void)
+{
+   const unsigned int limit = 1;
+   const primesArray_t expected_prime_array = { 0 };
```
ryanplusplus: Shouldn't this really be `{}`?

exercises/sieve/src/sieve.h
ryanplusplus: Eh, screw it. That's not the point of the exercise.

config.json
StevenRoot: I think it is the same as clock -- I will make them both a 2
 -- ratings are tough - you have to leave room for the Russian or American skater that is up last...  someone has to get a 10

exercises/sieve/src/example.c
StevenRoot: I can't because it is treating primes as a simple pointer at this point...

exercises/sieve/src/example.c
```diff
@@ -0,0 +1,40 @@
...
+              nonPrimeIndex += i) {
+            numberIsPrime[nonPrimeIndex] = 0;
+         }
+      }
+      // collect and count the primes found
+      for (unsigned int i = 1; i < limit + 1; i++) {
+         if (numberIsPrime[i]) {
+            primes[numberOfPrimes++] = i;
+         }
+      }
```
StevenRoot: I don't understand the comment...   the method here follows the algorithm 
you can pick up a little efficiency if you check at line 26 if the number is considered prime -- and add that to the result array there...  but after you are done you would still have to iterate from sqrt(limit) to the limit to check the remaining numbers that might still be considered prime.   
as the 2 outer loops would still be required  -- and the complexity is greater...  I would like to leave as is...

exercises/sieve/test/test_sieve.c
StevenRoot: agreed - done

exercises/sieve/test/test_sieve.c
StevenRoot: I like to separate the "Tests" from the helper functions...  since there are no helper functions in this case I can delete the comment

exercises/sieve/test/test_sieve.c
StevenRoot: you can't do that...  it has to have an entry or no initializer...   since I want to init the expected values and leave the rest as 0 to support memcmp  -- this was the easiest way as it will make everything 0

exercises/sieve/src/example.c
ryanplusplus: You can do `sizeof(*primes)`

exercises/sieve/src/example.c
ryanplusplus: You're not following the algorithm, see https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes. Specifically, see step 4:

> Find the first number greater than p in the list that is not marked. If there was no such number, stop. Otherwise, let p now equal this new number (which is the next prime), and repeat from step 3.

exercises/sieve/test/test_sieve.c
ryanplusplus: Whoops, you're right. I forgot that you need to explicitly initialize one member to get the rest to be implicitly initialized.

exercises/sieve/src/example.c
StevenRoot: agreed - modified...  but leaving the collection the same as I think that is easier..
still squeamish about adjusting an index of a for loop within the for loop (bad practice!)  but I think it meets the algo intent better and is safe here..

ryanplusplus:
Looks good to me

palisaide:
LGTM 👍

PR #101, opened 2017-01-04
Refactoring anagram test cases.

### LINKED ISSUE #65: Anagram tests include a tremendous amount of duplication

patricksjackson:
I cleaned up the test cases so they're more consistent. I could reduce the LOC by ~90 lines by using a macro to get rid of most of the boilerplate, but I felt that adding another layer of indirection would be too confusing.

Closes #65

### DIFF
--- a/exercises/anagram/test/test_anagram.c
+++ b/exercises/anagram/test/test_anagram.c
@@ -4,9 +4,33 @@
 #include "vendor/unity.h"
 #include "../src/anagram.h"
 
+// Empty vector for when expected output is nothing
+struct Vector empty = {
+   NULL,
+   0
+};
+
 int stringcmp(const void *a, const void *b)
 {
    return strcmp(a, b);
+}
+
+// Helper function to verify the correct output
+void check_anagrams(char *in, struct Vector vin, struct Vector expected)
+{
+   struct Vector vout = anagrams_for(in, vin);
+
+   // Sort actual and expected outputs for consistency
+   qsort(expected.vec, expected.size, sizeof(char) * MAX_STR_LEN, stringcmp);
+   qsort(vout.vec, vout.size, sizeof(char) * MAX_STR_LEN, stringcmp);
+
+   TEST_ASSERT_EQUAL(expected.size, vout.size);
+   for (int x = 0; x < vout.size; x++) {
+      TEST_ASSERT_EQUAL(strlen(expected.vec[x]), strlen(vout.vec[x]));
+      TEST_ASSERT_EQUAL_MEMORY(expected.vec[x], vout.vec[x],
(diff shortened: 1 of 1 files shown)


exercises/anagram/test/test_anagram.c
```diff
@@ -173,62 +168,47 @@ void test_case_insensitive_anagrams(void)
...
+   };
+
    struct Vector vin = {
       inputs,
       sizeof(inputs) / MAX_STR_LEN
    };
 
-   char outputs[][MAX_STR_LEN] = {
-      "ΒΓΑ",
-      "γβα"
```
patricksjackson: One of the outputs on this test shouldn't have been here, so I removed it. It was never hit in automation because the test is 'extra credit' and commented out.

exercises/anagram/test/test_anagram.c
```diff
@@ -4,11 +4,35 @@
...
+   0
+};
+
 int stringcmp(const void *a, const void *b)
 {
    return strcmp(a, b);
 }
 
+// Helper function to verify the correct output
+void check_anagrams(char *in, struct Vector vin, struct Vector expected)
```
ryanplusplus: I just realized that nothing in any of our tests is marked `static`. This function in particular has a name that might come up in someone's solution and get the linker confused. How about making at least the helper functions and helper function data (ie: `empty`) `static`?

ryanplusplus:
Very nice cleanup 👍

patricksjackson:
Good call on using static, there _shouldn't_ be very many collisions, but it may be worthwhile making those are in future test suites.

I did go back and change my refactoring a bit so that the helper function is just an assertion on the equality of two Vector objects. It makes each test case more clear about what's actually going on as opposed to adding a layer of indirection.

ryanplusplus:
👍

PR #104, opened 2017-01-08
Use legal include guard names

### LINKED ISSUE #102: #pragma once is more better than include guard
if we use modern c compiler, use #pragma once instead of old style #ifndef -- #endif .
macro names starting with one or two underscores, such as _HELLO_WORLD_H and __HELLO_WORLD_H, are reserved to the implementation and should not be used by the user.

````c
#pragma once

void hello(char *buffer, const char *name);

````

````c
#ifndef _HELLO_WORLD_H
#define _HELLO_WORLD_H

void hello(char *buffer, const char *name);

#endif
````

### PR DESCRIPTION
**Pull request by ryanplusplus**
Fixes #102

### DIFF
--- a/exercises/rna-transcription/src/example.h
+++ b/exercises/rna-transcription/src/example.h
@@ -1,5 +1,5 @@
 #ifndef  _RNA_TRANSCRIPTION_H
-#define _RNA_TRANSCRIPTION_H
+#define RNA_TRANSCRIPTION_H
 
 char *to_rna(const char *dna);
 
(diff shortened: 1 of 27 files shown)

### CODE REVIEW

File: exercises/rna-transcription/src/example.h [hunk 1]
```diff
@@ -1,5 +1,5 @@
 #ifndef  _RNA_TRANSCRIPTION_H
```
patricksjackson: Missed one

patricksjackson:
I checked out the PR and used grep to confirm that was the only one left, so it looks good to merge after that change.

exercises/rna-transcription/src/example.h
ryanplusplus: Ugh. I this has 2 spaces and so evaded my find/replace.

ryanplusplus:
@patricksjackson all fixed

PR #105, opened 2017-01-13
Added full track documentation

### LINKED ISSUE #100: Verify contents and format of track documentation
Each language track has documentation in the `docs/` directory, which gets included on the site
on each track-specific set of pages under [/languages](http://exercism.io/languages).

We've added some general guidelines about how we'd like the track to be documented in exercism/exercism.io#3315
which can be found at https://github.com/exercism/exercism.io/blob/master/docs/writing-track-documentation.md

Please take a moment to look through the documentation about documentation, and make sure that
the track is following these guidelines. Pay particularly close attention to **how to use images**
in the markdown files.

Lastly, if you find that the guidelines are confusing or missing important details, then a pull request
would be greatly appreciated.

ryanplusplus:
This is heavily influenced by the documentation at `xcpp` so it's probably missing some good resources, but it's far better than the mostly nothing that we have now.

Closes #100

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -15,13 +15,13 @@ build system as separate command-line tools.
 Linux users will need gcc or clang for the compiler, and `make` will be needed
 for the native build. Make is pre-installed on most unix systems, but is most
 likely available via your package manager. On Debian and Ubuntu based
-distributions it can be installed with `sudo apt-get install make` if not
-present.
+distributions everything you need can be installed with `sudo apt-get install build-essential`
+if not present.
 
 
 #### MacOS
 
-MacOS users can install gcc or clangwith [Homebrew](http://brew.sh/) via
+MacOS users can install gcc or clang with [Homebrew](http://brew.sh/) via
 `brew install gcc` or `brew install llvm`.
 
 
@@ -32,4 +32,3 @@ that will give you the Visual Studio 2013 IDE and the latest version of the
 Microsoft Visual C compiler.
 
 [vs-1013]: http://www.visualstudio.com/en-us/products/visual-studio-community-vs
-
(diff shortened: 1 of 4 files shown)


docs/INSTALLATION.md
```diff
@@ -32,4 +32,3 @@ that will give you the Visual Studio 2013 IDE and the latest version of the
 Microsoft Visual C compiler.
 
 [vs-1013]: http://www.visualstudio.com/en-us/products/visual-studio-community-vs
```
patricksjackson: Can we change this to not reference a specific version of Visual Studio? Just point the link to visualstudio.com (since I think the link keeps changing) and say that Microsoft offers a free version of the Visual Studio IDE that comes with everything you need (compiler/etc).

Actually I have no idea how easy it is to use VS for these Makefile projects, do you know?

patricksjackson:
Do these docs also need to explain the basics of using the exercises? (e.g. You'll need `src/hello_world.h/c` and then you can run `make` to execute all the tests?)

Everything looks great though. The resources and explanations are a superset of what I would've probably put 👍

docs/INSTALLATION.md
ryanplusplus: Good deal, I'll switch it over. Honestly I have no idea how much of a nightmare this is on Windows as I avoid it whenever possible :)

ryanplusplus:
@patricksjackson I think the documentation you're referring to is [here](https://github.com/exercism/xc/blob/master/docs/TESTS.md). Some documentation was already present and I didn't touch that part so it didn't show up here.

patricksjackson:
Ah, that would be it thanks.

LGTM 👍

palisaide:
Looks good!

PR #107, opened 2017-01-22
Remove spurious semi-colon

wolf99:
There was an extra, second, semi-colon at the end of an ASSERT line. This probably causes no harm but is nevertheless incorrect. This commit removes it.


patricksjackson:
Good catch, thanks :)

PR #108, opened 2017-01-23
Add format string to printf in test_your_birthday

wolf99:
Without this, using the code as-is, an error is given upon compilation:

~~~sh
$ make test
Compiling tests.out
test/test_gigasecond.c: In function ‘test_your_birthday’:
test/test_gigasecond.c:58:4: error: format not a string literal and no format arguments [-Werror=format-security]
    printf(ctime(&gigday));
    ^~~~~~
cc1: all warnings being treated as errors
make: *** [tests.out] Error 1
~~~


palisaide:
Looks good.

patricksjackson:
Ah, that's the danger of having commented out code I guess. Thanks for the fix @wolf99

PR #110, opened 2017-01-25
Add tests for invalid squares

wolf99:
A chess board does not have a zeroth square or more than 64 squares - solutions should not attempt to calculate these cases. If they did what return type would they use for calculating the grains on, for example, the 100th square? (Hint: there isn't a simple C data type large enough!)


wolf99:
The CI build fails of course, as the implementation being checked against probably doesn't conform to the new checks!

PR #112, opened 2017-01-25
triangle exercise added

DeathSec:


wolf99:
You have made changes to so many unrelated files? Even the test vendor files? Maybe your git line-endign configuration is not set properly.

Also, you _**deleted**_ the hello_world.c? I think that's supposed to be there! (For helping newbies get started with the exercism format).

PR #113, opened 2017-01-25
Add test for invalid indexes

wolf99:
A chess board does not have a zeroth square or more than 64 squares.


wolf99:
Cant figure out what it is that verify-indent is failing for :-\

PR #114, opened 2017-01-25
Add test for invalid indexes

wolf99:
A chess board does not have a zeroth square or more than 64 squares

### DIFF
--- a/exercises/grains/src/example.c
+++ b/exercises/grains/src/example.c
@@ -4,6 +4,10 @@
 
 uint64_t square(uint8_t index)
 {
+   if ((!index) || (index > NUMBER_OF_SQUARES)) {
+      return 0;
+   }
+
    return (uint64_t) 1 << (index - 1);
 }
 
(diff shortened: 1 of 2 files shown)


wolf99:
Ugh, I give up - I tried using indent.sh and it adds in a load of `^M`(#113) and CI fails. I manually format to the spec given in the indent command in the readme and it _still_ fails for indent.

It would be handy if the CI could ouptut _why_ the `verify-indent` fails maybe

wolf99:
BTW, it would be quite handy if the verify-indent tool in CI could output *why* it fails when it does.

I finally got this to pass indent, but I still dont know what Ive done differently this time, LOL

palisaide:
LGTM 👍

wolf99:
I note that the [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/grains/canonical-data.json) for this exercise in x-common also specifies that negative number squares be rejected (causes an exception, to be precise). The unsigned data types used in the current test and example implementation naturally preclude negative values being used.

Is use of the unsigned data types enough or should a test with a negative square number be introduced and all the types used be changed to signed?

patricksjackson:
I think how we have it now is fine even though it presents a few oddities.

Thanks for the fix.

PR #115, opened 2017-01-25
binary exercise added

DeathSec:


DeathSec:
@wolf99 Why this track is still not launched , this repo has more than 10 problems. ??

wolf99:
@DeathSec Don't know TBH. some of the official group people might (https://github.com/orgs/exercism/people). 

Although I'd imagine it's because exercism is small project (admin-wise) and the one person that is trying to work on it full time is probably rather busy.

DeathSec:
@patricksjackson can you merge this to the repo?

exercises/binary/test/test_binary.c
```diff
@@ -0,0 +1,85 @@
+#include "vendor/unity.h"
+#include "../src/binary.h"
+
+void b1_is_decimal_1(void)
```
palisaide: Can you prefix these guys with `test_` to better indicate that they're the actual tests and not just helper functions?

exercises/binary/test/test_binary.c
```diff
@@ -0,0 +1,85 @@
...
+   int result = convert("10001101000");
+   TEST_ASSERT_EQUAL(1128, result);
+}
+
+void invalid_entry_is_decimal_0(void)
+{
+   int result = convert("invalid entry");
+   TEST_ASSERT_EQUAL(0, result);
+}
+
```
palisaide: There are a few more tests in the [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/binary/canonical-data.json) that could be added as well.

exercises/binary/src/example.c
```diff
@@ -0,0 +1,23 @@
+int len(char *str)
+{
+   int count = 0;
+   while (*str) {
+      str++, count++;
+   }
+   return count;
+}
+
+int convert(char input[64])
```
palisaide: What's the significance of the `64` here.  Couldn't we just use `char *input`?

exercises/binary/src/example.c
```diff
@@ -0,0 +1,23 @@
...
+{
+   int lens = len(input);
```
palisaide: How about using `strlen()` here?

DeathSec: using first principles!

exercises/binary/src/example.c
DeathSec: my bad

DeathSec:
@palisaide i will add some more test from canonical-data.json , and i changed the array to pointer  in the 'convert' function , now it looks good

exercises/binary/src/example.c
DeathSec: using first principles means no use of standard libs , strlen is in 'string.h'

exercises/binary/test/test_binary.c
DeathSec: test_ ? like ....

typedef void test_;

test_ b10_is_decimal_2(){ } ... ?

exercises/binary/test/test_binary.c
```diff
@@ -0,0 +1,99 @@
...
+void test_a_number_with_internal_non_binary_characters_is_invalid(void)
+{
+   TEST_ASSERT_EQUAL(INVALID, convert("10nope10"));
+}
+
+void test_a_number_and_a_word_whitespace_spearated_is_invalid(void)
+{
+   TEST_ASSERT_EQUAL(INVALID, convert("001 100"));
```
patricksjackson: This test is a bit different than the canonical tests. The canonical tests are ambiguous with respect to whether whitespace alone invalidates the entry. (The corresponding test case is "001 nope")

I'm fine with it, just adding a note here for future reference.

exercises/binary/test/test_binary.c
```diff
@@ -0,0 +1,99 @@
+#include "vendor/unity.h"
+#include "../src/binary.h"
+
+void test_b0_is_decimal_0(void)
+{
+   TEST_ASSERT(0 == convert("0"));
```
patricksjackson: These two tests use `TEST_ASSERT` and the rest use `TEST_ASSERT_EQUAL`. Please standardize on one of them.

exercises/binary/src/example.h
```diff
@@ -0,0 +1,7 @@
+#ifndef BINARY_H
+#define BINARY_H
+#define INVALID -1
+
+int convert(char *input);
```
patricksjackson: Please use `const char *input` instead to clarify that the pointer data won't be modified.

exercises/binary/src/example.c
```diff
@@ -0,0 +1,23 @@
...
+   int count = 0;
+   while (*str++) {
+      count++;
+   }
+   return count;
+}
+
+int convert(char *input)
+{
+   int input_len = len(input);
```
patricksjackson: You should use the standard lib function `strlen()` instead of writing your own.

That also means that the variables `input_length` and `i` need to be of type `size_t`

exercises/binary/src/example.c
```diff
@@ -0,0 +1,23 @@
+int len(char *str)
```
patricksjackson: Please add the needed includes at the top of this file.

```
#include "binary.h"
#include <string.h"
```

patricksjackson:
Those nitpicks are just to make sure we keep a consistent style across the repo. It otherwise looks great.

exercises/binary/src/example.c
DeathSec: The Problem is "Write a program that will convert a binary number, represented as a string (e.g. '101010'), to its decimal equivalent using first principles "
                                                                           ^^^^^^^^^^^^
First Principles ==> If your language provides a method in the standard library to perform the conversion, pretend it doesn't exist and implement it yourself.

exercises/binary/test/test_binary.c
DeathSec: my bad

exercises/binary/src/example.h
DeathSec: my bad , i will check on it

exercises/binary/test/test_binary.c
wolf99: I'd imagine @palisaide meant `void test_b1_is_decimal_1(void)`, i.e. prefix test function names with `test_`, as this matches the convention used with the test framework in use and all other test files in xc.

exercises/binary/src/example.c
wolf99: I could be wrong but I thought the principle of not using provided functions referred to the conversion stated in the problem, not any incidental functions called in order to implement that solution? 
I.e. using functions from `string.h` etc would be OK, but if `string.h` provided a `strtobin()` function then using that would not be.

DeathSec: @wolf99  it is a part of my conversion so i thought i should implement myself , its not my interpretation it is given in exercism official website?..  , Anyways i will add strlen() in the next commit..

palisaide:
LGTM 👍

Issue #116, opened 2017-01-28
rna-transcription test results in segmentation fault


wolf99:
In the rna-transcription exercise's test file, the `test_transcription()` test helper function tries to use `strncmp()` and `free()` on the result of the UUT. But, following the TDD rules, at the start of the TDD process the UUT does not allocate any memory - this means that the project is un-buildable until at least some not insignificant portion of the implementation is already achieved, thus it breaks the TDD process that is supposed to be being fostered (according to the description of the exercise at least).

The helper function should be altered so as not to prevent early iterations of the implementation from being built. This could be as simple as adding in a check for `NULL`. e.g:

~~~c
void test_transcription(const char *dna, const char *expected)
{
   char *rna = to_rna(dna);
   if (rna != NULL) {
      TEST_ASSERT_TRUE(strncmp(rna, expected, strlen(dna)) == 0);
      free(rna);
   } else {
      TEST_FAIL();
   }
}
~~~

patricksjackson:
Well, the project can build at least, it will just segfault unhelpfully. I think that's a good idea for a fix, but if you add it please make sure there is at least a brief comment to explain it.

On another note on this piece of code, that `strncmp` call may be wrong. I think the length argument should instead be "`strlen(expected) + 1`" so that it could check if the returned `rna` value has extra garbage characters at the end.

wolf99:
Fixed by #117

PR #117, opened 2017-01-29
Prevent segfault early on in TDD cycle per issue #116

wolf99:
Also adjust strlen argument as mentioned by @patricksjackson, also in issue #116

### DIFF
--- a/exercises/rna-transcription/test/test_rna_transcription.c
+++ b/exercises/rna-transcription/test/test_rna_transcription.c
@@ -6,8 +6,8 @@
 void test_transcription(const char *dna, const char *expected)
 {
    char *rna = to_rna(dna);
-   if (rna != NULL) {           /* Prevent segfault on strlen() and free(). */
-      TEST_ASSERT_TRUE(strncmp(rna, expected, (strlen(expected) + 1)) == 0);
+   if (rna != NULL) {           /* Prevent segfault on string comparison and free(). */
+      TEST_ASSERT_EQUAL_STRING(expected, rna);
       free(rna);
    } else {
       TEST_FAIL();              /* to_rna() returned NULL when it was not epected. */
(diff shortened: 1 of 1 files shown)


wolf99:
Fixes #116

wolf99:
I haven't tested this out but it seems that the options used in `indent.sh` differ from the Linux style, not only by `-i3` but also by `-nut`, which may be where the difference is caused. If this *is* the case I would suggest changing either the rule in indent.sh, or the suggested command in README.md that uses the `-linux` option.

patricksjackson:
LGTM :+1:

I'll merge it myself in a few days if the other maintainers don't weigh in by then. My only mild concern is getting someone to confirm that the new `strncmp` argument is a good idea.

palisaide:
I'm questioning whether the `+ 1` will actually do anything.  The documentation for `strncmp()` states that it will terminate whenever a null terminator is reached in either string.  So, wouldn't it just stop as soon as it gets to the end of `dna`?

Otherwise, looks good to me.

patricksjackson:
@palisaide The point of that extra 1 is to make sure that `rna` is the same length as `expected`. e.g. if `rna = "ABC"` and `expected = "AB"` then without the +1 it wouldn't find the difference. (I just confirmed that this is the case).

I now realize though that since we're already relying on both `dna` and `expected` to be valid NULL-terminated strings, replacing the `strncmp` with just a call to `strcmp` is probably better anyways. It's safe as long as we know one of the strings is valid.

wolf99:
@patricksjackson & @palisaide, please expand my commit message for f5ba2f6 above to see reasoning about further changes I have made.

ryanplusplus:
@palisaide @patricksjackson this looks good to me. Is there any reason not to merge?

PR #119, opened 2017-01-31
Add no tabs specification to text and shell command

### LINKED ISSUE #118: indent.sh does not match indent -linux -i3
The `README.md` suggests use of `$ indent -linux -i3`. this does not match what is produced by `indent.sh` however. as the CI uses `indent.sh` to verify the indent on files then this causes problems for those that read the readme and simply ran the `indent` command as suggested.

This can be seen in commit a5d8645 on which I ran `$ indent -linux -i3 $(file)`. This found a small braces spacing issue and corrected it. Thinking all was good I commited, pushed and made the PR to start the CI. But the CI reported that `verify-indent` failed! To see what the issue was I tried running `$ indent.sh $(file)` which made another, *different*, spacing change. This committed as 1f32a18 and subsequently passed the CI test.

I haven't tested this out but it seems that the options used in indent.sh differ from the Linux style (as specified by man indent), not only by `-i3`, but also by `-nut`, which may possibly be where the difference is caused (I'm not really familiar with indent so mainly guessing here). If this is the case I would suggest changing either the rule in indent.sh, or the suggested command in README.md that uses the -linux option. Obviously the latter would have the least effect on the existing code, but I am unsure which of the styles was the intended one.

wolf99:
Fixes #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,14 @@ Please see the [Exercism contributing guide](https://github.com/exercism/x-commo
 
 ## Coding Style
 
-All test and example code should be written using the [Linux kernel coding style](https://www.kernel.org/doc/Documentation/CodingStyle) with 3 space indents.
+All test and example code should be written using the [Linux kernel coding style](https://www.kernel.org/doc/Documentation/CodingStyle) with 3 space indents and no tabs.
 
 If GNU `indent` is installed, this can be run on all files by executing `indent.sh`.
 
 To run it on a single file, you can execute:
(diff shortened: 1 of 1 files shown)


wolf99:
Realised this may contrast to #70 , but that PR seems to have died?

patricksjackson:
This looks good. I wouldn't worry about #70, this documents what we are doing _now_, if that changes we can make the necessary changes later.

PR #120, opened 2017-02-03
Update README.md to include details of example solutions

### LINKED ISSUE #103: Add an explanation of example.c/h to README
The main contributing guide explains a little about how example solutions work across the Exercism tracks, but the usage here of example.c, example.h, and the provided headers should be explicitly explained in our README.

The same patch would likely apply to the exercism/xcpp repo as well.

palisaide:
Closes #103 .  Tried to keep it brief, but maybe we need more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ indent -nbad -bap -nbc -bbo -hnl -br -brs -c33 -cd33 -ncdb -ce -ci4 -cli0 -d0 -d
 
 See the [GNU `indent` manual](https://www.gnu.org/software/indent/manual/indent.html#SEC4) for more information.
 
+## Example Solutions
+
+As stated in the [contributing guide](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#track-anatomy), each exercise should have an example solution that passes all of the given tests.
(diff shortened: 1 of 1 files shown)


wolf99:
Maybe could add a small bit abut the directory structure used, as this is slightly different from that outlined in the referenced anatomy section of the README?

patricksjackson:
Three extra details I think need to be included.

1. Note the reason for this is that files with `example` in the name are skipped when downloading to the client. It's not all that clear in the x-common guide.
2. References to the file names in the code always point to the expected name of the files and not `example.c/h`. (e.g. Include guards and statements)
3. The naming scheme means the checked in code is only able to compile using the `bin\run-tests` script, which will rename the files correctly.

wolf99:
Should this have been protected from the probot as part of #8 ?

ryanplusplus:
@wolf99 I think we can keep the issue open, but I don't want to merge this without @patricksjackson's comments addressed and it has been several months. So, I think we should keep this PR closed, but if you want to pick up @palisaide's changes and open a new PR I'd be happy to review and merge.

PR #121, opened 2017-02-03
triangle exercise added

DeathSec:
i don't know if this satisfies the [https://github.com/exercism/x-common/blob/master/exercises/triangle/canonical-data.json](canonical-data.json)

### DIFF
--- a/config.json
+++ b/config.json
@@ -218,6 +218,10 @@
      ]
   }, {
      "difficulty": 1,
+     "slug": "triangle",
+     "topics": []
+  }, {
+     "difficulty": 1,
      "slug": "binary",
      "topics": [
         "control-flow (loops)",
--- a/config.json
+++ b/config.json
@@ -220,10 +220,10 @@
      "difficulty": 1,
      "slug": "triangle",
      "topics": [
-		"control-flow (if-else statements)",
-		"Booleans",
-		"Structs",
-		"Mathematics"
+	"control-flow (if-else statements)",
+	"Booleans",
+	"Structs",
+	"Mathematics"
 	       ]
   }, {
      "difficulty": 1,
--- a/config.json
+++ b/config.json
@@ -219,7 +219,12 @@
   }, {
      "difficulty": 1,
      "slug": "triangle",
-     "topics": []
+     "topics": [
+		"control-flow (if-else statements)",
+		"Booleans",
+		"Structs",
+		"Mathematics"
+	       ]
   }, {
      "difficulty": 1,
      "slug": "binary",
--- a/exercises/triangle/src/example.c
+++ b/exercises/triangle/src/example.c
@@ -1,0 +1,23 @@
+#include "triangle.h"
+
+int triangle_equality(double a, double b, double c)
+{
+   return (a < (b + c) && b < (a + c) && c < (a + b));
+}
+
+int kind(double a, double b, double c)
+{
+   if (a == 0 || b == 0 || c == 0) {
+      return error;
+   }
+   if (a < 0 || b < 0 || c < 0 || !triangle_equality(a, b, c)) {
+      return error;
+   }
+   if (a == b && b == c) {
+      return equilateral;
+   }
+   if (a == b || b == c || a == c) {
+      return isosceles;
+   }
+   return scalene;
+}
--- a/exercises/triangle/src/example.c
+++ b/exercises/triangle/src/example.c
@@ -5,7 +5,7 @@ int triangle_equality(double a, double b, double c)
    return (a < (b + c) && b < (a + c) && c < (a + b));
 }
 
-flavor kind(double a, double b, double c)
+enum flavor kind(double a, double b, double c)
 {
    if (a == 0 || b == 0 || c == 0) {
       return invalid;
--- a/exercises/triangle/src/example.c
+++ b/exercises/triangle/src/example.c
@@ -21,10 +21,9 @@ bool is_equilateral(Triangle_t * input)
 
 bool is_isosceles(Triangle_t * input)
 {
-   return (triangle_equality(input) &&
-           ((input->point_a == input->point_b)
-            || (input->point_b == input->point_c)
-            || (input->point_a == input->point_c))
+   return (triangle_equality(input) && ((input->point_a == input->point_b)
+                                        || (input->point_b == input->point_c)
+                                        || (input->point_a == input->point_c))
        );
 }
 
--- a/exercises/triangle/src/example.c
+++ b/exercises/triangle/src/example.c
@@ -2,28 +2,25 @@
 
 bool triangle_equality(Triangle_t * input)
 {
-   return (input->point_a > 0
-           && input->point_b > 0
-           && input->point_c > 0
-           && input->point_a < (input->point_b + input->point_c)
-           && input->point_b < (input->point_a + input->point_c)
-           && input->point_c < (input->point_a + input->point_b)
+   return (input->a < (input->b + input->c)
+           && input->b < (input->a + input->c)
+           && input->c < (input->a + input->b)
        );
 }
 
 bool is_equilateral(Triangle_t * input)
 {
    return (triangle_equality(input)
-           && (input->point_a == input->point_b)
-           && (input->point_b == input->point_c)
+           && (input->a == input->b)
+           && (input->b == input->c)
        );
 }
 
 bool is_isosceles(Triangle_t * input)
 {
-   return (triangle_equality(input) && ((input->point_a == input->point_b)
-                                        || (input->point_b == input->point_c)
-                                        || (input->point_a == input->point_c))
+   return (triangle_equality(input) && ((input->a == input->b)
+                                        || (input->b == input->c)
+                                        || (input->a == input->c))
        );
 }
 
--- a/exercises/triangle/src/example.c
+++ b/exercises/triangle/src/example.c
@@ -1,23 +1,37 @@
 #include "triangle.h"
 
-int triangle_equality(double a, double b, double c)
+bool triangle_equality(Triangle_t * input)
 {
-   return (a < (b + c) && b < (a + c) && c < (a + b));
+   return (input->point_a > 0
+           && input->point_b > 0
+           && input->point_c > 0
+           && input->point_a < (input->point_b + input->point_c)
+           && input->point_b < (input->point_a + input->point_c)
+           && input->point_c < (input->point_a + input->point_b)
+       );
 }
 
-enum flavor kind(double a, double b, double c)
+bool is_equilateral(Triangle_t * input)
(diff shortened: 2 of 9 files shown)


wolf99:
You have defined an enum for the return values but then defined the functions as returning integers... was this intentional?

wolf99:
With regard to the tests; Instead of doing
~~~C
   TEST_ASSERT(expected == foo());
~~~
It would be more semantically correct to do 
~~~C
   TEST_ASSERT_EQUAL_INT(expected, foo()); // EQUAL for the relation and INT for the underlying type
~~~

DeathSec:
my bad , that was not intentional..

wolf99:
Cool 👍 LGTM. Though I don't know either way about satisfying the canonical-data spec. Some of the team members may have to check and confirm that.

config.json
```diff
@@ -218,6 +218,10 @@
      ]
   }, {
      "difficulty": 1,
+     "slug": "triangle",
+     "topics": []
```
patricksjackson: At a minimum, topics needs to include `control flow (if-else statements)`. There may be others that apply as well, see the list https://github.com/exercism/x-common/blob/master/TOPICS.txt

exercises/triangle/src/example.c
```diff
@@ -0,0 +1,23 @@
+#include "triangle.h"
+
+int triangle_equality(double a, double b, double c)
```
patricksjackson: This function should be declared as `static int triangle_equality`. We're trying to be more consistent with that :)

exercises/triangle/src/example.h
```diff
@@ -0,0 +1,13 @@
+#ifndef TRIANGLE_H
```
patricksjackson: Since we're defining enums here to use in the test cases, you need to rename the file to `triangle.h` so that this file is included in the download. See the Allergies exercise and #120. (Also, as a relatively new contributor, any input you may have on what we need to explain in #120 would be helpful).

exercises/triangle/src/example.h
```diff
@@ -0,0 +1,13 @@
...
+#define TRIANGLE_H
+
+enum flavor {
```
patricksjackson: To stay consistent with other exercises (since this will be public) copy the enum formatting used for the Allergies exercise.

So use a typedef instead, rename `flavor` to `Flavor_t`, and capitalize the values (`invalid` -> `Invalid`)

patricksjackson:
Ok, I see you just grabbed the solution from the xcpp track (which is good, much of that code can be shared). Unfortunately it seems that solution is out of date with the canonical test cases.

The idea behind the change is that since triangles can be multiple types (e.g. every Equilateral triangle is also Isosceles) it makes more sense to test Triangle properties rather than determining their type.

What you need to do then is to rewrite it to use 3 different functions that all return a `bool`: `is_equilateral`, `is_isosceles`, and `is_scalene`. Then test them with the tests in the canonical json.

I suggest defining a `Triangle_t` type in `triangle.h` and using that as input to each function.

wolf99:
@patricksjackson strictly speaking, the `_t` suffix shouldn't be used as it is reserved by POSIX (I know this isn't POSIX but as far as I'm aware good practice generally dictates not using them in case POSIX does get used in a potential future iteration). Maybe I'm being too picky but just my 2c 😄

DeathSec:
got it , i thought xcpp repo was well defined so i used it to maintain the consistent code style...

DeathSec:
@patricksjackson  indent is not working correctly! help!

wolf99:
@DeathSec you used `indent.sh`? or `$ indent -linux -i3 -nut`?

exercises/triangle/test/test_triangle.c
```diff
@@ -0,0 +1,127 @@
+#include "vendor/unity.h"
+#include "../src/triangle.h"
+
+void test_equilateral_is_true_if_all_sides_are_equal(void)
+{
+   Triangle_t sides = { 2, 2, 2 };
+   TEST_ASSERT(is_equilateral(&sides));
```
wolf99: I would recommend `TEST_ASSERT_TRUE()` used for these test cases

exercises/triangle/test/test_triangle.c
```diff
@@ -0,0 +1,127 @@
...
+}
+
+void test_equilateral_is_false_if_any_side_is_unequal(void)
+{
+   Triangle_t sides = { 2, 3, 2 };
+   TEST_ASSERT(!(is_equilateral(&sides)));
```
wolf99: And `TEST_ASSERT_FALSE()` for these ones

DeathSec:
gnu indent is not working , any solutions?...

exercises/triangle/test/test_triangle.c
```diff
@@ -4,101 +4,97 @@
...
-test_sides_that_violate_triangle_inequality_are_not_isosceles_even_if_two_are_equal
-(void)
+void test_sides_that_violate_triangle_inequality_are_not_isosceles_even_if_two_are_equal(void)
 {
    Triangle_t sides = { 1, 1, 3 };
-   TEST_ASSERT(!(is_isosceles(&sides)));
+    TEST_ASSERT_FALSE( is_isosceles(&sides ));
 }
 
 void test_in_isosceles_sides_may_be_floats(void)
```
wolf99: Remove `_in` here. Leave as `test_isosceles_sides_may_be_floats`

exercises/triangle/test/test_triangle.c
```diff
@@ -4,101 +4,97 @@
...
 }
 
 void test_equilateral_is_false_if_any_side_is_unequal(void)
 {
    Triangle_t sides = { 2, 3, 2 };
-   TEST_ASSERT(!(is_equilateral(&sides)));
+   TEST_ASSERT_FALSE(is_equilateral(&sides));
 }
 
 void test_all_zero_sides_are_illegal_so_the_triangle_is_not_equilateral(void)
```
wolf99: Rename to follow same convention used for the other tests, change `test_all_zero_sides_are_illegal_so_the_triangle_is_not_equilateral` to `test_equilateral_is_false_if_all_sides_zero`

exercises/triangle/test/test_triangle.c
```diff
@@ -4,101 +4,97 @@
...
 {
    Triangle_t sides = { 2, 3, 4 };
-   TEST_ASSERT(!(is_isosceles(&sides)));
+    TEST_ASSERT_FALSE( is_isosceles(&sides ));
 }
 
-void
-test_sides_that_violate_triangle_inequality_are_not_isosceles_even_if_two_are_equal
-(void)
+void test_sides_that_violate_triangle_inequality_are_not_isosceles_even_if_two_are_equal(void)
```
wolf99: This function name needs to be shorter to pass indent. Try changing to `test_isosceles_is_false_if_two_sides_equal_and_violate_inequality`

exercises/triangle/test/test_triangle.c
```diff
@@ -4,101 +4,97 @@
...
 }
 
 void test_all_zero_sides_are_illegal_so_the_triangle_is_not_equilateral(void)
 {
    Triangle_t sides = { 0, 0, 0 };
-   TEST_ASSERT(!(is_equilateral(&sides)));
+    TEST_ASSERT_FALSE(is_equilateral(&sides ));
 }
 
 void test_in_equilateral_sides_may_be_floats(void)
```
wolf99: Remove `_in` here. Leave as `test_equilateral_sides_may_be_floats`, as it makes more sense

exercises/triangle/test/test_triangle.c
```diff
@@ -4,101 +4,97 @@
...
-test_sides_that_violate_triangle_inequality_are_not_scalene_even_if_they_are_all_different
-(void)
+void test_sides_that_violate_triangle_inequality_are_not_scalene_even_if_they_are_all_different(void)
 {
    Triangle_t sides = { 7, 3, 2 };
-   TEST_ASSERT(!(is_scalene(&sides)));
+    TEST_ASSERT_FALSE( is_scalene(&sides ));
 }
 
 void test_in_scalene_sides_may_be_floats(void)
```
wolf99: Remove _in here. Leave as test_scalene_sides_may_be_floats

exercises/triangle/test/test_triangle.c
```diff
@@ -4,101 +4,97 @@
...
 {
    Triangle_t sides = { 4, 4, 3 };
-   TEST_ASSERT(!(is_scalene(&sides)));
+    TEST_ASSERT_FALSE( is_scalene(&sides ));
 }
 
-void
-test_sides_that_violate_triangle_inequality_are_not_scalene_even_if_they_are_all_different
-(void)
+void test_sides_that_violate_triangle_inequality_are_not_scalene_even_if_they_are_all_different(void)
```
wolf99: This function name needs to be shorter to pass indent. Try changing to `test_scalene_is_false_if_no_sides_equal_and_violate_inequality`

exercises/triangle/test/test_triangle.c
```diff
@@ -4,101 +4,97 @@
...
 }
 
 void test_in_scalene_sides_may_be_floats(void)
 {
    Triangle_t sides = { 0.5, 0.4, 0.6 };
-   TEST_ASSERT(is_scalene(&sides));
+    TEST_ASSERT_TRUE(is_scalene(&sides));
 }
 
 int main(void)
```
wolf99: Don't forget to update the function names called by the RUN_TEST macros 😃

DeathSec:
@wolf99 thank you..

DeathSec:
@wolf99 i don't know why but , verify-indent does not work in travis-ci , but it worked in my computer ??

wolf99:
LGTM 😃

config.json
```diff
@@ -218,6 +218,15 @@
      ]
   }, {
      "difficulty": 1,
+     "slug": "triangle",
+     "topics": [
+		"control-flow (if-else statements)",
+		"Booleans",
+		"Structs",
+		"Mathematics"
+	       ]
```
ryanplusplus: The indentation is off here with respect to the rest of the file. Can you correct this before merge?

exercises/triangle/src/example.h
```diff
@@ -0,0 +1,15 @@
+#ifndef TRIANGLE_H
+#define TRIANGLE_H
+#include <stdbool.h>
+
+typedef struct {
+   double point_a;
+   double point_b;
+   double point_c;
```
ryanplusplus: These are lengths... not points, right?

ryanplusplus:
I think this is good to go once a few small issues are resolved.

ryanplusplus:
Looks great, thanks for following up!

PR #122, opened 2017-02-17
Space age Exercice added

Guntau:
I think nothing is missing.
Sorry about the multiple commits :(


ryanplusplus:
You're failing CI but it looks like an issue either with Travis using a PPA that is either no longer supported or that is currently down. We can re-trigger the CI build later to see if it is resolved.

exercises/space-age/src/example.c
```diff
@@ -0,0 +1,41 @@
+#include "example.h"
+
+float convert_Earth_Age(const long input)
```
ryanplusplus: How about `convert_earth_age`? This will be consistent with the rest of the exercises.

exercises/space-age/src/example.h
```diff
@@ -0,0 +1,25 @@
...
+#define SECONDS_ON_EARTH_YEAR 31557600
+
+/* Orbital period of each planet, in Earth years */
+#define MERCURY_YEAR 0.2408467
+#define VENUS_YEAR 0.61519726
+#define MARS_YEAR 1.8808158
+#define JUPITER_YEAR 11.862615
+#define SATURN_YEAR 29.447498
+#define URANUS_YEAR 84.016846
+#define NEPTUNE_YEAR 164.79132
```
ryanplusplus: Any reason why these are in the header file? It seems like these should be private to the implementation and thus in the .c file.

exercises/space-age/src/example.h
```diff
@@ -0,0 +1,25 @@
+#ifndef EXAMPLE_H
+#define EXAMPLE_H
```
ryanplusplus: Examples for our other exercises use the actual module name for the include guard. Can you change this to `SPACE_AGE_H`?

exercises/space-age/test/test_space-age.c
```diff
@@ -0,0 +1,58 @@
...
+
+
+void test_convert_Earth_Year(void)
+{
+   TEST_ASSERT_EQUAL(1, convert_Earth_Age(SECONDS_ON_EARTH_YEAR));
+}
+
+void test_convert_Mercury_Year(void)
+{
+   TEST_ASSERT_EQUAL(0.2408467, convert_Mercury_Age(SECONDS_ON_EARTH_YEAR));
```
ryanplusplus: Unity includes an assertions meant for testing float equality (`TEST_ASSERT_EQUAL_FLOAT`). This should be used to the many complexities of comparing floating point numbers. You can read more about why this is important here: https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/.

ryanplusplus:
The CI build is now running correctly but the example fails the tests.

Guntau:
thank you for your time, this is the kind of feedback that makes me want to persevere 👍 .
I think everything is ok now

exercises/space-age/makefile
```diff
@@ -0,0 +1,16 @@
...
+
+test: tests.out
+	@./tests.out
+
+clean:
+	rm -f *.o *.out
+
+tests.out: test/test_space-age.c src/space-age.c src/space-age.h
+	@echo Compiling $@
+	@cc $(CFLAGS) src/space-age.c test/vendor/unity.c test/test_space-age.c -o tests.out
```
ryanplusplus: Looks like this expects the tests to be called `test_space-age.c` but the file was renamed to `test_space_age.c`. This is causing the CI failure.

ryanplusplus:
No problem, thanks for the quick follow up :smile:

exercises/space-age/makefile
Guntau: Woops, my bad..
It should be ok now !

exercises/space-age/test/test_space_age.c
```diff
@@ -0,0 +1,58 @@
...
+}
+
+void test_convert_uranus_year(void)
+{
+   TEST_ASSERT_EQUAL_FLOAT(84.016846, convert_uranus_age(SECONDS_ON_EARTH_YEAR));
+}
+
+void test_convert_neptune_year(void)
+{
+   TEST_ASSERT_EQUAL_FLOAT(164.79132, convert_neptune_age(SECONDS_ON_EARTH_YEAR));
```
ryanplusplus: Now your tests are failing because you're using `SECONDS_ON_EARTH_YEAR` in here but it is private to the implementation. Instead of using this, make sure you're using the input values from the standard test data: https://github.com/exercism/x-common/blob/master/exercises/space-age/canonical-data.json.

Guntau:
I tried another method, but I don't understand why the tests are failing in the first case..

ryanplusplus:
Have you run it locally to see what values you're getting? It could be that there's simply an issue with your sample implementation, or it could be that your answer is close but not within the delta that the float comparison uses. If you're just slightly off, there are other assertions that allow you to specify the delta that you could use to open up the range of accepted values.

wolf99:
@Guntau If the issue is as ryanplusplus descrivbes then the assertion you need to read up on is `TEST_ASSERT_FLOAT_WITHIN(delta, exp, actual)` from the unity assertions reference guides, here: [https://github.com/ThrowTheSwitch/Unity/tree/master/docs](https://github.com/ThrowTheSwitch/Unity/tree/master/docs)

Guntau:
I took a bigger delta for some tests, it seems to work. I also verified the values by hand, and they correspond with the canonical data

ryanplusplus:
This looks good to me

DeathSec:
`
#define EARTH 0
#define MERCURY 0.2408467
...

static float earth_year = 0.00000003168808781;

float space_age(float seconds, float planet)
{
   if (planet == 0) {
      return seconds * earth_year;
   }
   return (seconds * earth_year) / planet;
}
`
just a comment , for making the lines of code small and elegant looking , but your code Looks good to me

Guntau:
Yeah i thought about doing it that way, but the description of the exercise says that the only value given is the time in seconds. However I agree that your version is more elegant 👍

wolf99:
For simpler calculation at run-time you could instead do:
~~~c
#define FOOPLANET_YEAR_IN_SECONDS (FOOPLANET_YEAR * SECONDS_ON_EARTH_YEAR)

float convert_fooplanet_age(const long input)
{
    return input / FOOPLANET_YEAR_IN_SECONDS; // No need to call convert_earth_age()
}
~~~

But this is only a minor detail.

patricksjackson:
This looks good :+1: There is one future improvement I want to float out there though...

This problem has always annoyed me that you have to write a bunch of identical functions (maybe that's the point?). But I've always thought this could very easily be improved by just using an enum in the function argument for the planet conversion. So this would reduce down to a single function with the sig:

`float convert_planet_age(Planet_t planet, uint64_t input)`

That would make it much less annoying to write, and barely changes how it's used.

(I should've weighed in earlier, but real life got in the way :\ )

Guntau:
DeathSec proposed something like this too. I'll change it then, i think i just misunderstand the problem's description. thanks for the feedback

PR #123, opened 2017-02-18
Pascals triangle exercise added

DeathSec:
the maximum  that this program or rather c can handle is 20 rows , if row > 20 then it results in buffer-overflow

### DIFF
--- a/exercises/pascals-triangle/src/example.c
+++ b/exercises/pascals-triangle/src/example.c
@@ -1,22 +1,26 @@
 #include "pascals_triangle.h"
 
-long long unsigned int factorial(int n)
+size_t factorial(int n)
 {
-   return (n == 0) ? 1 : n * factorial(n - 1);
+   if (n > 20 || n < 0) {
+      return 0;
+   } else {
+      return (n == 0) ? 1 : n * factorial(n - 1);
+   }
 }
 
-long long unsigned int combination(int n, int r)
+size_t combination(int n, int r)
 {
-   return (r > n) ? 0 : factorial(n) / (factorial(n - r) * factorial(r));
+   return (r > n || n > 20
+           || n < 0) ? 0 : factorial(n) / (factorial(n - r) * factorial(r));
 }
 
-int pascals_triangle(int count, int input[][((count < 0) ? 1 : count)])
+int pascals_triangle(int count, size_t input[][((count < 0) ? 1 : count)])
 {
    if (count < 0) {
-      memset(input, -1, sizeof(int) * count * count);
+      input[0][0] = -1;
       return -1;
    }
-   memset(input, 0, sizeof(int) * count * count);
    for (int n = 0; n < count; n++) {
       for (int rn = 0; rn < n; rn++) {
          input[n][rn] = combination(n, rn);
--- a/exercises/pascals-triangle/src/example.c
+++ b/exercises/pascals-triangle/src/example.c
@@ -1,30 +1,50 @@
 #include "pascals_triangle.h"
 
-size_t factorial(int n)
+static void zero_out(size_t ** input, size_t count)
 {
-   if (n > 20 || n < 0) {
-      return 0;
-   } else {
-      return (n == 0) ? 1 : n * factorial(n - 1);
-   }
-}
-
-size_t combination(int n, int r)
-{
-   return (r > n || n > 20
-           || n < 0) ? 0 : factorial(n) / (factorial(n - r) * factorial(r));
-}
-
-int pascals_triangle(int count, size_t input[][((count < 0) ? 1 : count)])
-{
-   if (count < 0) {
-      input[0][0] = -1;
-      return -1;
-   }
-   for (int n = 0; n < count; n++) {
-      for (int rn = 0; rn < n; rn++) {
-         input[n][rn] = combination(n, rn);
+   size_t i, j;
+   for (i = 0; i < count; i++) {
+      for (j = 0; j < count; j++) {
+         input[i][j] = 0;
       }
    }
-   return 0;
+}
+
+static size_t factorial(size_t n)
+{
+   return (n == 0) ? 1 : n * factorial(n - 1);
+}
+
+static size_t combination(size_t n, size_t r)
+{
+   return (r > n) ? 0 : factorial(n) / (factorial(n - r) * factorial(r));
+}
+
+size_t **pascals_triangle(char *count)
+{
+   size_t **zero = (size_t **) malloc(sizeof(size_t *));
+   zero[0] = (size_t *) malloc(sizeof(size_t));
+   zero[0][0] = 0;
+   int i, j;
+   int rows = atoi(count);
+   if (!strcmp(count, "0") || !strcmp(count, "21")) {
+      return zero;
+   }
+   if ((rows && rows >= 0 && strcmp(count, ""))
+       || (rows && rows >= 0 && count != NULL)) {
+      size_t **result = (size_t **) malloc(sizeof(size_t *) * rows);
+      for (i = 0; i < rows; i++) {
+         result[i] = (size_t *) malloc(sizeof(size_t) * rows);
+      }
+      zero_out(result, rows);
+      for (i = 0; i < rows; i++) {
+         for (j = 0; j <= i; j++) {
+            result[i][j] = combination(i, j);
+         }
+      }
+      return result;
+   } else {
+      return (size_t **) - 1;
+   }
+   return zero;
 }
--- a/exercises/pascals-triangle/src/example.c
+++ b/exercises/pascals-triangle/src/example.c
@@ -27,7 +27,7 @@ size_t **pascals_triangle(char *count)
    zero[0][0] = 0;
    int i, j;
    int rows = atoi(count);
-   if (!strcmp(count, "0") || !strcmp(count, "21")) {
+   if (!strcmp(count, "0") || rows > 20) {
       return zero;
    }
    if ((rows && rows >= 0 && strcmp(count, ""))
--- a/exercises/pascals-triangle/src/example.c
+++ b/exercises/pascals-triangle/src/example.c
@@ -1,0 +1,26 @@
+#include "pascals_triangle.h"
+
+long long unsigned int factorial(int n)
+{
+   return (n == 0) ? 1 : n * factorial(n - 1);
+}
+
+long long unsigned int combination(int n, int r)
+{
+   return (r > n) ? 0 : factorial(n) / (factorial(n - r) * factorial(r));
+}
+
+int pascals_triangle(int count, int input[][((count < 0) ? 1 : count)])
+{
+   if (count < 0) {
+      memset(input, -1, sizeof(int) * count * count);
+      return -1;
+   }
+   memset(input, 0, sizeof(int) * count * count);
+   for (int n = 0; n < count; n++) {
+      for (int rn = 0; rn < n; rn++) {
+         input[n][rn] = combination(n, rn);
+      }
+   }
+   return 0;
+}
--- a/exercises/pascals-triangle/src/example.h
+++ b/exercises/pascals-triangle/src/example.h
@@ -1,7 +1,8 @@
 #ifndef PASCALS_TRIANGLE_H
 #define PASCALS_TRIANGLE_H
-#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
 
-int pascals_triangle(int count, size_t input[][count]);
+size_t **pascals_triangle(char *count);
 
 #endif
--- a/exercises/pascals-triangle/src/example.h
+++ b/exercises/pascals-triangle/src/example.h
@@ -1,7 +1,7 @@
 #ifndef PASCALS_TRIANGLE_H
 #define PASCALS_TRIANGLE_H
-#include <string.h>
+#include <stddef.h>
 
-int pascals_triangle(int count, int input[][count]);
+int pascals_triangle(int count, size_t input[][count]);
 
 #endif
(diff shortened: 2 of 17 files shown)


exercises/pascals-triangle/test/test_pascals_triangle.c
```diff
@@ -0,0 +1,88 @@
...
+   }
+   return 1;
+}
+
+void test_no_rows(void)
+{
+   int result[1][1];
+   int expected[1][1];
+   pascals_triangle(0, result);
+   TEST_ASSERT_TRUE(check(1, result, expected));
```
DeathSec: I used my own little function because Unity does not have a 2d array test , but it has a 1d array test , i could use that with 'for loops' but i thought it would be light if i used this little check function

exercises/pascals-triangle/test/test_pascals_triangle.c
```diff
@@ -0,0 +1,88 @@
+#include <stdbool.h>
+#include "../src/pascals_triangle.h"
+#include "vendor/unity.h"
+
+bool check(int count, int result[][count], int expected[][count])
+{
+   for (int i = 0; i < count; i++) {
+      for (int j = 0; j < i; j++) {
```
DeathSec: pascals triangle have n number of columns in n number of rows

ryanplusplus:
Everything looks good but the buffer overflow is a bit concerning. Where does the overflow occur?

DeathSec:
the buffer flow is in the factorial function , you know that c cannot store the multiplication of two numbers which exceed 64 bits if unsigned , sizeof(long long unsigned int) = 8 Bytes = 8*8 = 64 bits , maximum limit = 2^64 = 1.844e19 , which factorial exceeds easily.

ryanplusplus:
Oh, okay. That's integer overflow, not a buffer overflow. A buffer overflow is when a buffer is written to out of bounds and causes memory corruption (and usually a crash).

ryanplusplus:
Looks good to me.

@patricksjackson @palisaide can one of you guys give this a once-over before we merge this?

DeathSec:
Little confusion with overflows!.

wolf99:
You have used `sizeof` on a type rather than a variable I would suggest that `sizeof(input[0][0])` would be better. 

Also that your `combination()` function returns an `unsigned long long` but the result is assigned into a location in an array of type `int` - I imagine this would result in truncation a lot for larger numbers...

Might be overkill but re the overflow, you could add a check to to the factorial function to check for the overflow and return `ULLONG_MAX` if it will overflow, eg
~~~c
x = a * b;
if (a != 0 && x / a != b) {
    // overflow handling
}
~~~

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,26 @@
...
+
+long long unsigned int combination(int n, int r)
+{
+   return (r > n) ? 0 : factorial(n) / (factorial(n - r) * factorial(r));
+}
+
+int pascals_triangle(int count, int input[][((count < 0) ? 1 : count)])
+{
+   if (count < 0) {
+      memset(input, -1, sizeof(int) * count * count);
```
wolf99: if count is less than 0 then `sizeof(int) * count * count` will not be very good, though it will probably work in most cases. For example, if count is `-20` then this `memset()` will try and clear 400 int's worth of space... it only needs 1 ints worth.

 `sizeof(int)` would be better. Or `sizeof(input[0][0])` would be even better.

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,26 @@
...
+      return -1;
+   }
+   memset(input, 0, sizeof(int) * count * count);
```
wolf99: When count is zero, this should result in a warning. It also means that when count is zero the value in input is left at whatever it was previously - it is never actually explicitly set to 0. the test only passes because your environment happens to set the `auto` scoped `result` array to all zeros by default.

exercises/pascals-triangle/src/example.c
DeathSec: oh!

DeathSec: i never looked deep into this , good call!

exercises/pascals-triangle/src/example.c
DeathSec: we cannot give a warning , because it must handle anything quietly and canonical-data wants to return a 0 row array anyway , (where C89 we cannot assign int array[0][0]) , so i used a single row

DeathSec:
i will do the changes soon

exercises/pascals-triangle/src/example.c
wolf99: what I mean is that on a good compiler, the code as is, will give a warning - thus it should change. Here you have not used a single row as you say you intend, you have used however much is `count` and hoped that it is never less than `-1`, if a user or test were to enter `count = -20`, this would attempt to set 400 rows! (`count * count == -20 * -20 == 400`)

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,30 @@
+#include "pascals_triangle.h"
+
+size_t factorial(int n)
+{
+   if (n > 20 || n < 0) {
```
DeathSec: 2^64 > 20! , so the maximum the function can handle is 20!

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,30 @@
+#include "pascals_triangle.h"
+
+size_t factorial(int n)
```
patricksjackson: Internal helper functions should be declared static.

exercises/pascals-triangle/src/example.h
```diff
@@ -0,0 +1,7 @@
+#ifndef PASCALS_TRIANGLE_H
+#define PASCALS_TRIANGLE_H
+#include <stddef.h>
+
+int pascals_triangle(int count, size_t input[][count]);
```
patricksjackson: You never use the return value for this function. Either make the return value a void or change the test suite. (You already use the array as a return/error condition).

As a nitpick, you should rename `input` to be something like `result`, since it's not actually and input.

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,30 @@
...
+}
+
+int pascals_triangle(int count, size_t input[][((count < 0) ? 1 : count)])
+{
+   if (count < 0) {
+      input[0][0] = -1;
+      return -1;
+   }
+   for (int n = 0; n < count; n++) {
+      for (int rn = 0; rn < n; rn++) {
```
patricksjackson: This inner loop is wrong, and the test suite only passes because the same mistake is in `check()`.

`for (int rn = 0; rn <= n; rn++)`

exercises/pascals-triangle/test/test_pascals_triangle.c
```diff
@@ -0,0 +1,88 @@
...
+         }
+      }
+   }
+   return 1;
+}
+
+void test_no_rows(void)
+{
+   size_t result[1][1];
+   size_t expected[1][1];
```
patricksjackson: Arrays initialized here are not zeroed out, so `expected` has garbage values in it unless you explicitly save a 0 to it.

patricksjackson:
A few other things:

`count` (and all the loop variables) should really just be a `size_t` everywhere, there's not much point in allowing a negative argument.

If count == 0, what is the expected returned array?

Does it make sense to zero out the rest of the returned array? Right now it contains garbage values.

Isn't the max triangle size a lot larger than 20 rows? Yes, the factorial overflows but the max value is only `352716` on row 21.

DeathSec:
i will look into it soon..

exercises/pascals-triangle/test/test_pascals_triangle.c
```diff
@@ -0,0 +1,82 @@
...
+            return 0;
+         }
+      }
+   }
+   return 1;
+}
+
+void test_no_rows(void)
+{
+   size_t expected[1][1] = { {0} };
```
DeathSec: we cannot have or check a null 2d array in C

exercises/pascals-triangle/test/test_pascals_triangle.c
```diff
@@ -0,0 +1,82 @@
...
+   size_t expected[1][1] = {
+      {1}
+   };
+   TEST_ASSERT_TRUE(check(1, expected, pascals_triangle("1")));
+}
+
+void test_two_rows(void)
+{
+   size_t expected[2][2] = {
+      {1, 0},
```
DeathSec: the 0 is just a place holder to avoid garbage!

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+{
+   return (n == 0) ? 1 : n * factorial(n - 1);
+}
+
+static size_t combination(size_t n, size_t r)
+{
+   return (r > n) ? 0 : factorial(n) / (factorial(n - r) * factorial(r));
+}
+
+size_t **pascals_triangle(char *count)
```
DeathSec: declared count as (char *) because , we can satisfy the test data to have a null input and also handle negative input quietly

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+   }
+   if ((rows && rows >= 0 && strcmp(count, ""))
+       || (rows && rows >= 0 && count != NULL)) {
+      size_t **result = (size_t **) malloc(sizeof(size_t *) * rows);
+      for (i = 0; i < rows; i++) {
+         result[i] = (size_t *) malloc(sizeof(size_t) * rows);
+      }
+      zero_out(result, rows);
+      for (i = 0; i < rows; i++) {
+         for (j = 0; j <= i; j++) {
```
DeathSec: pascals triangle have n cols in n rows! so leave the rest be zero

exercises/pascals-triangle/test/test_pascals_triangle.c
DeathSec: so declare everything like Square Matrix

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+   int i, j;
+   int rows = atoi(count);
+   if (!strcmp(count, "0") || rows > 20) {
+      return zero;
+   }
+   if ((rows && rows >= 0 && strcmp(count, ""))
+       || (rows && rows >= 0 && count != NULL)) {
+      size_t **result = (size_t **) malloc(sizeof(size_t *) * rows);
+      for (i = 0; i < rows; i++) {
+         result[i] = (size_t *) malloc(sizeof(size_t) * rows);
```
patricksjackson: If you use `calloc` instead of malloc you can avoid the `zero_out` function entirely.

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+      }
+      zero_out(result, rows);
+      for (i = 0; i < rows; i++) {
+         for (j = 0; j <= i; j++) {
+            result[i][j] = combination(i, j);
+         }
+      }
+      return result;
+   } else {
+      return (size_t **) - 1;
```
patricksjackson: It would be best practice to return NULL on error instead of that.

exercises/pascals-triangle/test/test_pascals_triangle.c
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void test_null_row_is_not_allowed(void)
+{
+   TEST_ASSERT_TRUE((pascals_triangle("") == (size_t **) - 1));
+}
+
+int main(void)
+{
+   UnityBegin("test_pascals_triangle.c");
```
patricksjackson: All the other tests seem to write this as `UnityBegin("test/test_pascals_triangle.c");`

Admittedly, I don't know which way is correct though.

patricksjackson:
Sorry for dropping all those notes on you and then leaving for a while, (just started a new job, I've been a bit busy).

This, overall, looks very good :) I think this is fine to merge if you change the Error return value to NULL and make sure the `UnityBegin` call is correct, everything else I'm pointing out is mostly nitpicky.

So onto the nitpicks:
You never actually test it with a NULL input, and I think you don't handle that case. Passing `""` gives it a pointer to an empty string, not a NULL pointer.
You could go make sure that you clean up the memory. This leaks quite a bit. 

Also the last time I think I was trying to point out that you could avoid the limit on the size of the triangle by doing the calculations a different way. When you need to calculate the factorials you top out the integer calculations very quickly. Instead you could do something like this:

```
result[0][0] = 1;
for (size_t i = 1; i < rows; ++i) {
 result[i][0] = 1;
 for (size_t j = 1; j <= i; ++j) {
  result[i][j] = result[i-1][j] + result[i-1][j-1];
 }
}
```

DeathSec:
i will do the changes soon

DeathSec:
@patricksjackson your calculation on pascals triangle sure is very elegant , I never thought of this! , +1

patricksjackson:
@DeathSec - Can you rebase these changes on top of master? This solution won't pass the Valgrind tests that just got merged, so it'll need a small amount of tweaking.

[Instructions for rebasing](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#rebasing). Feel free to ask if you're not that familiar with it.

I believe you'll have to add an extra function to free your allocated 2D array at the end of every test to get it to pass.

```
void free_triangle(size_t **triangle, size_t rows)
```

DeathSec:
all ready done everything , i am closing this pull and creating a new one , the exercise is clean , no leaks are possible!

PR #125, opened 2017-03-03
Improved code for space age

Guntau:
I improved the code as suggested in the last [ pull request](https://github.com/exercism/xc/pull/122).
So there is only one function now, where you give the planet name (the value is defined in the header) and the number of seconds to get the result.
Is it a problem if the `#define` are in the header ?

### DIFF
--- a/exercises/space-age/src/example.c
+++ b/exercises/space-age/src/example.c
@@ -1,53 +1,6 @@
 #include "space_age.h"
 
-/* Number of seconds in a Earth year */
-#define SECONDS_ON_EARTH_YEAR 31557600
-
-/* Orbital period of each planet, in Earth years */
-#define MERCURY_YEAR (0.2408467 * SECONDS_ON_EARTH_YEAR)
-#define VENUS_YEAR (0.61519726 * SECONDS_ON_EARTH_YEAR)
-#define MARS_YEAR (1.8808158 * SECONDS_ON_EARTH_YEAR)
-#define JUPITER_YEAR (11.862615 * SECONDS_ON_EARTH_YEAR)
-#define SATURN_YEAR (29.447498 * SECONDS_ON_EARTH_YEAR)
-#define URANUS_YEAR (84.016846 * SECONDS_ON_EARTH_YEAR)
-#define NEPTUNE_YEAR (164.79132 * SECONDS_ON_EARTH_YEAR)
-
-float convert_earth_age(const long input)
+float convert_planet_age(float planet, const long input)
 {
-   return input / SECONDS_ON_EARTH_YEAR;
-}
-
-float convert_mercury_age(const long input)
-{
-   return input / MERCURY_YEAR;
-}
-
-float convert_venus_age(const long input)
-{
-   return input / VENUS_YEAR;
-}
-
-float convert_mars_age(const long input)
-{
-   return input / MARS_YEAR;
-}
-
-float convert_jupiter_age(const long input)
-{
-   return input / JUPITER_YEAR;
-}
-
-float convert_saturn_age(const long input)
-{
-   return input / SATURN_YEAR;
-}
-
-float convert_uranus_age(const long input)
-{
-   return input / URANUS_YEAR;
-}
-
-float convert_neptune_age(const long input)
-{
-   return input / NEPTUNE_YEAR;
+   return input / planet;
 }
--- a/exercises/space-age/src/example.c
+++ b/exercises/space-age/src/example.c
@@ -1,6 +1,43 @@
 #include "space_age.h"
 
-float convert_planet_age(float planet, const long input)
+/* Number of seconds in a Earth year */
+#define EARTH 31557600
+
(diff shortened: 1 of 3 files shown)


ryanplusplus:
This makes it a lot cleaner, but I have to say that it seems like it eliminates the reason for the exercise. Particularly because the interface nearly forces this to be implemented with a one liner, I'm not sure that this is an improvement from a learning standpoint.

wolf99:
@ryanplusplus In the current implementation they're all just one liners too. Multiple one-liners sure, but they all do the same thing so nothing new is learned with each function.

wolf99:
Idea! perhaps if the function accepts a simple `enum planets`, which the implementation should use to lookup the constant value in some way? There's an opportunity, at a minimum, to introduce the use of enums, but also lookups and so on, and potentially X macros. Learners could make the implementation as simple or as complex as they desire with such an interface.

wolf99:
For eg, something like:

~~~c
typedef enum planets
{
    mercury = 0,
    venus,
    earth,
    mars,
    jupiter,
    saturn,
    uranus,
    neptune,
    max_planet = neptune
};

float convert_planet_age(planets planet, const long input);
~~~

ryanplusplus:
@wolf99 With a bunch of one-liners, you don't learn something new with each function, but you do have to contend with how to manage all of that duplication. I think finding a clean solution to the duplication is the primary challenge of this exercise.

I think your idea to use an enum is a good one as it introduces new concepts in addition to simply eliminating the old.

wolf99:
@ryanplusplus Not sure which approach you're advocating for now :smile:

Have seen that one submission does use an approach to reduce the duplication. but the deduplication lesson is not forced - submissions can pass all tests without reducing the duplication.

With the enum approach at least the enum and lookups of some form are *required* to pass tests...

ryanplusplus:
@wolf99 I'm fine with either the original or with your suggestion.

You're right that the de-duplication is not forced since we can't do that with tests. Luckily Exercism encourages feedback from other users :)

exercises/space-age/src/example.c
```diff
@@ -1,6 +1,43 @@
...
+#define MARS (1.8808158 * EARTH)
+#define JUPITER (11.862615 * EARTH)
+#define SATURN (29.447498 * EARTH)
+#define URANUS (84.016846 * EARTH)
+#define NEPTUNE (164.79132 * EARTH)
+
+float convert_planet_age(Planets planet, const long input)
 {
-   return input / planet;
+   switch (planet) {
```
ryanplusplus: How about a table-based solution instead of using a large switch case?

Guntau: I don't know how to do it :(
Is it on the enum ?

exercises/space-age/src/example.h
```diff
@@ -1,18 +1,18 @@
 #ifndef SPACE_AGE_H
 #define SPACE_AGE_H
 
-/* Number of seconds in a Earth year */
-#define EARTH 31557600
+typedef enum planets {
+   mercury = 0,
```
ryanplusplus: The `= 0` is not necessary (this is the automatic behavior)

exercises/space-age/src/example.h
```diff
@@ -1,18 +1,18 @@
...
+   venus,
+   earth,
+   mars,
+   jupiter,
+   saturn,
+   uranus,
+   neptune,
+   max_planet = neptune
```
ryanplusplus: This is not necessary (and makes it more likely to break if they ever make Pluto a planet again) because this is the automatic behavior.

exercises/space-age/src/example.c
ryanplusplus: What you do is use the enum value as a table index. Since the enumerated values start at 0 and increment by 1, they are natural array keys. So, something like this:

```c
static const float factors = {
    543 // mercury,
    123, // venus
    // ...
    56789 // neptune
};

// ...

float convert_planet_age(Planets planet, const long input) {
   return input / factors[planet];
}
```

exercises/space-age/src/example.h
wolf99: It was my understanding that this is good usage as it prevents having to reference `neptune` directly in any limit case (e.g. a `for` loop). If for example, pluto is added, then in this manner only the `enum` needs changing. Sure it needs changing in two places (right next to each other), but this beats having to change it in the enum _and_ any limit cases throughout the code. Is this not correct?

ryanplusplus: Sorry, I was pretty unclear here. The `= neptune` part is unnecessary because `max_planet` will default to `neptune + 1` which you can use for a `for` limit without making the enum more difficult to maintain by needing to babysit the max definition as items are added.

wolf99: Ah, gotcha! 👍

Guntau:
I'll do the changes within a few days, i've a lot of work to do at the moment 👍

ryanplusplus:
👍

PR #126, opened 2017-03-06
react: Add to track

petertseng:
~~This is a work in progress:~~

1. ~~Not all tests have been added yet.~~
2. ~~The example does not yet have remove_callback. Should be easy enough to add, given everything else is reasonable~~.

At this point, it would be a good time to review the API being presented to the student in react.h to see if it is reasonable. ~~There are a few questions marked TODO that deserve attention as well~~.

If at any point you feel the answer should be "This exercise should not be added to this track", don't hesitate to say that as well.

### DIFF
--- a/exercises/react/src/example.c
+++ b/exercises/react/src/example.c
@@ -60,15 +60,9 @@ void destroy_reactor(struct reactor *r)
    free(r);
 }
 
-#define SUCCESS 1
-#define FAIL 0
-
-static int add_child(struct reactor *r, struct cell *cell)
+static void add_child(struct reactor *r, struct cell *cell)
 {
    struct child *child = calloc(1, sizeof(struct child));
-   if (!child) {
-      return FAIL;
-   }
    child->cell = cell;
    if (!r->first) {
       r->first = child;
@@ -76,7 +70,6 @@ static int add_child(struct reactor *r, struct cell *cell)
       r->last->next = child;
    }
    r->last = child;
-   return SUCCESS;
 }
 
 struct cell *create_input_cell(struct reactor *r, int initial_value)
@@ -85,10 +78,7 @@ struct cell *create_input_cell(struct reactor *r, int initial_value)
    if (!c) {
       return NULL;
    }
-   if (add_child(r, c) != SUCCESS) {
-      free(c);
-      return NULL;
-   }
+   add_child(r, c);
    c->reactor = r;
    c->kind = kind_input;
    c->value = initial_value;
@@ -102,10 +92,7 @@ struct cell *create_compute1_cell(struct reactor *r, struct cell *input,
    if (!c) {
       return NULL;
    }
-   if (add_child(r, c) != SUCCESS) {
-      free(c);
-      return NULL;
-   }
+   add_child(r, c);
    c->reactor = r;
    c->kind = kind_compute1;
    c->input1 = input;
@@ -121,10 +108,7 @@ struct cell *create_compute2_cell(struct reactor *r, struct cell *input1,
    if (!c) {
       return NULL;
    }
-   if (add_child(r, c) != SUCCESS) {
-      free(c);
-      return NULL;
-   }
+   add_child(r, c);
    c->reactor = r;
    c->kind = kind_compute2;
    c->input1 = input1;
--- a/exercises/react/src/example.c
+++ b/exercises/react/src/example.c
@@ -42,7 +42,7 @@ struct reactor *create_reactor()
    return r;
 }
 
-void free_reactor(struct reactor *r)
+void destroy_reactor(struct reactor *r)
 {
    struct child *child = r->first;
    while (child) {
--- a/exercises/react/src/example.c
+++ b/exercises/react/src/example.c
@@ -38,8 +38,7 @@ struct cell {
 
 struct reactor *create_reactor()
 {
-   struct reactor *r = calloc(1, sizeof(struct reactor));
-   return r;
+   return calloc(1, sizeof(struct reactor));
 }
 
 void destroy_reactor(struct reactor *r)
--- a/exercises/react/src/example.c
+++ b/exercises/react/src/example.c
@@ -1,0 +1,211 @@
+#include <stdlib.h>
+#include "react.h"
+
+enum cell_kind {
+   kind_input,
+   kind_compute1,
+   kind_compute2,
+};
+
+struct child {
+   struct cell *cell;
+   struct child *next;
+};
+
+struct cb {
+   callback_id id;
+   callback f;
+   void *obj;
+   struct cb *next;
+};
+
+struct reactor {
+   struct child *first;
+   struct child *last;
+};
+
+struct cell {
+   struct reactor *reactor;
+   int value;
+   enum cell_kind kind;
+   struct cell *input1;
+   struct cell *input2;
+   compute1 compute1;
+   compute2 compute2;
+   struct cb *cb;
+   int callbacks_issued;
+};
+
+struct reactor *create_reactor()
+{
+   struct reactor *r = calloc(1, sizeof(struct reactor));
+   return r;
+}
+
+void free_reactor(struct reactor *r)
+{
+   struct child *child = r->first;
+   while (child) {
+      struct cb *cb = child->cell->cb;
+      while (cb) {
+         struct cb *next_cb = cb->next;
+         free(cb);
+         cb = next_cb;
+      }
+      free(child->cell);
+
+      struct child *next = child->next;
+      free(child);
+      child = next;
+   }
+   free(r);
+}
+
+#define SUCCESS 1
+#define FAIL 0
+
+static int add_child(struct reactor *r, struct cell *cell)
+{
+   struct child *child = calloc(1, sizeof(struct child));
+   if (!child) {
+      return FAIL;
+   }
+   child->cell = cell;
+   if (!r->first) {
+      r->first = child;
+   } else {
+      r->last->next = child;
+   }
+   r->last = child;
+   return SUCCESS;
+}
+
+struct cell *create_input_cell(struct reactor *r, int initial_value)
+{
+   struct cell *c = calloc(1, sizeof(struct cell));
+   if (!c) {
+      return NULL;
+   }
+   if (add_child(r, c) != SUCCESS) {
+      free(c);
+      return NULL;
+   }
+   c->reactor = r;
+   c->kind = kind_input;
+   c->value = initial_value;
+   return c;
+}
+
+struct cell *create_compute1_cell(struct reactor *r, struct cell *input,
+                                  compute1 compute)
+{
+   struct cell *c = calloc(1, sizeof(struct cell));
+   if (!c) {
+      return NULL;
+   }
+   if (add_child(r, c) != SUCCESS) {
+      free(c);
+      return NULL;
+   }
+   c->reactor = r;
+   c->kind = kind_compute1;
+   c->input1 = input;
+   c->compute1 = compute;
+   c->value = compute(cell_value(input));
+   return c;
+}
+
+struct cell *create_compute2_cell(struct reactor *r, struct cell *input1,
+                                  struct cell *input2, compute2 compute)
+{
+   struct cell *c = calloc(1, sizeof(struct cell));
+   if (!c) {
+      return NULL;
+   }
+   if (add_child(r, c) != SUCCESS) {
+      free(c);
+      return NULL;
+   }
+   c->reactor = r;
+   c->kind = kind_compute2;
+   c->input1 = input1;
+   c->input2 = input2;
+   c->compute2 = compute;
+   c->value = compute(cell_value(input1), cell_value(input2));
+   return c;
+}
+
+int cell_value(struct cell *c)
+{
+   return c->value;
+}
+
+static void propagate(struct cell *c)
+{
+   int new_value;
+   switch (c->kind) {
+   case kind_compute1:
+      new_value = c->compute1(cell_value(c->input1));
+      break;
+   case kind_compute2:
+      new_value = c->compute2(cell_value(c->input1), cell_value(c->input2));
+      break;
+   default:
+      return;
+   }
+
+   if (new_value != c->value) {
+      c->value = new_value;
+      for (struct cb * cb = c->cb; cb; cb = cb->next) {
+         cb->f(cb->obj, new_value);
+      }
+   }
+
+}
+
+void cell_value_set(struct cell *c, int new_value)
+{
+   c->value = new_value;
+   struct reactor *r = c->reactor;
+
+   // We take the very naive route of updating all cells.
+   // Traversing the tree and only updating the cells dependent on the just-changed cell is possible,
+   // but requires much more memory management to make each cell aware of its dependents.
+   for (struct child * child = r->first; child; child = child->next) {
+      propagate(child->cell);
+   }
+}
+
+callback_id add_callback(struct cell *c, void *obj, callback f)
+{
+   struct cb *cb = calloc(1, sizeof(struct cb));
+   if (!cb) {
+      return -1;
+   }
+   cb->id = c->callbacks_issued++;
+   cb->next = c->cb;
+   cb->obj = obj;
+   cb->f = f;
+   c->cb = cb;
+   return cb->id;
+}
+
+void remove_callback(struct cell *c, callback_id to_remove)
+{
+   if (!c->cb) {
+      return;
+   } else if (c->cb->id == to_remove) {
+      struct cb *to_free = c->cb;
+      c->cb = c->cb->next;
+      free(to_free);
+      return;
+   }
+   for (struct cb * prev = c->cb, *cb = c->cb->next; cb;
+        prev = prev->next, cb = cb->next) {
+      if (cb->id == to_remove) {
+         prev->next = cb->next;
+         free(cb);
+         return;
+      }
+   }
+}
--- a/exercises/react/src/example.c
+++ b/exercises/react/src/example.c
@@ -91,7 +91,7 @@ struct cell *create_compute1_cell(struct reactor *r, struct cell *input,
    c->kind = kind_compute1;
    c->input1 = input;
    c->compute1 = compute;
-   c->value = compute(cell_value(input));
+   c->value = compute(get_cell_value(input));
    return c;
 }
 
@@ -105,11 +105,11 @@ struct cell *create_compute2_cell(struct reactor *r, struct cell *input1,
    c->input1 = input1;
    c->input2 = input2;
    c->compute2 = compute;
-   c->value = compute(cell_value(input1), cell_value(input2));
+   c->value = compute(get_cell_value(input1), get_cell_value(input2));
    return c;
 }
 
-int cell_value(struct cell *c)
+int get_cell_value(struct cell *c)
 {
    return c->value;
 }
@@ -119,10 +119,11 @@ static void propagate(struct cell *c)
    int new_value;
    switch (c->kind) {
    case kind_compute1:
-      new_value = c->compute1(cell_value(c->input1));
+      new_value = c->compute1(get_cell_value(c->input1));
       break;
    case kind_compute2:
-      new_value = c->compute2(cell_value(c->input1), cell_value(c->input2));
+      new_value =
+          c->compute2(get_cell_value(c->input1), get_cell_value(c->input2));
       break;
    default:
       return;
--- a/exercises/react/src/example.c
+++ b/exercises/react/src/example.c
@@ -165,13 +165,12 @@ void remove_callback(struct cell *c, callback_id to_remove)
 {
    if (!c->cb) {
       return;
-   } else if (c->cb->id == to_remove) {
-      struct cb *to_free = c->cb;
-      c->cb = c->cb->next;
-      free(to_free);
-      return;
    }
-   for (struct cb * prev = c->cb, *cb = c->cb->next; cb;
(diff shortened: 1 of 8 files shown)


exercises/react/test/test_react.c
```diff
@@ -0,0 +1,110 @@
...
+{
+   return x * 30;
+}
+
+static int plus(int x, int y)
+{
+   return x + y;
+}
+
+// TODO: any better way to represent these?
```
petertseng: In particular: This is going to hurt every time I need to use this cb1, since I need to reset the value of `cb_1_called` and `cb1_last`. Plus, there is the test where I will need to have a total of three callbacks. https://github.com/exercism/x-common/blob/master/exercises/react/canonical-data.json#L284

Having to have three sets of globals + function to manipulate them seems too much. I should perhaps allow a `void *` to be passed into each `add_callback`, and the API will call the callback function passing the `void *` back in.

wolf99:
My 2c on the architecture of this; If I have this right, this essentially creates a linked list of compute cells that each have either 1 or 2 descendant input cells? Like so:

07 03 2017 12 16 office lens

Would it make more sense instead to have a list of cell objects, each of which has a list of references to cells that depend on it? This way a cell could be the dependent of multiple cells which could be updated whenever the cell changed by calling each parent cell's assigned calculation function (callback) whenever it's own value is updated.

07 03 2017 12 10 office lens

Not entirely sure how cells that depend on both non-calculated cells and calculated value cells would decide precedence though (e.g. in the case where cell `a` depends on a value in cell `b` _and_ cell `c` where the value in cell `c` _also depends on the value in cell `b`), or handle cicular dependencies... but these may be out of scope for this exercise?

petertseng:
> If I have this right, this essentially creates a linked list of compute cells

Careful! The input cells are also in the linked list

> that each have either 1 or 2 descendant input cells

The pointers may be to other compute cells as well.

That means I would describe it as: A linked of all cells in creation order, where the compute cells have links (which must necessarily be backwards) to their dependencies.

> Would it make more sense instead to have a list of cell objects, each of which has a list of references to cells that depend on it?

I imagine it's certainly *possible* to make such a thing.

I would like issue the following advisories about this:

It is given to us within the constraints of this problem and in the API in react.h that each cell has a maximum of two dependencies.
This means the example could place those dependencies as two pointers within `struct cell`.

The constraints of this problem do not limit the number of dependents a cell can have.
So, if cells keep track of their dependents, this implies one of the two following:

* Each submitted solution has to set an arbitrary limit on the number of dependents each cell can support, like in https://github.com/exercism/xc/blob/master/exercises/anagram/src/anagram.h#L4
* When creating a compute1 cell, we must dynamically allocate a `struct child` so that its parent knows about it. When creating a compute2 cell, we must dynamically allocate two `struct child`. Be careful to take care of the case when the second allocation fails; you don't want to leave a garbage child in the first parent's child list!

In addition, note that for non-commutative functions, each compute2 cell either needs to keep track of which parent is which, or each parent needs to keep track of whether it is parent1 or parent2 for a compute2 cell.
This means that having cells keep track of their dependents (rather than their dependencies) is likely not to be a memory savings, but is likely to be a time savings (we can update only the cells that need to be updated).

It is likely that I will implement such a solution for my personal submission to exercism.io.
It will likely have more code complexity.
As a result, I'm not a maintainer of this track I know it would be rude of me to inflict the maintenance of a higher-complexity example solution upon the maintainers of the track, so I don't do it.
It can be explored if truly desired.
Maybe we will find ways to reduce the complexity.

> cells that depend on both non-calculated cells and calculated value cells

There's no test case for that (you could add it to x-common if it seems important) but there is a test case for diamond dependency (input cell A; compute cell B depends on A; compute cell C depends on A; compute cell D depends on B and D) that serves the same purpose and requires the same question to be asked.

> circular dependencies... but these may be out of scope for this exercise?

While the problem specification does not explicitly forbid these, the API in react.h and in canonical-data.json don't really allow for them, because cells can only depend on cells that were created before themselves, and you cannot change a cell's dependencies after it's created.
If we supposed that it were possible:

* We could have a stable loop. In these cases it might be interesting to see whether the solution terminates after it determines that the state is stable rather than looping forever.
* We could of course have oscillation in the states. In these cases it's unclear what behaviour to expect the system to have. Maybe it would loop forever, or maybe we might expect that the system is expected to signal the oscillation after some number of oscillations (again, increases problem scope).

Further discussion of that should be had at x-common.

petertseng:
EBWODP: Also, the free logic will have to be careful to avoid double-free because each cell can have an unbounded number of dependents and dependencies up to two. With the flat list, it is easy to read the code and understand that there is no double-free.

petertseng:
EBWODP: 

> compute cell D depends on B and D

Wrong. compute cell D depends on B and C.

wolf99:
I see I had initially misunderstood your implementation. As you describe it, your solution definitely is less complex in terms of maintenance (at least from my imagining of my proposed solution). Of course the implementation that allows the simplest maintenance going forward would be best 😃

exercises/react/test/test_react.c
```diff
@@ -0,0 +1,295 @@
...
+   struct reactor *r = create_reactor();
+   struct cell *input = create_input_cell(r, 1);
+   struct cell *output = create_compute1_cell(r, input, plus1);
+
+   TEST_ASSERT_EQUAL_INT(2, cell_value(output));
+
+   free_reactor(r);
+}
+
+static int concat_digits(int a, int b)
```
petertseng: lot of functions scattered throughout the test file. I wonder if they should all be defined at the top, or rather defined close to where they are used.

I defined them as close to where they are used as possible.

exercises/react/src/react.h
```diff
@@ -0,0 +1,29 @@
...
+struct cell *create_compute2_cell(struct reactor *, struct cell *,
+                                  struct cell *, compute2);
+
+int cell_value(struct cell *);
+void cell_value_set(struct cell *, int new_value);
+
+typedef void (*callback) (void *, int);
+typedef int callback_id;
+
+// The callback should be called with the same void * given in add_callback.
```
petertseng: note: my way of dealing with no closure capture. if I didn't do this, I'd have to have at least three functions that do the same thing but on three different global variables, since one test requires having three callbacks.

exercises/react/test/test_react.c
```diff
@@ -0,0 +1,295 @@
...
+   return x - 1;
+}
+
+static int times(int x, int y)
+{
+   return x * y;
+}
+
+void
+test_callbacks_are_only_called_once_even_if_multiple_dependencies_change(void)
```
petertseng: I probably should take this to mean the test name is too long, but it's the name in canonical-data.json, and am unsure how to abbreviate.

exercises/react/test/test_react.c
```diff
@@ -0,0 +1,295 @@
...
+   free_reactor(r);
+}
+
+static int minus(int x, int y)
+{
+   return x - y;
+}
+
+void
+test_callbacks_are_not_called_if_dependencies_change_but_output_value_doesnt_change
```
petertseng: > I probably should take this to mean the test name is too long, but it's the name in canonical-data.json, and am unsure how to abbreviate.

same.

exercises/react/src/example.c
```diff
@@ -0,0 +1,211 @@
...
+   for (struct child * child = r->first; child; child = child->next) {
+      propagate(child->cell);
+   }
+}
+
+callback_id add_callback(struct cell *c, void *obj, callback f)
+{
+   struct cb *cb = calloc(1, sizeof(struct cb));
+   if (!cb) {
+      return -1;
```
petertseng: this isn't documented in the header (I explained the significance of the `void*`). But the tests will never check for what happens if a dynamic allocation fails (and it'd be really hard to do so unless we LD_PRELOAD in a calloc that is designed to fail), so it's unclear what to do.

exercises/react/src/react.h
```diff
@@ -0,0 +1,29 @@
...
+struct cell *create_input_cell(struct reactor *, int initial_value);
+struct cell *create_compute1_cell(struct reactor *, struct cell *, compute1);
+struct cell *create_compute2_cell(struct reactor *, struct cell *,
+                                  struct cell *, compute2);
+
+int cell_value(struct cell *);
+void cell_value_set(struct cell *, int new_value);
+
+typedef void (*callback) (void *, int);
+typedef int callback_id;
```
petertseng: I suppose I could say `struct callback_id` and let the student define it, but since it'd then be an incomplete type `add_callback` would have to return `struct callback *` and I didn't feel like doing more memory management for that.

exercises/react/test/test_react.c
petertseng: This gets `indent.sh` in an oscillating loop.

Run it with

```c
void
test_callbacks_are_not_called_if_dependencies_change_but_output_value_doesnt_change
(void)
```

and `indent` will change the code to:


```c
void test_callbacks_are_not_called_if_dependencies_change_but_output_value_doesnt_change(void)
```

Run it again, and `indent` changes it back. So I will be forced to shorten the name.

exercises/react/src/react.h
```diff
@@ -0,0 +1,29 @@
...
+
+struct reactor;
+struct cell;
+
+typedef int (*compute1) (int);
+typedef int (*compute2) (int, int);
+
+struct reactor *create_reactor();
+// free_reactor should free all cells created under that reactor.
+void free_reactor(struct reactor *);
```
ryanplusplus: How about `destroy_reactor` since that's more clearly the opposite of create?

exercises/react/test/test_react.c
ryanplusplus: I prefer them this way: as close to use as possible.

exercises/react/test/test_react.c
```diff
@@ -0,0 +1,289 @@
...
+
+   cell_value_set(input, 3);
+   TEST_ASSERT_EQUAL_INT(1, cbinfo.times_called);
+   TEST_ASSERT_EQUAL_INT(4, cbinfo.last_value);
+
+   free_reactor(r);
+}
+
+static void cb_noop(void *obj __attribute__ ((unused)), int v
+                    __attribute__ ((unused)))
```
ryanplusplus: Instead of using the GNU extension `__attribute__`, how about declaring these unused like this:

```c
static void cb_noop(void *obj, int v)
{
   (void)obj;
   (void)v;
}
```

exercises/react/test/test_react.c
```diff
@@ -0,0 +1,289 @@
...
+
+   free_reactor(r);
+}
+
+struct cbinfo {
+   int last_value;
+   int times_called;
+};
+
+static void cb_record_last_value(void *obj, int v)
```
ryanplusplus: How about `cb_spy` or something similar? I think that `spy` more succinctly communicates what this does and is an idiomatic term for this type of test double.

exercises/react/src/react.h
ryanplusplus: 👍

petertseng:
Hello, changes are coming in as separate commits so that it's clear what's changed since last review. squashing on merge seems appropriate once everything is settled

exercises/react/src/react.h
```diff
@@ -0,0 +1,29 @@
...
+#define REACT_H
+
+struct reactor;
+struct cell;
+
+typedef int (*compute1) (int);
+typedef int (*compute2) (int, int);
+
+struct reactor *create_reactor();
+// destroy_reactor should free all cells created under that reactor.
```
petertseng: Oh, since I only replaced `free_reactor` with `destroy_reactor`, this sentence still says "free all cells". Perhaps it should say "destroy all cells" instead?

Of course, the test doesn't enforce such a thing, so there is no force behind that "should". I believe (valgrind) the example does. But for submissions to the website there's no guarantee anyone ran valgrind.

ryanplusplus: Well, to me there are two separate levels of abstraction: malloc/free at the implementation level and create/destroy at the interface level (which elides certain details about what's actually happening). This is an odd case where we're at the boundary of those two concerns. However, since these are instructions for the implementer, I'm okay with it still saying "free".

exercises/react/src/example.c
```diff
@@ -0,0 +1,211 @@
...
+   compute1 compute1;
+   compute2 compute2;
+   struct cb *cb;
+   int callbacks_issued;
+};
+
+struct reactor *create_reactor()
+{
+   struct reactor *r = calloc(1, sizeof(struct reactor));
+   return r;
```
ryanplusplus: Is the local `r` necessary?

petertseng: Apparently not!

Holdover from the times when I was using `malloc` and manually setting fields.

exercises/react/src/example.c
ryanplusplus: I had the same thought in the `create_blah_cell` functions. I hate having untested code, but Unity doesn't really give you the tools to test this unless you inject `calloc`. I'm okay with either leaving these or, if we feel like this isn't really the point of the exercise and we have enough complexity already, I'm okay with removing the allocation checks.

exercises/react/src/example.c
```diff
@@ -0,0 +1,210 @@
...
+struct cb {
+   callback_id id;
+   callback f;
+   void *obj;
+   struct cb *next;
+};
+
+struct reactor {
+   struct child *first;
+   struct child *last;
```
ryanplusplus: Is this necessary? It looks like it's a write-only field.

exercises/react/src/example.c
```diff
@@ -0,0 +1,211 @@
...
+};
+
+struct cell {
+   struct reactor *reactor;
+   int value;
+   enum cell_kind kind;
+   struct cell *input1;
+   struct cell *input2;
```
ryanplusplus: Personally, I'm okay with casting since it's all internal to this module, but I don't think the extra complexity is warranted. Let's leave it as-is to keep the solution more approachable.

exercises/react/src/example.c
petertseng: Since there are no tests for it, each student can do as desired in each individual solution.

This example solution has chosen to check each allocation so as not to mistakenly try to offset off of NULL if an allocation fails, despite no enforcement of that behaviour. I'll add a commit that removes all such checks so we can examine what it looks like. If we like it great, if not, then we revert that commit.

exercises/react/src/example.c
```diff
@@ -0,0 +1,210 @@
...
+}
+
+void cell_value_set(struct cell *c, int new_value)
+{
+   c->value = new_value;
+   struct reactor *r = c->reactor;
+
+   // We take the very naive route of updating all cells.
+   // Traversing the tree and only updating the cells dependent on the just-changed cell is possible,
+   // but requires much more memory management to make each cell aware of its dependents.
```
ryanplusplus: I think this is really against the spirit of reactive programming, but it is a pretty elegant (if non-performant) solution. If we had more of the basic building blocks available to us (like STL) then I'd say we should definitely target a more robust solution, but I feel like this would quickly spin out of control in no-dependencies C.

Thoughts?

exercises/react/src/example.c
```diff
@@ -0,0 +1,210 @@
...
+      return;
+   }
+   for (struct cb * prev = c->cb, *cb = c->cb->next; cb;
+        prev = prev->next, cb = cb->next) {
+      if (cb->id == to_remove) {
+         prev->next = cb->next;
+         free(cb);
+         return;
+      }
+   }
```
ryanplusplus: Could you combine the `for` and `else if` cases by clever use of a dummy starting node?

exercises/react/src/example.c
ryanplusplus: I think that I favor the commits that remove the checks since they make the example more instructive regarding the exercise content by removing noise (even if the solution is less "correct" from a memory management standpoint).

ryanplusplus:
Overall this looks great. I was afraid the complexity for this would spiral out of control in C, but it's really quite nice. Well done!

exercises/react/src/example.c
petertseng: It helps because if the list is populated, then we do `r->last->next = child` (since that `r->last` is a read it's not write-only)

It's true that it'd be possible to elide this field, and just find `last` by traversing the list - though saving this memory would mean more time (and also more code! since we have to write the code to traverse and find last).

An alternative question to ask is: "Why must newly-created cells be placed at the *tail* of the list? Could we delete this *last* field if we put newly-created cells at the *head* of the list?"

I think the answer to this question is related to `destroy_reactor` - I think I needed to destroy a cell before all its dependents, but now I think that might not be true, since I can just destroy the `struct child` object, only destroy the `cell` when I come to it on the list.

so, I'll see if I can keep only one end of this list, or report back with a reason why I must keep both ends.

ryanplusplus: Whoops, I didn't look closely enough and only took note of the fact that `->last` was on the left side of the assignment.

I think you could add at the head of the list, but then you'd have to traverse in reverse order when you do an update or dependencies won't be updated before dependents. My 2 cents: leave it as-is.

exercises/react/src/example.c
petertseng: I initially tried to take the path mentioned (only update the necessary cells) and got pretty mired down in memory management. I still do intend to attempt such a solution in my submission to the website, but I may need a lot more time to do it. If I complete it and it looks reasonable, I can then propose it be the example solution.

In general, the tests do not check for performance... they **could**, if we had a compute function that keeps track of how often it is called and then enforce that it is only called a certain number of times (and not called if its input cell doesn't change). If I added such a test, I would probably deal with it in this solution by adding a dirty bit on each cell that tracks whether to call its compute function again. Of course, it is still iterating through the entire list of cells, but it's not possible to test that aspect of performance.

In general it seems hard to test performance and/or implementation details like that.

There's a larger question of "What standard must example solutions be held to? Surely they should be good style and idiomatic for the language. But Must they be efficient?" I'm not sure any track has come up with a definitive answer, but I see:

* https://github.com/exercism/xhaskell#example-solution: It doesn't need to be perfect or very elegant. But it should be efficient enough for the test suite to finish in only a few seconds.
* https://github.com/exercism/xgo#example-solutions: there is less need for this code to be a model of style, expression and readability, or to use the best algorithm. Examples can be plain, simple, concise, even naïve, as long as they are correct.

I do predict that the more efficient solution will be have more complex code, so I biased toward keeping it simple for the example solution since it may be the case that track maintainers may have to maintain/make changes in the future.

exercises/react/src/example.c
petertseng: Oh you're right, I forgot that I need the ordering both in `destroy_reactor` and `cell_set_value`. Yeah, probably have to leave as-is then.

exercises/react/src/react.h
```diff
@@ -0,0 +1,29 @@
...
+void destroy_reactor(struct reactor *);
+
+struct cell *create_input_cell(struct reactor *, int initial_value);
+struct cell *create_compute1_cell(struct reactor *, struct cell *, compute1);
+struct cell *create_compute2_cell(struct reactor *, struct cell *,
+                                  struct cell *, compute2);
+
+int cell_value(struct cell *);
+void cell_value_set(struct cell *, int new_value);
```
petertseng: naming question: `cell_value_set` or `set_cell_value`? Note that current name is different from above `destroy_reactor` and `create_input_cell` which are all `verb_thing`.

exercises/react/src/example.c
petertseng: nice idea. let me know if you think the explanatory comment is unnecessary.

exercises/react/src/example.c
```diff
@@ -0,0 +1,181 @@
...
+      return;
+   }
+   // this dummy starting node reduces code duplication,
+   // in case the ID to remove is at the head of the list.
+   struct cb dummy;
+   dummy.next = c->cb;
+   for (struct cb * prev = &dummy, *cb = c->cb; cb;
+        prev = prev->next, cb = cb->next) {
+      if (cb->id == to_remove) {
+         prev->next = cb->next;
```
petertseng: hold on, but this doesn't deal with the case where the most-recently-added callback is removed (c->cb should get set but isn't). guess the tests don't check that.

exercises/react/src/react.h
ryanplusplus: `get_cell_value` and `set_cell_value` seem like reasonable names and let us settle on `verb_thing`.

exercises/react/src/example.c
ryanplusplus: Should just be a quick `if(c->cb == cb) c->cb = cb->next;`, right?

petertseng: This is true (and it has been added), but I added a test that I thought would fail without this code, and I didn't see a failure like I thought I would. I'm not sure why.

petertseng: Never mind, I got the test to fail as I expected it would. Error on my side (react.c vs example.c)

ryanplusplus: Oh good. I was just about to add a comment expressing my confusion.

petertseng: note for me later: I'll see about adding this case to x-common.

The other case that this track has that x-common doesn't have (test_compute_cells_dont_access_callback_obj) would be difficult to express in x-common

petertseng:
just a second. I forgot to add a case (input_cells_value_can_be_set)

ryanplusplus:
Looks good to me.

@palisaide @patricksjackson any additional comments? Otherwise I'm going to go ahead and merge this.

ryanplusplus:
Okay, let's go ahead and merge this. We can always make improvements later if additional suggestions crop up.

PR #128, opened 2017-03-06
Nucleotide Count - Exercise Added

StevenRoot:
Closed the work in progress pull request and created this one after finishing the basic implementation.

### DIFF
--- a/config.json
+++ b/config.json
@@ -207,6 +207,15 @@
         "preprocessor (x-macros in test)"
      ]
   }, {
+        "difficulty": 2,
+        "slug": "nucleotide-count",
+        "topics": [
+           "control-flow (loops, switch, if-statements)",
+           "strings",
+           "text formatting",
+           "memory management"
+        ]
+  }, {
      "difficulty": 2,
      "slug": "robot-simulator",
      "topics": [
--- a/config.json
+++ b/config.json
@@ -207,14 +207,14 @@
         "preprocessor (x-macros in test)"
      ]
   }, {
-        "difficulty": 2,
-        "slug": "nucleotide-count",
-        "topics": [
-           "control-flow (loops, switch, if-statements)",
-           "strings",
-           "text formatting",
-           "memory management"
-        ]
+     "difficulty": 2,
+     "slug": "nucleotide-count",
+     "topics": [
+        "control-flow (loops, switch, if-statements)",
+        "strings",
+        "text formatting",
+        "memory management"
+     ]
   }, {
      "difficulty": 2,
      "slug": "robot-simulator",
--- a/exercises/nucleotide-count/src/example.c
+++ b/exercises/nucleotide-count/src/example.c
@@ -2,6 +2,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include "nucleotide_count.h"
 
 char *count(const char *dna_strand)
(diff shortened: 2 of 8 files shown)


exercises/nucleotide-count/src/example.c
```diff
@@ -0,0 +1,44 @@
...
+#include "nucleotide_count.h"
+
+char *count(const char *dna_strand)
+{
+   bool invalidChar = false;
+   uint16_t index;
+   uint16_t count_a = 0;
+   uint16_t count_c = 0;
+   uint16_t count_g = 0;
+   uint16_t count_t = 0;
```
wolf99: Very very minor adjustment: maybe count names could instead be of the form `x_count`? This way `count_t` would no longer seem like a type name (`_t` suffix is often used to represent type names), yet the names would still make sense 😃

config.json
```diff
@@ -207,6 +207,15 @@
...
      ]
   }, {
+        "difficulty": 2,
+        "slug": "nucleotide-count",
+        "topics": [
+           "control-flow (loops, switch, if-statements)",
+           "strings",
+           "text formatting",
+           "memory management"
+        ]
```
ryanplusplus: Formatting is out of whack on this guy

exercises/nucleotide-count/src/example.c
```diff
@@ -0,0 +1,44 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include "nucleotide_count.h"
+
+char *count(const char *dna_strand)
+{
+   bool invalidChar = false;
+   uint16_t index;
```
ryanplusplus: I realize that it's way overkill, but why not `size_t` here? This is guaranteed to be capable of handling any string whereas `uint16_t` seems a bit arbitrary.

exercises/nucleotide-count/src/example.h
```diff
@@ -0,0 +1,6 @@
+#ifndef  _NUCLEOTIDE_COUNT_H
```
ryanplusplus: Extra space here

exercises/nucleotide-count/src/example.c
```diff
@@ -0,0 +1,44 @@
...
+         break;
+      }
+   }
+
+   if (!invalidChar) {
+      sprintf(count_results, "A:%u C:%u G:%u T:%u", nucleotide_A_count,
+              nucleotide_C_count, nucleotide_G_count, nucleotide_T_count);
+   }
+   return count_results;
+}
```
ryanplusplus: There's a mix of camel case and snake case here. I think all you need to do is make `invalidChar` into `invalid_char` to resolve this.

ryanplusplus:
Looks good, just a few minor comments

config.json
StevenRoot: fixed

exercises/nucleotide-count/src/example.c
StevenRoot: -- agnostic changed -
complicates printf %zd versus %u -- but does allow you to avoid stdint.h. 
memory size much larger than needed for tests...

exercises/nucleotide-count/src/example.c
StevenRoot: changed

ryanplusplus:
Sorry for the late follow-up, looks great!

Issue #127, opened 2017-03-06
Should CI run and check valgrind?


wolf99:
As per title, should solutions be checked by valgrind for memory leaks, overrruns and possibly also thread errors (if any examples are added that use thread).

ryanplusplus:
I think adding valgrind would be excellent

wolf99:
Will look into attempting to add this when I next have time so! :smile:

PR #129, opened 2017-03-10
Palindrome Product exercise added

Guntau:
[description](https://github.com/exercism/x-common/blob/master/exercises/palindrome-products/description.md).
I did a first implementation of the Palindrome Product exercise, but there is surely some things that can be improved.

### DIFF
--- a/exercises/palindrome/src/example.c
+++ b/exercises/palindrome/src/example.c
@@ -15,7 +15,7 @@ bool isPalindromeProduct(int a, int b)
 
 Pair getPalindromeProduct(int min, int max)
 {
-   Pair ret;
+   Pair ret = {0};
    bool largest = false;
 
    for (int i = min; i <= max; i++) {
--- a/exercises/palindrome/src/example.c
+++ b/exercises/palindrome/src/example.c
@@ -15,7 +15,9 @@ bool isPalindromeProduct(int a, int b)
 
 Pair getPalindromeProduct(int min, int max)
 {
-   Pair ret = { 0 };
+   Pair ret;
+   ret.small_palind = 0;
+   ret.larg_palind = 0;
    bool largest = false;
 
    for (int i = min; i <= max; i++) {
--- a/exercises/palindrome/src/example.c
+++ b/exercises/palindrome/src/example.c
@@ -19,7 +19,6 @@ bool isPalindromeProduct(int a, int b)
          return false;
       }
    }
-
    return true;
 }
 
@@ -37,7 +36,6 @@ Pair *getPalindromeFactors(int palindrome)
          }
       }
    }
-
    assert((tableP = malloc(sizeTableP * sizeof(Pair))) != NULL);
 
    // for each possibility, if i*j is a product of the palindrome, add it to the return
@@ -50,7 +48,6 @@ Pair *getPalindromeFactors(int palindrome)
          }
       }
    }
-
    return tableP;
 }
 
--- a/exercises/palindrome/src/example.c
+++ b/exercises/palindrome/src/example.c
@@ -1,0 +1,46 @@
+#include "palindrome.h"
+#include "stdio.h"
+#include "assert.h"
+#include "string.h"
+
+
+int isPalindrome(int a, int b){
+
+   char number[9];
+   int i,length;
+
+   /* negative numbers not allowed */
+   if(a*b<0)
+      return 1;
+
+
+   sprintf(number,"%d",a*b);
+   length = strlen(number);
+
+   for(i = 0; i< length/2; i++){
+
+      if(number[i] != number[length-i-1]){
+         return 1;
+      }
+   }
+   return 0;
+}
+
+Paire* palindromeFactors(int min, int max){
+   
+   Paire* tableP = malloc(sizeof(Paire));
+   Paire stock;
+   int i,j,tablePcount = 0;
+   for( i = min; i <= max; i++){
+      for( j = min; j <= max; j++){
+         if(isPalindrome(i,j)==0){
+           stock.a = i;
+           stock.b = j;
+           tableP[tablePcount] = stock;
+           tablePcount++;
+           realloc(tableP,sizeof(Paire);
+         }
+      }
+   }
+   return tableP;
+}
--- a/exercises/palindrome/src/example.c
+++ b/exercises/palindrome/src/example.c
@@ -2,7 +2,7 @@
 #include "stdio.h"
 #include "assert.h"
 #include "string.h"
-
+#include "stdlib.h"
 
 int isPalindrome(int a, int b){
 
@@ -28,7 +28,8 @@ int isPalindrome(int a, int b){
 
 Paire* palindromeFactors(int min, int max){
    
-   Paire* tableP = malloc(sizeof(Paire));
+   Paire* tableP;
+   assert((tableP = malloc(sizeof(Paire)))!=NULL);
    Paire stock;
    int i,j,tablePcount = 0;
    for( i = min; i <= max; i++){
@@ -38,7 +39,7 @@ Paire* palindromeFactors(int min, int max){
            stock.b = j;
            tableP[tablePcount] = stock;
            tablePcount++;
-           realloc(tableP,sizeof(Paire);
+           assert(realloc(tableP,sizeof(Paire))!=NULL);
          }
       }
    }
--- a/exercises/palindrome/src/example.c
+++ b/exercises/palindrome/src/example.c
@@ -4,44 +4,82 @@
 #include "string.h"
 #include "stdlib.h"
 
-int isPalindrome(int a, int b){
+/* Return 0 if the product of a and b is a palindrome, 1 otherwise */
+int isPalindrome(int a, int b)
+{
 
-   char number[9];
-   int i,length;
+   char number[10];
+   int i, length;
 
    /* negative numbers not allowed */
-   if(a*b<0)
+   if (a * b < 0)
       return 1;
 
-
-   sprintf(number,"%d",a*b);
+   sprintf(number, "%d", a * b);
    length = strlen(number);
 
-   for(i = 0; i< length/2; i++){
-
-      if(number[i] != number[length-i-1]){
+   for (i = 0; i < length / 2; i++) {
+      if (number[i] != number[length - i - 1]) {
          return 1;
       }
    }
    return 0;
 }
 
-Paire* palindromeFactors(int min, int max){
-   
-   Paire* tableP;
-   assert((tableP = malloc(sizeof(Paire)))!=NULL);
-   Paire stock;
-   int i,j,tablePcount = 0;
-   for( i = min; i <= max; i++){
-      for( j = min; j <= max; j++){
-         if(isPalindrome(i,j)==0){
-           stock.a = i;
-           stock.b = j;
-           tableP[tablePcount] = stock;
-           tablePcount++;
-           assert(realloc(tableP,sizeof(Paire))!=NULL);
+/* return the factors of a palindrome */
+Paire *getPalindromeFactors(int palindrome)
+{
+
+   Paire *tableP;
+   int i, j, sizeTableP = 0, tablePcount = 0;
+
+   /* Compute the size of the array */
+   for (i = 1; i <= palindrome; i++) {
+      for (j = 1; j <= palindrome; j++) {
+         if (i * j == palindrome) {
+            sizeTableP++;
          }
       }
    }
+
+   assert((tableP = malloc(sizeTableP * sizeof(Paire))) != NULL);
+
+   // for each possibility, if i*j is a product of the palindrome, add it to the return
+   for (i = 1; i <= palindrome; i++) {
+      for (j = 1; j <= palindrome; j++) {
+         if (i * j == palindrome) {
+            tableP[tablePcount].a = i;
+            tableP[tablePcount].b = j;
+            tablePcount++;
+         }
+      }
+   }
+
    return tableP;
+}
+
+/* return a struct of PalindromeProduct, containing a palindrome and an array of his factor */
+PalindromeProduct *getPalindromeProduct(int min, int max)
+{
+
+   int i, j, retPosition = 0;
+
+   // Creating an array
+   PalindromeProduct *ret;
+   assert((ret = malloc(2 * sizeof(PalindromeProduct))) != NULL);
+
+   // for each possibility, verify if it's a palindrome
+   for (i = min; i <= max; i++) {
+      for (j = min; j <= max; j++) {
+         if (isPalindrome(i, j) == 0) {
+
+            ret[retPosition].palindrome = i * j;
+            ret[retPosition].factors = getPalindromeFactors(i * j);
+
+            // set the position at the last element of the array, once the first palindrome is found
+            retPosition = 1;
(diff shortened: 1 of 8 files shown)


exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,85 @@
+#include "palindrome.h"
+#include "stdio.h"
+#include "assert.h"
+#include "string.h"
+#include "stdlib.h"
+
+/* Return 0 if the product of a and b is a palindrome, 1 otherwise */
+int isPalindrome(int a, int b)
+{
+
```
ryanplusplus: Can you remove this extra whitespace?

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,85 @@
+#include "palindrome.h"
+#include "stdio.h"
```
ryanplusplus: Make sure to use `<stdio.h>`, `<string.h>`, and `<stdlib.h>`

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,85 @@
...
+#include "assert.h"
+#include "string.h"
+#include "stdlib.h"
+
+/* Return 0 if the product of a and b is a palindrome, 1 otherwise */
+int isPalindrome(int a, int b)
```
ryanplusplus: Instead of using an `int` for your result, how about using `stdbool.h` so that you can use `true` and `false`?

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,85 @@
...
+         return 1;
+      }
+   }
+   return 0;
+}
+
+/* return the factors of a palindrome */
+Paire *getPalindromeFactors(int palindrome)
+{
+
```
ryanplusplus: Seemingly extra newline here

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,85 @@
...
+      }
+   }
+
+   return tableP;
+}
+
+/* return a struct of PalindromeProduct, containing a palindrome and an array of his factor */
+PalindromeProduct *getPalindromeProduct(int min, int max)
+{
+
```
ryanplusplus: Seemingly extra newline here

exercises/palindrome/src/example.h
```diff
@@ -0,0 +1,25 @@
+#ifndef PALINDROME_H
+#define PALINDROME_H
+
+/* structure containing a pair of integer */
+typedef struct paire {
+   int a;
+   int b;
+} Paire;
```
ryanplusplus: `Pair` instead of `Paire`?

ryanplusplus:
We're generally using the [Linux kernel style guide](https://01.org/linuxgraphics/gfx-docs/drm/process/coding-style.html)? Can you reference that and other solutions to make sure that your naming matches the rest of the exercises?

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,85 @@
...
+   // for each possibility, verify if it's a palindrome
+   for (i = min; i <= max; i++) {
+      for (j = min; j <= max; j++) {
+         if (isPalindrome(i, j) == 0) {
+
+            ret[retPosition].palindrome = i * j;
+            ret[retPosition].factors = getPalindromeFactors(i * j);
+
+            // set the position at the last element of the array, once the first palindrome is found
+            retPosition = 1;
```
ryanplusplus: Doesn't this just mean that you're now searching for the max? This seems a bit convoluted and it obscures what's actually going on here.

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,85 @@
...
+{
+
+   int i, j, retPosition = 0;
+
+   // Creating an array
+   PalindromeProduct *ret;
+   assert((ret = malloc(2 * sizeof(PalindromeProduct))) != NULL);
+
+   // for each possibility, verify if it's a palindrome
+   for (i = min; i <= max; i++) {
```
ryanplusplus: Why not declare `i` and `j` here?

```c
for (int i = min; ...
  for (int j = min; ...
```

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,85 @@
...
+      for (j = min; j <= max; j++) {
+         if (isPalindrome(i, j) == 0) {
+
```
ryanplusplus: Seemingly extra newline here

exercises/palindrome/test/test_palindrome.c
```diff
@@ -0,0 +1,50 @@
...
+
+void test_palindrome_odd_fail(void)
+{
+   TEST_ASSERT_EQUAL_INT(1, isPalindrome(10, 12));
+}
+
+// palindromeFactors tests
+void test_palindromes_list(void)
+{
+
```
ryanplusplus: Seemingly extra newline

exercises/palindrome/src/example.h
```diff
@@ -0,0 +1,25 @@
...
+typedef struct palindProduct {
+   int palindrome;
+   Paire *factors;
+} PalindromeProduct;
+
+/* Return 0 if the product of a and b is a palindrome, 1 otherwise */
+int isPalindrome(int a, int b);
+
+/* return the Paire of factors that give the palindrome ( see Paire) */
+Paire *palindromeFactors(int palindrome);
```
ryanplusplus: This function is never tested. Is it necessary to make it public?

exercises/palindrome/src/example.h
```diff
@@ -0,0 +1,25 @@
...
+} Paire;
+
+/* structure containing a palindrome and an array of his factors */
+typedef struct palindProduct {
+   int palindrome;
+   Paire *factors;
+} PalindromeProduct;
+
+/* Return 0 if the product of a and b is a palindrome, 1 otherwise */
+int isPalindrome(int a, int b);
```
ryanplusplus: Is there any reason that this function is public? It seems a bit unnecessary and on Exercism testing intermediate functions is generally frowned upon.

exercises/palindrome/src/example.h
```diff
@@ -0,0 +1,25 @@
...
+/* structure containing a pair of integer */
+typedef struct paire {
+   int a;
+   int b;
+} Paire;
+
+/* structure containing a palindrome and an array of his factors */
+typedef struct palindProduct {
+   int palindrome;
+   Paire *factors;
```
ryanplusplus: How is the client to know how many factors there were?

exercises/palindrome/test/test_palindrome.c
```diff
@@ -0,0 +1,50 @@
...
+   PalindromeProduct *palind = getPalindromeProduct(1, 9);
+
+   TEST_ASSERT_EQUAL_INT(1, palind[0].palindrome);
+   TEST_ASSERT_EQUAL_INT(1, palind[0].factors[0].a);
+   TEST_ASSERT_EQUAL_INT(1, palind[0].factors[0].b);
+
+   TEST_ASSERT_EQUAL_INT(9, palind[1].palindrome);
+   TEST_ASSERT_EQUAL_INT(1, palind[1].factors[0].a);
+   TEST_ASSERT_EQUAL_INT(9, palind[1].factors[0].b);
+}
```
ryanplusplus: This is the only test for `getPalindromeProduct` and I think we need significantly more testing of this function. For example, this doesn't show that the result can contain multiple factors for each palindrome.

exercises/palindrome/src/example.h
Guntau: I wanted to test every function i used,, but it might be too much. 
`
void test_palindrome_even(void)
{
   TEST_ASSERT_EQUAL_INT(0, isPalindrome(91, 99));
}`
I'll remove these tests

exercises/palindrome/src/example.h
Guntau: I'm changing it

exercises/palindrome/src/example.c
Guntau: I used to declare all my variables at the beginning of the function, to separate them from the algorithm. I'm changing it

exercises/palindrome/src/example.c
Guntau: I personally never use `stdbool.h`, but i can change it

Guntau:
I'll change for a Boolean within a few hours

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,75 @@
...
+   /* Compute the size of the array */
+   for (int i = 1; i <= palindrome; i++) {
+      for (int j = 1; j <= palindrome; j++) {
+         if (i * j == palindrome) {
+            sizeTableP++;
+         }
+      }
+   }
+   assert((tableP = malloc(sizeTableP * sizeof(Pair))) != NULL);
```
patricksjackson: This is problematic. If you compiled a release build of this file you'd never allocate anything.

When you turn off asserts for a release build, the expression contained in it never executes, so the malloc would never be run. It's best to avoid doing anything inside an assert that has side-effects.

It's probably best to allocate the memory and return null if it fails.

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,75 @@
...
+   }
+   return tableP;
+}
+
+/* return a struct of PalindromeProduct, containing a palindrome and an array of his factor */
+PalindromeProduct *getPalindromeProduct(int min, int max)
+{
+   // Creating an array
+   PalindromeProduct *ret;
+   assert((ret = malloc(2 * sizeof(PalindromeProduct))) != NULL);
```
patricksjackson: Same as above, don't allocate inside an assert.

exercises/palindrome/src/example.h
```diff
@@ -0,0 +1,19 @@
+#ifndef PALINDROME_H
+#define PALINDROME_H
+
+/* structure containing a pair of integer */
+typedef struct pair {
```
patricksjackson: AFAIK, you don't need `pair` in this line. Writing `typedef struct {` has the exact same effect.

exercises/palindrome/test/test_palindrome.c
```diff
@@ -0,0 +1,24 @@
...
+   TEST_ASSERT_EQUAL_INT(1, palind[0].factors[0].b);
+   /* largest palindrome */
+   TEST_ASSERT_EQUAL_INT(9, palind[1].palindrome);
+   TEST_ASSERT_EQUAL_INT(1, palind[1].factors[0].a);
+   TEST_ASSERT_EQUAL_INT(9, palind[1].factors[0].b);
+}
+
+int main(void)
+{
+   UnityBegin("test/test_space_age.c");
```
patricksjackson: Make sure this points to the right file.

exercises/palindrome/src/example.c
```diff
@@ -0,0 +1,75 @@
...
+#include <stdlib.h>
+#include <stdbool.h>
+
+/* Return true if the product of a and b is a palindrome, false otherwise */
+bool isPalindromeProduct(int a, int b)
+{
+   char number[10];
+   int length;
+
+   sprintf(number, "%d", a * b);
```
patricksjackson: By the way, you could avoid dealing with strings entirely (getting rid of several includes and speeding this up a lot) by just reversing the digits of the number and then comparing.

```
int rev(int a) {
  int ret = 0;
  while (a) {
    ret *= 10;
    ret += a%10;
    a /= 10;
  }
  return ret;
}
```
Then this function is just `return (a*b) == rev(a*b);`

patricksjackson:
This problem seems very complex. Did you base it off a different languages implementation?

I don't understand why you are returning the smallest palindrome, the largest palindrome, and then _every_ factor of those palindromes (actually, your code returns every factor _twice_).

This would be a lot simpler if you just returned the smallest and largest palindrome products generated from the range in a pair.

This example also never frees any of the allocated memory, so if we add the valgrind checks as suggested in #127 this will fail.

Also, please add several more tests.

Guntau:
I wanted to keep every factors of the palindromes but it's a lot more complicated than just returning the smallest and largest palindrome.

exercises/palindrome/test/test_palindrome.c
```diff
@@ -0,0 +1,33 @@
+#include "vendor/unity.h"
+#include "../src/palindrome.h"
+#include <stdlib.h>
+
+// palindromeFactors tests
```
ryanplusplus: Is this necessary?

exercises/palindrome/test/test_palindrome.c
```diff
@@ -0,0 +1,33 @@
...
+void test_palindromes_list(void)
+{
+   Pair palind = getPalindromeProduct(1, 9);
+
+   /* smallest palindrome : 1 */
```
ryanplusplus: Are these comments necessary? The code itself seems to explain this quite well.

ryanplusplus:
Just a few small issues left. I really like how the simplification that @patricksjackson suggested turned out.

exercises/palindrome/test/test_palindrome.c
ryanplusplus: I think this still needs to be taken care of

exercises/palindrome/test/test_palindrome.c
```diff
@@ -0,0 +1,29 @@
...
+   TEST_ASSERT_EQUAL_INT(1, palind.small_palind);
+   /* largest palindrome : 9 */
+   TEST_ASSERT_EQUAL_INT(9, palind.larg_palind);
+}
+
+void test_palindromes_list2(void)
+{
+   Pair palind = getPalindromeProduct(10, 99);
+   /* smallest palindrome : 121 */
```
ryanplusplus: I think these comments still need to be removed

Guntau:
Should I give default values for elements of the Pair in `getPalindromeProduct` function ? In case there is no palindrome in the range given

wolf99:
@Guntau Maybe `Pair ret = {0, 0};`?

ryanplusplus:
Hey @Guntau, everything is looking great but I still need you to fix up this line in your tests:

```c
UnityBegin("test/test_space_age.c");
```

This looks like a simple copy-paste issue and once it's fixed I will go ahead and merge.

kytrinyx:
This is throwing an error during deploy because we don't have an exercise definition named `palindrome`.

Would you rename the directory to `palindrome-products`? https://github.com/exercism/x-common/blob/master/exercises/palindrome-products/metadata.yml

ryanplusplus:
@kytrinyx sure, sorry about that. Should `configlet` have caught this?

ryanplusplus:
@kytrinyx fixed, sorry for the trouble

kytrinyx:
> Should configlet have caught this?

Yeah, kind of. I'm torn on how to make it happen though, because it would have to do network i/o to get the list of projects. Or a git submodule to include x-common, which I think is really dirty in this case.

Maybe we could do a thing where configlet only checks this if it has network access and can talk to the API.

ryanplusplus:
Ah, got it. I thought this type of check was already in there.

kytrinyx:
Yeah, I opened an issue about it almost 6 months ago (https://github.com/exercism/configlet/issues/10), but I haven't decided on how to move forward, because I don't like the idea of requiring network access.

PR #130, opened 2017-03-12
Link to first page of article

wgj:
The edit is to point the link to the root of the article, instead of the second page. I thought the article was really hard to follow, and poorly written. Turns out I was on the second page, which wasn't clear until getting to the end of the article.

### DIFF
--- a/docs/LEARNING.md
+++ b/docs/LEARNING.md
@@ -28,7 +28,7 @@ written for you.  You progress through the exercise one test at a time,
 writing an implementation incrementally to satisfy the tests.
 
 If you are new to test-driven development, it is strongly recommended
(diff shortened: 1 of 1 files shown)


wolf99:
Umm, not sure this is right... the first page is most of what deals with the Unity test framework - which is the one in use. Page two only deals with it's output (admittedly the main concern for anyone just doing the exercises) and CppUTest, the latter of which makes up the majority of the page and likely to be _even more_ confusing as it is not the framework being used.

wgj:
Figured that it was done on purpose, that's why I shared my experience as a reader. Feel free to close without merging. :)

wolf99:
Whoops, I misunderstood your meaning (blame it on a late Sunday evening!) You meant that the _original_ links to the second page where _should_ link to the _first_ page. You are correct there then 😄

PR #131, opened 2017-03-17
react: Update only dependents of the just-set cell

petertseng:
The initial version of react (see #126) let the reactor store all cells
in a linked list, then blindly updated all cells whenever any cell was
set. This was correct but wasteful.

This commit rearchitects the example solution so that:

* reactors only keep track of the input cells
* all other cells keep track of their dependent compute cells
* only dependents get updated when a cell is set

Space cost of:

* one extra `struct child` allocated per compute2 cell created.
* one extra integer field per cell.

### DIFF
--- a/exercises/react/src/example.c
+++ b/exercises/react/src/example.c
@@ -79,18 +79,38 @@ void destroy_reactor(struct reactor *r)
    free(r);
 }
 
-static void add_child(struct child **list, struct cell *cell)
+#define SUCCESS 1
+#define FAIL 0
+
+static int add_child(struct child **list, struct cell *cell)
 {
    struct child *child = calloc(1, sizeof(struct child));
+   if (!child) {
+      return FAIL;
+   }
    child->cell = cell;
    child->next = *list;
    *list = child;
+   return SUCCESS;
+}
+
+static void remove_child(struct child **list)
+{
+   struct child *to_remove = *list;
+   *list = to_remove->next;
+   free(to_remove);
 }
 
 struct cell *create_input_cell(struct reactor *r, int initial_value)
 {
    struct cell *c = calloc(1, sizeof(struct cell));
-   add_child(&r->input, c);
+   if (!c) {
+      return NULL;
+   }
+   if (add_child(&r->input, c) != SUCCESS) {
+      free(c);
+      return NULL;
+   }
    c->kind = kind_input;
    c->value = initial_value;
    return c;
@@ -101,7 +121,13 @@ struct cell *create_compute1_cell(struct reactor *r, struct cell *input,
 {
    (void)r;
    struct cell *c = calloc(1, sizeof(struct cell));
-   add_child(&input->child, c);
+   if (!c) {
+      return NULL;
+   }
+   if (add_child(&input->child, c) != SUCCESS) {
+      free(c);
+      return NULL;
+   }
    c->kind = kind_compute1;
    c->input1 = input;
    c->compute1 = compute;
@@ -115,8 +141,18 @@ struct cell *create_compute2_cell(struct reactor *r, struct cell *input1,
 {
    (void)r;
    struct cell *c = calloc(1, sizeof(struct cell));
-   add_child(&input1->child, c);
-   add_child(&input2->child, c);
+   if (!c) {
+      return NULL;
+   }
+   if (add_child(&input1->child, c) != SUCCESS) {
+      free(c);
+      return NULL;
+   }
+   if (add_child(&input2->child, c) != SUCCESS) {
+      remove_child(&input1->child);
+      free(c);
+      return NULL;
+   }
    c->kind = kind_compute2;
    c->input1 = input1;
    c->input2 = input2;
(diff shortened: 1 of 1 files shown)


exercises/react/src/example.c
```diff
@@ -41,42 +43,54 @@ struct reactor *create_reactor()
...
+      struct cb *next_cb = cb->next;
+      free(cb);
+      cb = next_cb;
+   }
 
+   struct child *child = c->child;
+   while (child) {
       struct child *next = child->next;
+      if (c == child->cell->input1) {
+         // Don't double-free for a compute2 cell.
```
petertseng: That was my main concern with making this. I thought it would be very tricky to free every cell exactly once, but every compute cell has exactly one `input1`, so it works out.

exercises/react/src/example.c
```diff
@@ -85,30 +99,35 @@ struct cell *create_input_cell(struct reactor *r, int initial_value)
...
    c->kind = kind_compute2;
    c->input1 = input1;
    c->input2 = input2;
    c->compute2 = compute;
    c->value = compute(get_cell_value(input1), get_cell_value(input2));
+   c->last_cb_value = c->value;
    return c;
 }
 
+#define each_child(c) struct child *child = (c)->child; child; child = child->next
```
petertseng: too obscure? I didn't like repeating myself.

exercises/react/src/example.c
```diff
@@ -41,42 +43,54 @@ struct reactor *create_reactor()
...
-   while (child) {
-      struct cb *cb = child->cell->cb;
-      while (cb) {
-         struct cb *next_cb = cb->next;
-         free(cb);
-         cb = next_cb;
-      }
-      free(child->cell);
+   struct cb *cb = c->cb;
+   while (cb) {
```
petertseng: my submission on the website asks whether it is possible to reduce the duplication between the three while loops

exercises/react/src/example.c
ryanplusplus: I don't really understand the need for this `if`. Can you explain this for me?

exercises/react/src/example.c
ryanplusplus: I like it

ryanplusplus:
Looks good to me outside of a bit of confusion in `destroy_cell`. I thought that this would be much more complicated than it is, particularly in making sure that callbacks weren't fired too many times, but this is really quite nice.

exercises/react/src/example.c
petertseng: let's say it's not there, so that we always call `destroy_cell(child->cell)`.

In that case, even this:

```c
   struct cell *input1 = create_input_cell(r, 1);
   struct cell *input2 = create_input_cell(r, 2);
   struct cell *output = create_compute2_cell(r, input1, input2, plus);
```

will double-free:

```
destroy_reactor calls destroy_cell(input1)
destroy_cell(input1) calls destroy_cell(output)
destroy_reactor calls destroy_cell(input2)
destroy_cell(input2) calls destroy_cell(output)
```

Everything else is correct; it just needed to avoid a second call to `destroy_cell(output)`

ryanplusplus: Ahh, that makes much more sense. Thanks!

patricksjackson:
I like that you're updating this :+1:, but if you're going to do that I'd suggest also at least adding NULL checks...

At some point I may bring back up the problem, but after I actually write a solution. (I've spent the last 20 minutes just attempting to get it to compile an empty skeleton. The API has 9 functions and 2 structs that need to be defined, and all of their inputs need to be used (to avoid the compiler warnings as errors), before any of the tests can even compile.)

petertseng:
> adding NULL checks

Careful - I wouldn't want the maintainers of this track to get in a back-and-forth on https://github.com/exercism/xc/pull/126#discussion_r105526960 and this, because it's not my place to mediate such a one. But I'm here to let you know that whichever you choose (NULL checks or no NULL checks), it can happen. There is now a commit dedicated to adding the NULL checks, and if y'all decide not to have NULL checks that commit can simply be reverted.

> last 20 minutes just attempting to get it to compile an empty skeleton
> The API has 9 functions and 2 structs that need to be defined, and all of their inputs need to be used (to avoid the compiler warnings as errors)

I admit that when I was working on this to submit to the site, I removed `-Werror` from my makefile as that was too annoying. It is possible the track can provide a stub file with `(void)` on all inputs to make that easier. It would be an exception to #43. Any other exercises with many functions that need to be defined, that could also benefit from that? Ones that come to mind immediately are custom-set and circular-buffer, neither of which this track has.

petertseng:
Hi! As I explained, it's not necessary to merge this--the only reason to do this is if it's really important that the example solution not do unnecessary work. Given this fact, and the fact that I need to keep my [open PRs I've created](https://github.com/pulls) clean, I need to close this for the sole purpose of getting it out of my queue. It doesn't mean I've given up on the idea! Y'all are free to reopen it for the purpose of merging it of course.

ryanplusplus:
Hey @petertseng I'm sorry this fell off my radar. I think we'd like this, thanks for your help!

PR #132, opened 2017-03-26
Adding Valgrind check to Travis CI

### LINKED ISSUE #127: Should CI run and check valgrind?

patricksjackson:
Opening this to test against Travis. I expect that this will need to patch a lot of the existing solutions to get it to pass.

Closes #127

**Edit:** Apparently only 5 problems needed patching, and all of them were in the example solutions, so these patches don't change any of the user interface.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: bash
 
 before_install:
 - sudo apt-get -qq update
-- sudo apt-get install -y indent
+- sudo apt-get install -y indent valgrind
 
 script:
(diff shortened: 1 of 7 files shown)


wolf99:
Fair play - Totally forgot I'd opened #127! 😳

ryanplusplus:
Awesome! Thanks for doing this.

PR #135, opened 2017-03-26
Adding the new pascals triangle

DeathSec:
Every thing is OK

### DIFF
--- a/exercises/pascals-triangle/src/example.c
+++ b/exercises/pascals-triangle/src/example.c
@@ -15,7 +15,6 @@ size_t **pascals_triangle(char *count)
    }
    size_t **zero = (size_t **) calloc(1, sizeof(size_t *));
    zero[0] = (size_t *) calloc(1, sizeof(size_t));
-   int i, j;
    int rows = atoi(count);
    if (!strcmp(count, "0")) {
       return zero;
@@ -26,13 +25,13 @@ size_t **pascals_triangle(char *count)
       pt_free(zero, 1);
    }
    size_t **result = (size_t **) calloc(rows, sizeof(size_t *));
-   for (i = 0; i < rows; i++) {
+   for (int i = 0; i < rows; i++) {
       result[i] = (size_t *) calloc(rows, sizeof(size_t));
    }
    result[0][0] = 1;
-   for (i = 1; i < rows; ++i) {
+   for (int i = 1; i < rows; ++i) {
       result[i][0] = 1;
-      for (j = 1; j <= i; ++j) {
+      for (int j = 1; j <= i; ++j) {
          result[i][j] = result[i - 1][j] + result[i - 1][j - 1];
       }
    }
--- a/exercises/pascals-triangle/src/example.c
+++ b/exercises/pascals-triangle/src/example.c
@@ -1,0 +1,40 @@
+#include "pascals_triangle.h"
+
+void pt_free(size_t ** mem_block, size_t n)
+{
+   for (size_t i = 0; i < n; i++) {
+      free(mem_block[i]);
+   }
+   free(mem_block);
+}
+
+size_t **pascals_triangle(char *count)
+{
+   if (count == NULL) {
+      return (size_t **) NULL;
+   }
+   size_t **zero = (size_t **) calloc(1, sizeof(size_t *));
+   zero[0] = (size_t *) calloc(1, sizeof(size_t));
+   int i, j;
+   int rows = atoi(count);
+   if (!strcmp(count, "0")) {
+      return zero;
+   } else if (rows <= 0) {
+      pt_free(zero, 1);
+      return (size_t **) NULL;
+   } else {
+      pt_free(zero, 1);
+   }
+   size_t **result = (size_t **) calloc(rows, sizeof(size_t *));
+   for (i = 0; i < rows; i++) {
+      result[i] = (size_t *) calloc(rows, sizeof(size_t));
+   }
+   result[0][0] = 1;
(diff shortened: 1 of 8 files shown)


exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,40 @@
...
+}
+
+size_t **pascals_triangle(char *count)
+{
+   if (count == NULL) {
+      return (size_t **) NULL;
+   }
+   size_t **zero = (size_t **) calloc(1, sizeof(size_t *));
+   zero[0] = (size_t *) calloc(1, sizeof(size_t));
+   int i, j;
```
patricksjackson: These indices should be `size_t`, all the others are. Also, the rest of the codebase typically uses the style of defining them within the loop declaration (I think).

patricksjackson:
This is looking good :+1: I want somebody else to review this too before we merge it though. @ryanplusplus @palisaide @wolf99 ?

Also, this may be a stupid question, but why is the input a string?

exercises/pascals-triangle/src/example.c
DeathSec: rows is int , so i thought comparison between int and size_t will not be good , the reason i did not declare i , j in the loop is that , According to C89 standards its illegal , but exercism/xc is using C99 standards , so i will commit the changes soon

DeathSec: and so comparing int and size_t was a bad idea , using int is also good sometimes , because for size_t it has to set widths so it drops the performance... , but that does not matter here.

`src/pascals_triangle.c: In function 'pascals_triangle':
src/pascals_triangle.c:28:25: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
    for (size_t i = 0; i < rows; i++) {
                         ^
src/pascals_triangle.c:32:25: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
    for (size_t i = 1; i < rows; ++i) {
`

DeathSec:
The input is a string because we can handle NULL and "" empty input , inputs quietly as described in x-common , test data!.. , also if its a integer input , if i input NULL it thinks its 0 and returns 0 rows instead of a error as required in x-common test data...

DeathSec:
and also , as mentioned in [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/pascals-triangle/canonical-data.json) , it requires the function to return -1 as a sort of error , is returning NULL a good idea ? , or should i return -1 instead ?

ryanplusplus:
@DeathSec I think that having the input be a string is a bit odd and will be off-putting for someone doing this exercise. The x-common tests call out a null input, but that doesn't really work in C since integers (the natural type) are not nullable. So I think we should just omit that test here.

exercises/pascals-triangle/src/example.h
```diff
@@ -0,0 +1,9 @@
+#ifndef PASCALS_TRIANGLE_H
+#define PASCALS_TRIANGLE_H
+#include <stdlib.h>
+#include <string.h>
+
+void pt_free(size_t **, size_t);
```
ryanplusplus: I might just be dense, but it took me a little bit to figure out what `pt_free` meant. How about making this match the other function? Perhaps `create_triangle`/`generate_triangle` and `free_triangle`?

exercises/pascals-triangle/test/test_pascals_triangle.c
```diff
@@ -0,0 +1,93 @@
...
+      {1, 3, 3, 1}
+   };
+   size_t **r = pascals_triangle("4");
+   TEST_ASSERT_TRUE(check(4, expected, r));
+   pt_free(r, 4);
+}
+
+void test_negative_rows(void)
+{
+   TEST_ASSERT_TRUE((pascals_triangle("-1") == (size_t **) NULL));
```
ryanplusplus: Is the cast to `(size_t **)` necessary?

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,39 @@
...
+   }
+   free(mem_block);
+}
+
+size_t **pascals_triangle(char *count)
+{
+   if (count == NULL) {
+      return (size_t **) NULL;
+   }
+   size_t **zero = (size_t **) calloc(1, sizeof(size_t *));
```
ryanplusplus: Why do you allocate `zero` here if you only use it in the first `if` below? This forces you to deallocate when you don't use it which is actually in two different cases.

exercises/pascals-triangle/src/example.c
```diff
@@ -0,0 +1,39 @@
...
+   int rows = atoi(count);
+   if (!strcmp(count, "0")) {
+      return zero;
+   } else if (rows <= 0) {
+      pt_free(zero, 1);
+      return (size_t **) NULL;
+   } else {
+      pt_free(zero, 1);
+   }
+   size_t **result = (size_t **) calloc(rows, sizeof(size_t *));
```
ryanplusplus: It is not recommended to cast the result of `malloc`/`calloc`: http://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

ryanplusplus:
Sorry for the delay in reviewing this.

Overall the solution looks good. Just a few remaining nitpicks then I'm happy to merge. Thanks for your extra effort in improving the solution.

DeathSec:
sure do.

wolf99:
LGTM, good work 👍

ryanplusplus:
LGTM, thanks for your contribution!

PR #137, opened 2017-03-27
Phone-number: Add missing tests and adjust example to suit changes

### LINKED ISSUE #124: phone-number tests dont check for length greater than 11
The problem brief mentions:

> - If the phone number is more than 11 digits assume that it is a bad number

But the tests do not check for this requirement.

wolf99:
Fixes #124


wolf99:
I note that the x-common test spec expects `null` to be returned for invalid cases. Currently the example returns, and the tests expect, `"000000000"`. Is this acceptable or should it be changed?

patricksjackson:
This looks fine to me. :+1: 

The x-common tests are up for interpretation per each language. Returning all 0's is probably fine, though it may be a little weird. I'd be supportive of a change to using null though.

PR #138, opened 2017-03-30
Adding Perfect Number exercise

DeathSec:
Every test is from [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/perfect-numbers/canonical-data.json) , and as mentioned in canonical-data.json ,  0 is not a natural number , is that correct , because as i recall , Natural Numbers are set of positive and non-negative numbers ? .. which we can say 0 is a natural number? ..

### DIFF
--- a/exercises/perfect-numbers/src/example.c
+++ b/exercises/perfect-numbers/src/example.c
@@ -13,16 +13,16 @@ static int aliquot_sum(int n)
 
 kind classify_number(int n)
 {
-   if (n <= 0) {
-      return error;
+   kind class = error;
+   if (n > 0) {
+      int buf = aliquot_sum(n);
+      if (buf > n) {
+         class = abundant_number;
+      } else if (buf < n) {
+         class = deficient_number;
+      } else {
+         class = perfect_number;
+      }
    }
-   int buf = aliquot_sum(n);
-   if (buf > n) {
-      return abundant_number;
-   } else if (buf < n) {
-      return deficient_number;
-   } else {
-      return perfect_number;
-   }
-   return error;
+   return class;
 }
--- a/exercises/perfect-numbers/src/example.c
+++ b/exercises/perfect-numbers/src/example.c
@@ -17,12 +17,12 @@ kind classify_number(int n)
       return error;
    }
    int buf = aliquot_sum(n);
-   if (buf == n) {
-      return perfect_number;
-   } else if (buf > n) {
+   if (buf > n) {
       return abundant_number;
    } else if (buf < n) {
       return deficient_number;
+   } else {
+      return perfect_number;
    }
    return error;
 }
--- a/exercises/perfect-numbers/src/example.c
+++ b/exercises/perfect-numbers/src/example.c
@@ -1,0 +1,28 @@
+#include "perfect_numbers.h"
+
+static int aliquot_sum(int n)
+{
+   int result = 0;
+   for (int i = 1; i < n; ++i) {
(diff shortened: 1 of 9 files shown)


exercises/perfect-numbers/test/test_perfect_numbers.c
```diff
@@ -0,0 +1,87 @@
+#include "vendor/unity.h"
+#include "../src/perfect_numbers.h"
+
+void test_smallest_perfect_number_is_classified_correctly(void)
+{
+   TEST_ASSERT_TRUE(classify_number(6) == perfect_number);
```
wolf99: Small detail only, but I would make these of the form `TEST_ASSERT_EQUAL(perfect_number, classify_number(6));`

exercises/perfect-numbers/src/example.c
```diff
@@ -0,0 +1,28 @@
...
+   return result;
+}
+
+kind classify_number(int n)
+{
+   if (n <= 0) {
+      return error;
+   }
+   int buf = aliquot_sum(n);
+   if (buf == n) {
```
wolf99: Again, only a very small detail; I would use the rule that: an `if` should end with an `else` if practical. So this would become: 
```C
if (buf > n) { 
    return abundant_number; 
} else if (buf < n) {
    return deficient_number;
} else {
    return perfect_number;
}
```
Though I guess this would be somewhat pedantic coding! 😆  (It comes from the MISRA rule regarding `if ... else if` usage IIRC)

DeathSec: oh! :smile: :+1:

wolf99:
As regards the naturalness of 0, the matter is debatable. However the definition of _perfect_ numbers more usually refers to whole positive integers, which is a set that does **not** include zero.

exercises/perfect-numbers/test/test_perfect_numbers.c
DeathSec: :taco: :+1:

wolf99:
LGTM 👍 

@deathsec If you're interested in MISRA (i.e. it would not be necessary at all for exercism code) for the safer coding aspect take a look at [this gist](https://gist.github.com/wolf99/42c612bbd00b85b44c468a5b12165689) which shows the pattern I would generally follow  in cases like this to follow the rule for only having one exit point (`return`) per function. Like I mentioned for the `else`, unless required I usually only follow MISRA rules where practical (e.g. it doesn't obfuscate the code too much or impede more efficient code).

DeathSec:
@wolf99 you used to work at MISRA ? , Until now i have no idea about MISRA-C , nice to work with experienced people.. :smile:

exercises/perfect-numbers/src/example.c
```diff
@@ -0,0 +1,28 @@
...
+      }
+   }
+   return result;
+}
+
+kind classify_number(int n)
+{
+   kind class = error;
+   if (n > 0) {
+      int buf = aliquot_sum(n);
```
ryanplusplus: Instead of `buf`, how about `sum`?

DeathSec: sure do

ryanplusplus:
Looks great, thanks! I'll merge as soon as Travis approves.

wolf99:
@DeathSec I work with embedded systems that sometimes require some level of MISRA standard conformance. I'm still learning myself though! FYI, for a first year high school student you are pretty good at C, well done and well done on contributing to open source!

PR #139, opened 2017-04-01
Adding Wordy Exercise!

DeathSec:
**EDIT**: Clarifying the Code  
Get a string as input -> put const char* instructions = "What is x plus y plus z?" , 
check if last character is '?' if not return result == 0
let input be a block of memory of size instructions_len
copy instructions to input avoiding the last character '?'
let l be a array of strings split at delimiter 'space' .
i.e l = [ "What" -> 0   
              ,"is"  -> 1
              ,"x"  -> 2
              ,"plus"  -> 3
              ,"y" -> 4
              ,"plus" -> 5  
              ,"z" -> 6
]
By the above we can say that the 2nd item must be a integer.
and 3rd item must be a operator , if not it will be a syntax error.
, 4th item must be a integer , if not it will be a syntax error.
So when item no. == 2 in the loop, check if its a number then let result = x;
now the second if is only valid if the third item is a number and item no. != 2;
but the third item is or will be a operator , so set the operator = ADDITION.
now the item no. == 4 and the second if becomes active and checks the operator
then do (result = result + item no. 4) , then set operator = -1;
now the item no. == 5 and the third if is active because its not a no. set operator = ADD
now the item no. == 6 and its a number and the item no. != 2 , so do.
result = result + item no. 6.
and the answer returns as x + y + z = result!

strdup and strsep is not implemented in C99 standards so , i made my own with reference to the official libc.

raise_to function is implemented to avoid typecasting of double to int and as the exercise does not need complex math , we can use simple function to gain performance

**Bonus**: As given in the wordy description from x-common repo , this example can handle exponentials.

**resource**: [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/wordy/canonical-data.json) , [Description](https://github.com/exercism/x-common/blob/master/exercises/wordy/description.md)

if have a better solution please , notify me!

**EDIT**: is_integer function solves the problem in atoi , (i.e lets say "What is 0 plus 5?")
, atoi cannot report error if its a string it returns error 0 and if its 0 it returns 0 , so i implemented is_integer , it returns the same input if there is no words but numbers and for error reporting it returns a NULL pointer.

### DIFF
--- a/exercises/wordy/src/example.c
+++ b/exercises/wordy/src/example.c
@@ -1,0 +1,174 @@
+#include "wordy.h"
+
+#define ADDITION 1
+#define SUBRACTION 2
+#define DIVISION 3
+#define MULTIPLICATION 4
+#define EXPO 5
+
+static char *strdup(const char *str)
+{
+   char *copy;
+   if ((copy = calloc(strlen(str) + 1, sizeof(char *))) != NULL) {
+      strcpy(copy, str);
+   }
+   return copy;
+}
+
+static char *strsep(char **stringp, const char *delim)
+{
+   char *s;
+   const char *spanp;
+   int c, sc;
+   char *tok;
+   if ((s = *stringp) == NULL)
+      return (NULL);
+   for (tok = s;;) {
+      c = *s++;
+      spanp = delim;
+      do {
+         if ((sc = *spanp++) == c) {
+            if (c == 0)
+               s = NULL;
+            else
+               s[-1] = 0;
+            *stringp = s;
+            return (tok);
+         }
+      } while (sc != 0);
+   }
+}
+
+static int raise_to(int base, int exp)
+{
+   int ret = 1;
+   for (int i = 0; i < exp; i++) {
+      ret = ret * base;
+   }
+   return ret;
+}
+
+static void free_list(char **list, size_t n)
+{
+   for (size_t i = 0; i < n; ++i) {
+      free(list[i]);
+   }
+   free(list);
+}
+
+static char *is_integer(char *s)
+{
+   size_t s_len = strlen(s);
+   size_t i = 0;
+   if (strstr(s, "-")) {
+      i = 1;
+   }
+   for (; i < s_len; i++) {
+      if (!isdigit(s[i])) {
+         return NULL;
+      }
+   }
+   return s;
+}
+
+static char **create_list(const char *input, size_t * n)
+{
+   char *s = strdup(input);
+   size_t list_alloc = 1;
+   size_t list_used = 0;
+   char **lists = calloc(list_alloc, sizeof(char *));
+   char *list, *rest = s;
+   while ((list = strsep(&rest, " ")) != NULL) {
+      if (list_used == list_alloc) {
+         list_alloc *= 2;
+         lists = realloc(lists, list_alloc * sizeof(char *));
+      }
+      lists[list_used++] = strdup(list);
+   }
+   if (list_used == 0) {
+      free(lists);
+      lists = NULL;
+   } else {
+      lists = realloc(lists, list_used * sizeof(char *));
+   }
+   *n = list_used;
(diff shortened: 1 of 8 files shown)


exercises/wordy/src/example.c
```diff
@@ -0,0 +1,174 @@
...
+}
+
+static char **create_list(const char *input, size_t * n)
+{
+   char *s = strdup(input);
+   size_t list_alloc = 1;
+   size_t list_used = 0;
+   char **lists = calloc(list_alloc, sizeof(char *));
+   char *list, *rest = s;
+   while ((list = strsep(&rest, " ")) != NULL) {
```
DeathSec: using delimiter as space , by default!

exercises/wordy/src/example.c
```diff
@@ -0,0 +1,174 @@
...
+   int no_op = 1;
+   char *buf1 = NULL;
+   char *buf2 = NULL;
+   size_t n = 0;
+   size_t instructions_len = strlen(instructions);
+   if (instructions[instructions_len - 1] != '?') {
+      return result;
+   }
+   char *input = calloc(instructions_len, sizeof(char *));
+   strncpy(input, instructions, instructions_len - 1);
```
DeathSec: After checking the '?' at the end , copying to input , without the '?' .

exercises/wordy/src/example.c
```diff
@@ -0,0 +1,174 @@
...
+         no_op = 0;
+      }
+      if (!strcmp(l[i], "raised") && !strcmp(l[i + 1], "to")
+          && !strcmp(l[i + 2], "the")) {
+         operator = EXPO;
+         no_op = 0;
+      }
+   }
+   free_list(l, n);
+   if (no_op) {
```
DeathSec: if no operation is done return NULL , as the test data requires us to do , "What is x?" -> all syntax is right , so result will be x , if we don't check for operations so if no operations are done then return 0 aka NULL

wolf99: NULL is *normally* only used with pointers. I think to be correct for C semantics here the tests and return value should both use `0` rather than `NULL`

exercises/wordy/src/example.c
```diff
@@ -0,0 +1,174 @@
...
+      free_list(l, n);
+      return result;
+   }
+   for (size_t i = 2; i < n; ++i) {
+      if ((buf1 = is_integer(l[i])) != NULL && i == 2) {
+         result = atoi(buf1);
+      }
+      if (((buf1 = is_integer(l[i])) != NULL && i != 2)
+          || (strstr(l[i], "th") && i != 2 && isdigit(l[i][0]))) {
+         if (operator == ADDITION) {
```
wolf99: Operato hasn't been set yet so will always be -1 at this point won't it?

DeathSec: i will look into it

DeathSec: yes , the parent loop never enters the code block until first digit is detected in the first item in the list because of the syntax , Always "What is x plus y?" or "What is x plus y multiplied by z divided by 1?" , so when the string is divided into list , "What" , "is" , x , "plus" , y , "......" . by this x must be the second item in the list , if not it will be a syntax error. and the third item must be a operator so the operator will be set right after x is detected if not it results in syntax error , all the test ran successfully

exercises/wordy/src/example.c
DeathSec: i am returning zero but test it as null in the test suit , i think i must clean the code a little..

exercises/wordy/src/example.c
wolf99: :+1: my bad!

DeathSec: i think my code is too cryptic, i will make simpler soon

DeathSec:
I am closing this pull request for now!

PR #141, opened 2017-04-23
perfect_number: Optimize example solution

patricksjackson:
When running all the tests using bin/run-tests there is a notable pause while running perfect_numbers, so I made a small optimization to speed up the example solution.

This is totally unnecessary, but the new version is a better example solution for anyone that ends up looking at it.


ryanplusplus:
This looks great. Thanks!

PR #142, opened 2017-05-13
Add scrabble-score exercise

subpop:
I implemented all the tests from canonical-data.json. I wasn't sure on difficulty, nor does it look like config.json is sorted by difficulty, so I used the rust implementation as reference.

### DIFF
--- a/exercises/scrabble-score/src/example.c
+++ b/exercises/scrabble-score/src/example.c
@@ -1,0 +1,43 @@
+#include "scrabble_score.h"
+#include <ctype.h>
+
+static unsigned int score_book[] = {
+   1,
+   3,
+   3,
+   2,
+   1,
+   4,
+   2,
+   4,
+   1,
+   8,
+   5,
+   1,
+   3,
+   1,
+   1,
+   3,
+   10,
+   1,
+   1,
+   1,
+   1,
+   4,
+   4,
+   8,
+   4,
+   10
+};
+
+unsigned int score(const char *word)
+{
+   char c;
+   unsigned int word_score = 0;
+
+   while ((c = tolower(*word++)) != '\0') {
+      word_score += score_book[c - 'a'];
+   }
+
+   return word_score;
+}
--- a/exercises/scrabble-score/src/example.h
+++ b/exercises/scrabble-score/src/example.h
@@ -1,0 +1,6 @@
+#ifndef SCRABBLE_SCORE_H
+#define SCRABBLE_SCORE_H
+
+unsigned int score(const char *word);
+
+#endif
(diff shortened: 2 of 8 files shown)


exercises/scrabble-score/src/example.h
```diff
@@ -0,0 +1 @@
+unsigned int score(const char *word);
```
ryanplusplus: Can you add include guards to this header?

exercises/scrabble-score/src/example.c
```diff
@@ -0,0 +1,63 @@
...
+   struct letter_score l;
+
+   for (int i = 0; i < 27; i++) {
+      l = score_book[i];
+      if (l.letter == c) {
+         return &score_book[i];
+      }
+   }
+   return 0;
+}
```
ryanplusplus: Instead of doing a linear search, you could index by the ASCII value of the character. Simply subtract `'a'` and you will get the index of the score.

ryanplusplus:
Looks great so far. A few small changes and I'll give it a thumbs up.

exercises/scrabble-score/src/example.c
subpop: Good idea. Changed in c8dfc8b

exercises/scrabble-score/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+};
+
+unsigned int score(const char *word)
+{
+   char c;
+   struct l;
+   unsigned int word_score = 0;
+
+   while ((c = tolower(*word++)) != '\0') {
+      struct letter_score l = score_book[c - 'a'];
```
ryanplusplus: This is better, but now the `char` in `struct letter_score` is unused and unnecessary.

subpop: True. The whole struct becomes unnecessary.

ryanplusplus:
Looks great. I'm going to give other reviewers a chance to take a look but will merge if nobody bites.

wolf99:
LGTM.

Though I might possibly have done:

```
#define A_LETTER_SCORE 1
#define B_LETTER_SCORE 1
// and so on ...

static unsigned int score_book[] = {
   A_LETTER_SCORE, B_LETTER_SCORE, // and so on...
```
Just to display a bit more meaning. Probably be overkill though!

ryanplusplus:
@wolf99 Yeah, I think a bit overkill on the `#define`s. If anything, I'd just add `// a`, `// b` after each corresponding value.

As far as the `NULL` checks, I agree in general, but since this is an easy exercise I think it's probably fine to just keep it simple (especially since this isn't covered in the canonical tests).

PR #143, opened 2017-05-28
Remove license info from README

kytrinyx:
We don't need the duplication, especially now that the GitHub interface shows the
license information on the main page of the repository when it can be detected
directly from the LICENSE file.


patricksjackson:
Thanks.

PR #150, opened 2017-06-27
Add acronym exercise

wolf99:
Added example.c/.h and test source, added item to config JSON.
Can't recall if I need to do anything further to add an exercise?

### DIFF
--- a/config.json
+++ b/config.json
@@ -65,13 +65,13 @@
       ]
     },
     {
-        "difficulty": 1,
-        "slug": "acronym",
-        "topics": [
-            "strings",
-            "memory management",
-            "control-flow (loops)"
-        ]
+      "difficulty": 1,
+      "slug": "acronym",
+      "topics": [
+        "strings",
+        "memory management",
+        "control-flow (loops)"
+      ]
     },
     {
       "difficulty": 1,
--- a/config.json
+++ b/config.json
@@ -65,6 +65,15 @@
       ]
     },
     {
+        "difficulty": 1,
+        "slug": "acronym",
+        "topics": [
+            "strings",
+            "memory management",
+            "control-flow (loops)"
+        ]
+    },
+    {
       "difficulty": 1,
       "slug": "grains",
       "topics": [
--- a/exercises/acronym/src/example.c
+++ b/exercises/acronym/src/example.c
@@ -42,4 +42,3 @@ char *abbreviate(const char phrase[])
    }
    return acronym;
 }
-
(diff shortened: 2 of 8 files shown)


config.json
```diff
@@ -65,6 +65,15 @@
       ]
     },
     {
+        "difficulty": 1,
+        "slug": "acronym",
+        "topics": [
+            "strings",
+            "memory management",
+            "control-flow (loops)"
+        ]
```
ryanplusplus: Formatting is off here, 2 spaces are used for indents in this file.

exercises/acronym/src/example.c
```diff
@@ -0,0 +1,50 @@
+#include <ctype.h>
+#include <stdlib.h>
+
+static int count_words(const char phrase[]);
+static int is_first_letter_of_word(int c, int b);
```
ryanplusplus: You can pretty easily reorganize your functions to avoid static prototypes. This makes your life easier because it reduces the duplication (and thus maintenance cost), but also introduces an ordering that readers can use when examining your source code (code that is lower in the file is generally at a higher level).

exercises/acronym/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+      if (is_first_letter_of_word(phrase[i], phrase[i - 1]))
+         ++count;               /* subsequent words */
+   return count;
+}
+
+static int is_first_letter_of_word(int character, int preceding_character)
+{
+   /* Check if the current character is a letter and 
+    * the preceding character is a space or hyphen
+    */
```
ryanplusplus: This comment seems unnecessary as the code tells me exactly this.

ryanplusplus:
Just a few nitpicks

exercises/acronym/src/example.c
wolf99: Ah, I normally work the other way, so that the file starts with an index of sorts of the static functions then the main functionality front and center and the static helper cruft comes after. Will adjust as you suggest though

wolf99:
thanks @ryanplusplus, all fixed up, should be good to go.

exercises/acronym/src/example.c
ryanplusplus: Yeah, I used to work like that for the same reason, but grew really tired of updating static prototypes. Additionally, I realized that a decent IDE will give you an index of all of your static functions without any tedious bookkeeping by the developer.

ryanplusplus:
Changes look great. I'm going to give @patricksjackson a chance to look but will otherwise merge tomorrow.

PR #152, opened 2017-06-27
Add isogram exercise

wolf99:
Added tests, example.c/.h and added info to config.json

### DIFF
--- a/config.json
+++ b/config.json
@@ -23,7 +23,7 @@
 		"control-flow (if-else statements)",
 		"control-flow (loops)"
 	  ]
-	}
+	},
     {
       "difficulty": 5,
       "slug": "anagram",
--- a/config.json
+++ b/config.json
@@ -14,16 +14,16 @@
         "text formatting"
       ]
     },
-	{
-	  "slug": "isogram",
-	  "difficulty": 1,
-	  "topics": [
-	    "arrays",
-	    "strings",
-		"control-flow (if-else statements)",
-		"control-flow (loops)"
-	  ]
-	},
+    {
+      "slug": "isogram",
+      "difficulty": 1,
+      "topics": [
+        "arrays",
+        "strings",
+        "control-flow (if-else statements)",
+        "control-flow (loops)"
+      ]
+    },
     {
       "difficulty": 5,
       "slug": "anagram",
--- a/config.json
+++ b/config.json
@@ -14,6 +14,16 @@
         "text formatting"
       ]
     },
+	{
+	  "slug": "isogram",
+	  "difficulty": 1,
+	  "topics": [
+	    "arrays",
+	    "strings",
+		"control-flow (if-else statements)",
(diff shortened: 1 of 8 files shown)


exercises/isogram/src/example.c
```diff
@@ -0,0 +1,25 @@
+#include "isogram.h"
+#include <ctype.h>
+#include <stddef.h>
+
+bool isIsogram(const char phrase[])
+{
+   long usedLetters = 0;
```
ryanplusplus: This only needs to be 32-bit, how about using `uint32_t` to both guarantee that it will always be big enough and also to make sure we aren't using more memory than necessary?

exercises/isogram/src/example.c
```diff
@@ -0,0 +1,25 @@
...
+
+   if (phrase == NULL)
+      return false;
+
+   for (int i = 0; phrase[i] != '\0'; ++i) {
+      if (isalpha(phrase[i])) {
+         int bitNumber = toupper(phrase[i]) - 'A';
+         long bitMask = 1 << bitNumber;
```
ryanplusplus: Same here, can we switch to `uint32_t`?

exercises/isogram/src/example.c
```diff
@@ -0,0 +1,25 @@
...
+bool isIsogram(const char phrase[])
+{
+   long usedLetters = 0;
+
+   if (phrase == NULL)
+      return false;
+
+   for (int i = 0; phrase[i] != '\0'; ++i) {
+      if (isalpha(phrase[i])) {
+         int bitNumber = toupper(phrase[i]) - 'A';
```
ryanplusplus: This doesn't need to be signed, right?

config.json
```diff
@@ -14,6 +14,16 @@
...
       ]
     },
+	{
+	  "slug": "isogram",
+	  "difficulty": 1,
+	  "topics": [
+	    "arrays",
+	    "strings",
+		"control-flow (if-else statements)",
+		"control-flow (loops)"
```
ryanplusplus: Looks like you've mixed tabs and spaces. Be sure to always use spaces in order to maximize your income (https://stackoverflow.blog/2017/06/15/developers-use-spaces-make-money-use-tabs/).

ryanplusplus:
Looks good outside of a couple of nitpicks.

config.json
wolf99: Ha ha, that article is getting huge coverage! Normally I use spaces but did this on an old machine I had and was fighting Notepad++ on the tabs/spaces front 😕

wolf99:
Thanks for the review @ryanplusplus agreed and changed on all points. Should be good to go know I think.

ryanplusplus:
Everything looks great. I'm going to give @patricksjackson a chance to look, but otherwise I will merge tomorrow.

PR #153, opened 2017-07-03
Add all-your-base

wolf99:
Add all-your-base exercise.
I've tried to make an implementation that didn't require dynamic memory management; this has made the function signature and thus the test file slightly more complicated than otherwise, though still easily understandable I think.

### DIFF
--- a/exercises/all-your-base/src/all_your_base.h
+++ b/exercises/all-your-base/src/all_your_base.h
@@ -1,0 +1,12 @@
+#ifndef ALL_YOUR_BASE_H
+#define ALL_YOUR_BASE_H
+
+#include <stddef.h>
+#include <stdint.h>
+
+#define DIGITS_ARRAY_SIZE 64
+
+size_t rebase(int16_t digits[DIGITS_ARRAY_SIZE], int16_t fromBase,
+              int16_t toBase, size_t numDigits);
+
+#endif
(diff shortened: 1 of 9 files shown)


exercises/all-your-base/src/example.c
```diff
@@ -0,0 +1,38 @@
+#include "all_your_base.h"
+#include <math.h>
+
+size_t rebase(int16_t digits[DIGITS_ARRAY_SIZE], int16_t fromBase,
+              int16_t toBase, size_t numDigits)
+{
+   uint16_t denary = 0;
+   size_t newNumDigits = 0;
+   //bool leadingZeros = true;
```
wolf99: Flip, I forgot to remove this!

exercises/all-your-base/src/all_your_base.h
```diff
@@ -0,0 +1,12 @@
+#ifndef ALL_YOUR_BASE_H
+#define ALL_YOUR_BASE_H
+
+#include <stddef.h>
+#include <stdint.h>
+
+#define DIGITS_ARRAY_SIZE 64
+
+size_t rebase(int16_t digits[DIGITS_ARRAY_SIZE], int16_t fromBase,
```
wolf99: Should have removed the signature and the includes I think.

wolf99:
Forgot to remove commented line from example.c. Also left too much in all_your_base.h.

PR #154, opened 2017-07-05
Add all-your-base

wolf99:
Tried to have the implementation not require dynamic memory allocation. This makes the signature and associated tests slightly busier than otherwise, but still simple enough.


exercises/all-your-base/src/example.c
```diff
@@ -0,0 +1,37 @@
+#include "all_your_base.h"
+#include <math.h>
+
+size_t rebase(int16_t digits[DIGITS_ARRAY_SIZE], int16_t fromBase,
+              int16_t toBase, size_t numDigits)
+{
+   uint16_t denary = 0;
```
ryanplusplus: I had to look this one up. Apparently it's the base-10 equivalent of binary. Neat!

exercises/all-your-base/src/example.c
```diff
@@ -0,0 +1,37 @@
...
+   /* calculate and store new digits */
+   for (size_t i = DIGITS_ARRAY_SIZE - 1; denary > 0; --i) {
+      digits[i] = denary % toBase;
+      denary /= toBase;
+      ++newNumDigits;
+   }
+
+   /* ensure rest of array is zero */
+   for (int i = DIGITS_ARRAY_SIZE - (newNumDigits + 1); i >= 0; --i)
+      digits[i] = 0;
```
ryanplusplus: Any reason to not use `memset` for this?

exercises/all-your-base/test/test_all_your_base.c
```diff
@@ -0,0 +1,200 @@
...
+
+   TEST_ASSERT_EQUAL_INT32(expected_length, actual_length);
+
+   if (expected_length > 0)
+      TEST_ASSERT_EQUAL_INT16_ARRAY(ex_digits, digits, expected_length);
+}
+
+void test_single_bit_to_decimal(void)
+{
+   int16_t input[] = { 1 };
```
ryanplusplus: If you did:

```c
int16_t input[DIGITS_ARRAY_SIZE] = { 1 };
```

then it would automatically 0 fill the rest of the elements (making `fill_array` unnecessary).

Even better, this would allow you to get rid of the array length arguments to `test_rebase`. If we don't go this route, we should at least use `sizeof` and `sizeof(arr[0])` to calculate the number of elements instead of having them be hand-calculated.

exercises/all-your-base/src/example.h
```diff
@@ -0,0 +1,12 @@
+#ifndef ALL_YOUR_BASE_H
+#define ALL_YOUR_BASE_H
+
+#include <stddef.h>
+#include <stdint.h>
+
+#define DIGITS_ARRAY_SIZE 64
+
+size_t rebase(int16_t digits[DIGITS_ARRAY_SIZE], int16_t fromBase,
```
ryanplusplus: Can the digits array be an array of `uint8_t`s? Can the bases be `int8_t`s?

exercises/all-your-base/test/test_all_your_base.c
ryanplusplus: My bad, I thought `fill_array` was zero filling for some reason.

exercises/all-your-base/test/test_all_your_base.c
```diff
@@ -0,0 +1,200 @@
+#include "vendor/unity.h"
+#include <stddef.h>
+#include <stdint.h>
+#include "../src/all_your_base.h"
+
+#include <stdio.h>
+
+void fill_array(int16_t data[], size_t data_length, int16_t array[],
+                size_t array_length)
```
ryanplusplus: This is basically right justifying the array, right? That wasn't clear to me from the name alone. So, we could rename this, but I guess I'm left wondering why we need to right justify it in the first place. Can't we just let the data be left justified? I think that would take some of the mystery out of the test setup. As it stands, I think people will have some trouble figuring out exactly what format the input and output will be based without digging into the details of these helpers. I think we can remove some of the complexity from the helpers and make it more "what you see is what you get".

ryanplusplus: One thing that might allow us to keep the tests and implementation simple is to let `digits[0]` be the ones digit, not the most significant digit. This means that the digits will look "backwards" in the tests, but the digits will be in the correct order and placement in the tests without resorting to a reformatting helper and the implementation will not need to know the number of resulting digits ahead of time to properly format the output. As a bonus, the index will match the exponent of the base for the digit.

Thoughts?

wolf99: I realised I started with the right-justification because when initially designing the core logic the implementation didn't have the length of the input digits. With left-justification it wouldn't have known where to stop counting trailing zeros as digits ... but in the final implementation the function has the input digits length as an argument anyway so yes, the justification can be done away with.

As regards having the digit significance reversed.. I'm not sure without going in a bit more thoroughly but it would be moving away from the "what you see is what you get" idea a bit". 

I'll have a play around on both points this evening 👍

exercises/all-your-base/src/example.h
wolf99: Both could be `int8_t`s. The array cannot be unsigned however as the test spec allows (but errors on) negative numbers

exercises/all-your-base/src/example.c
wolf99: Good spot!

exercises/all-your-base/src/example.h
ryanplusplus: Got it

exercises/all-your-base/test/test_all_your_base.c
ryanplusplus: Makes sense and sounds good. This is surprisingly deep when implemented in C!

exercises/all-your-base/src/example.c
wolf99: BTW, would `memset()` actually have worked here? The data type of the array being `int` at the time.
It's changed to `uint8_t` now so it works now, but `memst()` is designed for strings and writes characters, so I'm guessing as part of it's optimisation it would at the least only end up clearing half the length of the `int` array?

wolf99:
I've kept the `digit[0]` as the least significant digit for the moment, simply because I like the input in the test cases to read as closely as the numbers really would (also matching the test config as closely as possible then). This did men an extra loop to calculate the number of digits required for the re-based number before actually calculating the digits of that number and writing them to the array, but I figure that's pretty minor

exercises/all-your-base/src/example.c
ryanplusplus: Yeah `memset` will work because while it's in `string.h`, it's really just filling things character by character. In order to zero out an entire `int` array, you just have to make sure that you provide it the length, in bytes, of the entire `int` array (with appropriate use of `sizeof`) and it will dutifully zero out every single byte. It doesn't know the type of the array (or even whether it's an array!) and views it as just a big block of memory.

ryanplusplus:
Looks great, thanks for refactoring! I agree that the extra loop isn't too bad and strikes a good balance between performance and usability.

PR #163, opened 2017-08-21
Update INSTALLATION.md with Visual Studio version

iam13islucky:
Visual Studio 2017 Community Edition is currently what is downloadable from that link, so it should reflect that change.


ryanplusplus:
👍

PR #165, opened 2017-08-25
simplify snippet

wolf99:
Unsure if the snippet is intended to be the easiest or the simplest. 
This is probably better for the latter option though as it reduces the code by 4 lines, one of which includes an entirely unused library.


ryanplusplus:
`NULL` comes from `stddef.h` so it's not unused, but I like this a lot more.

PR #166, opened 2017-09-15
deleted duplicate test runs in raindrops

Gamecock:
I am assuming this was an unintentional copy/paste error.  If it was intentional then reject this change.  If you let me know the pedagogical reason for running these tests twice I can add a comment, and my guess is we can get the same benefit just duplicating a couple.


ryanplusplus:
@Gamecock great catch! This is definitely not on purpose. Some test suites can randomize or repeat tests in order to ferret out subtle problems, but that wasn't the intent here.

PR #170, opened 2017-09-16
Added test cases for nullptr and empty string to acronym

Gamecock:
I added the test cases for nullptr and empty string to match the example implementation.  We can change the expected results and the example implementation if you want.  Second test should maybe return "", but either way is OK.


ryanplusplus:
Looks good. If you run:

```shell
git checkout master -- exercises/acronym/src/example.c
```

then it should revert to the old formatting for that file and you should be able to pass CI.

ryanplusplus:
Looks good, thanks!

PR #172, opened 2017-09-16
Detect and use gindent if available

### LINKED ISSUE #171: Use gindent if available in indent.sh
This will make it so that `indent.sh` does not have to be modified to be run on MacOS.

ryanplusplus:
Closes #171

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,9 @@ Please see the [Exercism contributing guide](https://github.com/exercism/docs/bl
 
 All test and example code should be written using the [Linux kernel coding style](https://www.kernel.org/doc/Documentation/CodingStyle) with 3 space indents and no tabs.
 
-If GNU `indent` is installed, this can be run on all files by executing `indent.sh`.
+If GNU `indent` is installed, this can be run on all files by executing `indent.sh`. On a Mac you can install the package `gnu-indent` using [Homebrew](http://brew.sh).
 
-To run it on a single file, you can execute:
+To manually run it on a single file, you can execute:
 
 ```shell
 indent -linux -i3 -nut $(file)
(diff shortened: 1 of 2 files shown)


patricksjackson:
Should this be accompanied by any change to the README?

I don't have a machine to test this with, but it looks fine.

ryanplusplus:
@patricksjackson good call. Notes added to `README.md`.

README.md
```diff
@@ -46,7 +46,7 @@ Each exercise should be contained in a directory ```c/exercises/<my exercise>```
...
+* `src` - contains the example files ```example.c``` and ```example.h```. These are both skipped by the ```exercism``` cli when downloading to the client, so it is imperative that you do not reference the names of the files in your code. If you need to provide a header file example that is necessary to run your tests it should be named ```<my exercise>.h``` instead.  Please also use [```include```](http://faculty.cs.niu.edu/~mcmahon/CS241/c241man/node90.html) guards in your header files. The tests can be run using the [```bin/run-tests```](https://github.com/exercism/c/blob/master/bin/run-tests) script which will rename the ```example.{c|h}``` files accordingly.
```
ryanplusplus: The GitHub diff view doesn't show it clearly, but all that happened here was that trailing whitespace was automatically removed by my editor.

Gamecock:
This works on OSX, I added a PR with an extra print that might be helpful.

Gamecock:
Stupid question, but I cannot find where this README appears on `exercism.io`, which is why it didn't help me.  Might be because the track is not launched yet.

ryanplusplus:
@Gamecock I'm honestly not sure. The documentation structure has changed a few times recently in the run-up to Exercism v2 and I don't know where stuff goes anymore.

patricksjackson:
That README doesn't show up on exercism.io. It's specifically for just this repo. My understanding of the split is that details about contributing to this track specifically go in that README, details about contributing to the exercism tracks generally can go in the docs repo https://github.com/exercism/docs, and docs for _using_ this track go in https://github.com/exercism/c/tree/master/docs, which show up on exercism.io under this tracks info.

PR #173, opened 2017-09-17
Added message output to indent.sh

Gamecock:
I think this would help new developers I missed the notes about the indent in the README.md

### DIFF
--- a/indent.sh
+++ b/indent.sh
@@ -3,16 +3,22 @@
 # Indents all .c and .h files in the project that are not vendor code (ie: Unity).
 #
 
-INDENT=indent
+INDENT=""
 
 if [ -x "$(command -v gindent)" ]; then
-   INDENT=gindent
+  INDENT=indent
+fi
+if [ -x "$(command -v gindent)" ]; then
+  INDENT=gindent
 elif [[ "$OSTYPE" == "darwin"* ]]; then
-   echo "***********************************************\n*  On OSX install gnu-indent using home brew  *\n***********************************************"
+  echo "***********************************************\n*  On OSX install gnu-indent using home brew  *\n***********************************************"
+fi
+if [[ -z "$INDENT" ]]; then
+echo "****************************************\n*  Install indent to run this script  *\n****************************************"
+exit 126
 fi
 
-
 for f in $(find exercises/ -not -path '*/vendor/*' -name '*.c' -or -not -path '*/vendor/*' -name '*.h'); do
-$INDENT -nbad -bap -nbc -bbo -hnl -br -brs -c33 -cd33 -ncdb -ce -ci4 -cli0 -d0 -di1 -nfc1 -i3 -nut -ip0 -l80 -lp -npcs -nprs -npsl -sai -saf -saw -ncs -nsc -sob -nfca -cp33 -ss -il1 $f
-rm $f~ 2> /dev/null
+  $INDENT -nbad -bap -nbc -bbo -hnl -br -brs -c33 -cd33 -ncdb -ce -ci4 -cli0 -d0 -di1 -nfc1 -i3 -nut -ip0 -l80 -lp -npcs -nprs -npsl -sai -saf -saw -ncs -nsc -sob -nfca -cp33 -ss -il1 $f
+  rm $f~ 2> /dev/null
 done
--- a/indent.sh
+++ b/indent.sh
@@ -6,10 +6,13 @@
 INDENT=indent
 
 if [ -x "$(command -v gindent)" ]; then
-  INDENT=gindent
+   INDENT=gindent
+elif [[ "$OSTYPE" == "darwin"* ]]; then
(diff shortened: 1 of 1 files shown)


patricksjackson:
Is there any reason to have this after the README changes @ryanplusplus is making in the other commit? And if so, it'd probably be better to have a general check to see if `indent` is installed instead of just for OSX.

Also, please try not to make whitespace edits unless there's a reason :)

Gamecock:
I like a meaningful error message better than a sentance in the docs, so I think it would help..  I can add the other check and revert the whitespace changes.  I had assumed the bash script was following the  same indenting rules as the code.  I cannot test it on anything other than OSX.

ryanplusplus:
I like the idea of having an error message, but second @patricksjackson's feelings regarding whitespace changes. However, I think we should actually generate an error instead of just a message. This will make the over the top formatting unnecessary.

Something like this:

```shell
echo "Missing GNU indent. You can blah blah blah"
exit 1
```

Gamecock:
Fixed the whitespace  and added the check for indent per @patricksjackson , did not see @ryanplusplus comment about the error code until after submitting.  I'll wait for any more comments before adding that.

indent.sh
```diff
@@ -3,10 +3,19 @@
 # Indents all .c and .h files in the project that are not vendor code (ie: Unity).
 #
 
-INDENT=indent
+INDENT=""
 
 if [ -x "$(command -v gindent)" ]; then
```
ryanplusplus: Should this be `$(command -v indent)`?

Gamecock:
I've addressed all the comments, but would like someone to run on a windows machine before this is merged.

ryanplusplus:
Changes look good to me. I fixed up some of the indentation, but otherwise everything looked fine.

@patricksjackson do you have a Windows machine that we can try this on? I only have MacOS and Linux at my disposal.

indent.sh
```diff
@@ -3,10 +3,22 @@
...
   INDENT=gindent
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+  echo "ERROR: install gnu-indent using home brew to run this script."
+  exit 1
+fi
+
+if [[ -z "$INDENT" ]]; then
+  echo "Error:  Install indent to run this script."
+  exit 126
```
patricksjackson: Is there some significance to 126?

ryanplusplus: Not sure if it was intended, but http://tldp.org/LDP/abs/html/exitcodes.html

patricksjackson:
This script has a bug in Ubuntu (which is the most popular distro for WSL on Windows, and the CI). Note in the CI output it throws warnings too. The issue is with processing the two if statements with double brackets, so it is running indent on Travis, it will just fail to spit out the nice error message.

It's because Ubuntu defaults to dash instead of bash, we can fix it in the CI script if we explicitly invoke the indent script with `bash indent.sh`, but it may be worthwhile making it fully dash compliant for anybody using Ubuntu.

Also just FYI, but this repo doesn't really work on normal Windows at all without using some form of unix. I found a bug in Unity that prevents it from compiling in the Visual Studio compiler a few months ago but I never followed up on it, and I haven't taken the time to double check or fix it.

ryanplusplus:
@patricksjackson Thanks for the info. I tested on an LTS machine that still has 16.04 so that didn't show up for me. I have another machine with 17.04 so I can test once this is updated.

I suspect that using `#!/bin/bash` will work fine on MacOS, but we can probably also change the offending bit to `[ "$INDENT" = "" ]`.

indent.sh
Gamecock: Yes,  I did read (google) the documentation.

Gamecock:
@ryanplusplus  suggestion on the if syntax worked for me, but the old one worked for me also.

ryanplusplus:
@patricksjackson you should be able to update @Gamecock's PR by adding another remote to your clone:

```
git remote add <remote name> <url>
```

That will let you check out his branch and push to it after you've made changes:

```
git push <remote name> <branch name>
```

ryanplusplus:
@patricksjackson I added your change using the very lazy method: I simply took the contents of your copy pasted over the top of the existing code in the web editor. No muss, no fuss.

patricksjackson:
Thanks @ryanplusplus I knew there was some way I was some way of doing it in git :)

I also had no idea you could just edit on the webpage...that makes it even easier. We have that feature at work though I should've figured it was here as well...

This change looks fine to me after we ack that works on Mac.

ryanplusplus:
Okay, it now works on Mac and Ubuntu 17.10. I had to reorganize a bit since `indent` _is_ available on Mac, it's just not GNU indent (presumably it's of BSD origin). So now it checks if `gindent` is available and if not and you're on a Mac it fails out. Only after that does it search for `indent` and then ultimately fail out if it can't find anything.

patricksjackson:
This looks good, I'll go ahead and merge since I believe we've now tested everywhere.

PR #176, opened 2017-09-22
Update unit tests all-your-base

### LINKED ISSUE #175: all-your-base exercise assigns negative numbers to unsigned data types.
All-your-base exercise has a couple of unit tests that assign negative numbers to uint data types.
example:
```
void test_first_base_is_negative(void)
{
   int8_t input[] = { 1 };
   int8_t expected[] = { 0 };
   test_rebase(-2, input, LENGTH(input), 10, expected, 0);
}
```
test_rebase is defined in the test case as:
```
void test_rebase(uint16_t input_base, int8_t input_digits[],
                 size_t input_length, uint16_t output_base,
                 int8_t expected_digits[], size_t expected_length)
```

We should  either just delete those test cases or change the function prototype to allow negative bases.
I think deleting the test case is more `C` like, allowing negative numbers there seems like a `Java` thing. 

One other test case was not implemented per the [cannonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/all-your-base/canonical-data.json)
```
    {
      "description": "negative digit",
      "property": "rebase",
      "input_base": 2,
      "input_digits": [1, -1, 1, 0, 1, 0],
      "output_base": 10,
      "expected": null
    },```
The `C` track is implemented with a negative base instead of a negative.  I can fix that case, or just delete it and make the arrays uint8_t[] vs int8_t[].

Let me know how you want me to proceed.

Gamecock:
Change helper function from uint16_t to int16_t.
Fix negative data test case

Closes #175

### DIFF
--- a/exercises/all-your-base/test/test_all_your_base.c
+++ b/exercises/all-your-base/test/test_all_your_base.c
@@ -15,8 +15,8 @@ void copy_array(int8_t src[], int8_t dest[DIGITS_ARRAY_SIZE], size_t n)
       dest[i] = src[i];
 }
 
-void test_rebase(uint16_t input_base, int8_t input_digits[],
-                 size_t input_length, uint16_t output_base,
+void test_rebase(int16_t input_base, int8_t input_digits[],
+                 size_t input_length, int16_t output_base,
                  int8_t expected_digits[], size_t expected_length)
 {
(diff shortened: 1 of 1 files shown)


ryanplusplus:
LGTM, but I'm going to give @patricksjackson a chance to look before merging.

patricksjackson:
These changes look fine 👍 I'll go ahead and merge

Issue #167, opened 2017-09-15
C runs all tests the first time.


Gamecock:
Normally in TDD you get one test to fail, get it to pass, refactor, start on next test.  In the C track all tests run the first time you build.
We could use the macros from cpp fairly easily.  
```
#if defined(EXERCISM_RUN_ALL_TESTS)
#endif
```
I happy to contribute to this effort if we want to do it.  I might need some guidance on everywhere we need to change.

patricksjackson:
To add this you'd need to make that modification to all of the test suites and change the make call in [bin\run-tests](https://github.com/exercism/c/blob/876941a8ee876439ea4863a4ccb131664d677145/bin/run-tests#L27) (I believe to `make -DEXERCISM_RUN_ALL_TESTS clean >> /dev/null`, but I may have confused the syntax for that call.)

If I recall correctly, the reason I didn't bring this up before is that you need to add the #define's around the calls to `RUN_TEST` that are at the bottom of the test suite files, and adding them around the function definitions is optional. Since normally when you do the problem you tend to just look at the function definitions, it's a little clunkier for the end-user in this repo than it is in cpp where each test doesn't need to be listed twice.

Depending on how this ultimately looks to the end-user it might not be a given that this should go in. I'd suggest waiting for @ryanplusplus to weigh in until working on it.

(Alternatively, we could probably hack together a way to do it the way cpp (Boost Test) does, so essentially get rid of the `main()` definition in the test suite file.)

ryanplusplus:
@patricksjackson an alternative is to do some source modification within the CI test script. For instance, we could find all commented lines within the test file's `main()` and uncomment them on the fly. This wouldn't be too difficult and avoids leaving anything confusing looking inside of the test files. We already do something similar when we move the skeleton out of the way and replace it with the examples.

Thoughts?

Gamecock:
How do you run the `bin\run-tests`?  I need to figure that out, and then I can update the appropriate documentation page.
Once I've done that I might have a better idea of how to implement it.  I can dig around in Unity a little also.

ryanplusplus:
@Gamecock if you're on MacOS or Linux, you can just execute `./bin/run-tests` in the terminal. If you're on Windows, you should be able to do the same from Cygwin. You also have the option of running Ubuntu on Windows: https://msdn.microsoft.com/en-us/commandline/wsl/about

patricksjackson:
[...]

PR #179, opened 2017-09-24
Sample of ignoring tests

Gamecock:
Sample to have test run 1 at a time for #167.
Changed isogram to run 1 test at a time.  Modified bin/run-tests to run all tests.

### DIFF
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -3,12 +3,15 @@ set -e
 
 for D in exercises/*; do
     CURRENT_DIR=$(pwd)
-    echo ${D}
+
     if [ -d "${D}" ]; then
         # Get the exercise name from the test file
         TEST_FILE=$(basename $(ls ${D}/test/test*))
         STRIPPED_OF_EXTENSION="${TEST_FILE%.*}"
         EXERCISE_NAME="${STRIPPED_OF_EXTENSION:5}"
+
+        #remove the ignore line
+        sed -i.bak 's/TEST_IGNORE();//' ${D}/test/test_*.c
 
         # Copy the examples with the correct name for the exercise
         if [ -e "${D}/src/example.c" ]
@@ -25,12 +28,12 @@ for D in exercises/*; do
         { cd ${D};
           echo "Running tests for ${EXERCISE_NAME}";
           make clean >> /dev/null;
-          if make CFLAGS='-DRUN_ALL_TESTS -Wno-macro-redefined' | grep FAIL: ; then
+          if make | grep FAIL: ; then
             exit 1
           elif [ ${PIPESTATUS[0]} -ne 0 ]; then
             exit 1
           fi
-#          valgrind --quiet --leak-check=full --error-exitcode=1 ./tests.out >> /dev/null;
+          valgrind --quiet --leak-check=full --error-exitcode=1 ./tests.out >> /dev/null;
           cd ${CURRENT_DIR};
         }
     fi
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -3,7 +3,7 @@ set -e
 
 for D in exercises/*; do
     CURRENT_DIR=$(pwd)
-
+    echo ${D}
     if [ -d "${D}" ]; then
         # Get the exercise name from the test file
         TEST_FILE=$(basename $(ls ${D}/test/test*))
@@ -25,12 +25,12 @@ for D in exercises/*; do
         { cd ${D};
           echo "Running tests for ${EXERCISE_NAME}";
           make clean >> /dev/null;
-          if make | grep FAIL: ; then
+          if make CFLAGS='-DRUN_ALL_TESTS -Wno-macro-redefined' | grep FAIL: ; then
             exit 1
           elif [ ${PIPESTATUS[0]} -ne 0 ]; then
             exit 1
           fi
-          valgrind --quiet --leak-check=full --error-exitcode=1 ./tests.out >> /dev/null;
+#          valgrind --quiet --leak-check=full --error-exitcode=1 ./tests.out >> /dev/null;
           cd ${CURRENT_DIR};
         }
     fi
--- a/exercises/isogram/test/test_isogram.c
+++ b/exercises/isogram/test/test_isogram.c
@@ -2,10 +2,6 @@
 #include "../src/isogram.h"
 #include <stdlib.h>
 
-#ifdef RUN_ALL_TESTS
-#define TEST_IGNORE_MESSAGE(X)
-#endif
-
 void test_empty_string(void)
 {
    TEST_ASSERT_TRUE(isIsogram(""));
@@ -13,7 +9,7 @@ void test_empty_string(void)
 
 void test_lower_case_only(void)
 {
-   TEST_IGNORE(); // delete this line to run test
+   TEST_IGNORE();               // delete this line to run test
    TEST_ASSERT_TRUE(isIsogram("isogram"));
 }
 
--- a/exercises/isogram/test/test_isogram.c
+++ b/exercises/isogram/test/test_isogram.c
@@ -2,6 +2,10 @@
 #include "../src/isogram.h"
 #include <stdlib.h>
 
+#ifdef RUN_ALL_TESTS
+#define TEST_IGNORE_MESSAGE(X)
+#endif
+
 void test_empty_string(void)
 {
    TEST_ASSERT_TRUE(isIsogram(""));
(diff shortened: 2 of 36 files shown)


exercises/isogram/test/test_isogram.c
```diff
@@ -2,48 +2,60 @@
 #include "../src/isogram.h"
 #include <stdlib.h>
 
+#ifdef RUN_ALL_TESTS
+#define TEST_IGNORE_MESSAGE(X)
```
ryanplusplus: Clever. Does it complain about redefining this?

Gamecock: not on my MAC, but Travis doesn't like it.

bin/run-tests
```diff
@@ -25,7 +26,7 @@ for D in exercises/*; do
         { cd ${D};
           echo "Running tests for ${EXERCISE_NAME}";
           make clean >> /dev/null;
-          if make | grep FAIL: ; then
+          if make CFLAGS='-DRUN_ALL_TESTS -std=c99 -Wall -Wextra -pedantic -Werror ' | grep FAIL: ; then
```
Gamecock: I know his is ugly, but fixing it requires touching every make file.  I can if we think it's worth the effort.

bin/run-tests
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
 
+
```
Gamecock: I'll eliminate the whitespace change in the final commit.

Gamecock:
Running in gcc on Travis and clang on OSX, that's all the environments I have available to test in.

bin/run-tests
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
```
Gamecock: Completely unrelated, I can't find -e option in the [manual](http://man7.org/linux/man-pages/man1/bash.1.html#OPTIONS), does anyone know it's purpose?

wolf99: http://www.davidpashley.com/articles/writing-robust-shell-scripts/#id2382181

Gamecock: Thanks

bin/run-tests
ryanplusplus: Eh, I think we should probably touch them all. All we need to do is change the `=` to a `+=` on the first line. I think this is better than scratching our heads later.

ryanplusplus:
Sorry for the delay, I've been out of town. I think that this is generally looking pretty good. My vote is to fix the last few nitpicks and then apply this to all exercises.

bin/run-tests
Gamecock: Do you have access to a Windows machine to test this on before I go change everything?

ryanplusplus: No. @patricksjackson?

Gamecock:
@ryanplusplus do you want a new clean PR to implement this across everything so it does not have the 10 tries I took at making it work?

ryanplusplus:
@gamecock don't worry about it; I can squash when I merge it. If you're interested in trying it out, it's also not too hard to squash on the CLI.

bin/run-tests
patricksjackson: I agree, the first line of the default makefile should be updated to `CFLAGS += -std=c99` anyways. The fast way (in case you were planning on doing it manually):
```
find . -name makefile -print0 | xargs -0 sed -i 's/CFLAGS  =/CFLAGS +=/g'
```

For changes to the Makefiles / scripts, if it runs on Ubuntu/Other Linux just assume it works on Windows. We should just push people to use the WSL for that and that is literally Ubuntu LTS (or SUSE, or Fedora (soon), and hopefully others in future).

The shell scripts are an issue (with the repo checked out in Windows they have CRLF line endings, and the shell programs don't like that), but that's just a matter of changing the line endings with `dos2unix`. `make` though works on CRLF files, so they can use it extremely easily.

It's probably on me to make a big README update for Windows before the track launches / soon.

exercises/isogram/test/test_isogram.c
```diff
@@ -2,48 +2,61 @@
...
+#endif
+
 void test_empty_string(void)
 {
    TEST_ASSERT_TRUE(isIsogram(""));
 }
 
 void test_lower_case_only(void)
 {
+   TEST_IGNORE_MESSAGE("Delete this line after the above test passes");
```
patricksjackson: I'd prefer this to be `TEST_IGNORE() // Delete this line to run the test`, and maybe only need to have the comment on the first one in each test suite. The message makes the first run output very busy:
```
Compiling tests.out
test/test_isogram.c:67:test_empty_string:PASS
test/test_isogram.c:17:test_lower_case_only:IGNORE: Delete this line after the above test passes
test/test_isogram.c:23:test_duplicated_letter:IGNORE: Delete this line after the above test passes
test/test_isogram.c:29:test_longest_known_isogram:IGNORE: Delete this line after the above test passes
test/test_isogram.c:35:test_duplicated_letter_mixed_case:IGNORE: Delete this line after the above test passes
test/test_isogram.c:41:test_non_letter_char:IGNORE: Delete this line after the above test passes
test/test_isogram.c:47:test_duplicated_non_letter_char:IGNORE: Delete this line after the above test passes
test/test_isogram.c:53:test_multiple_whitespace:IGNORE: Delete this line after the above test passes
test/test_isogram.c:59:test_duplicated_letter_within_word:IGNORE: Delete this line after the above test passes

-----------------------
9 Tests 0 Failures 8 Ignored
OK
```

patricksjackson:
I really don't like the need for us to handle the `RUN_ALL_TESTS` setting within the test suite.

The alternative (outside of patching Unity) is we can run a custom sed script in the `bin/run-tests` script. e.g. adding `sed -i s/TEST_IGNORE();//' test/test_*.c` if we took out the message, or `sed -i 's/TEST_IGNORE_MESSAGE(.*);//' test/test_*.c` if we left it like it is here.

What do you think of that approach?

The downside is that way makes it more difficult to add a custom make target for all tests if we wanted.

ryanplusplus:
@patricksjackson my initial thought was to modify the source as you suggested with `sed` so I'm fine with that. It definitely cleans up the test suite and gives the user fewer things to scratch their head about.

I also second your `TEST_IGNORE` suggestion. It should have been obvious, but I didn't realize it was an option.

Gamecock:
If you like this pattern I can implement it across the rest of these tests.

patricksjackson:
I am ok with this pattern. Sounds like @ryanplusplus is as well but might want to make sure he acks this final version just in case.

I had to locally verify that the new bin/run-tests works and runs all the tests, since it doesn't print out the list of tests run. It's not an issue, just noting it here that I did check that :)

Gamecock:
There are a couple of tests that are created by expanding XMACROS, that cannot do the TEST_IGNORE().  I could do manual comment outs, but do not know how bin/run_tests.sh could override.  I assume you are OK skipping?
```
// Test cases
#define SERIES_TEST_CASES(ENTRY) \
   ENTRY(test_slices_of_one, "01234", 1, 5, { ASSEMBLE_STRING_ARRAY("0","1","2","3","4") })\
   ENTRY(test_slices_of_two, "97867564", 2, 7, { ASSEMBLE_STRING_ARRAY("97","78","86","67","75","56", "64") })\
   ENTRY(test_slices_of_three, "97867564", 3, 6, { ASSEMBLE_STRING_ARRAY("978","786","867","675","756","564") })\
   ENTRY(test_slices_of_four, "01234", 4, 2, { ASSEMBLE_STRING_ARRAY("0123","1234") })\
   ENTRY(test_slices_of_five, "01234", 5, 1, { ASSEMBLE_STRING_ARRAY("01234") })\
   ENTRY(test_overly_long_slice, "012", 4, 0, { ASSEMBLE_STRING_ARRAY("") })\
   ENTRY(test_overly_short_slice, "01234", 0, 0, { ASSEMBLE_STRING_ARRAY("") })\

SERIES_TEST_CASES(EXPAND_SERIES_TEST_CASES_AS_TEST)

ryanplusplus:
@Gamecock The latest changes look good. Regarding the XMACROs tests, I think we should probably just rework those tests regardless. I think it will be a win for simplification and user understanding if we just write them like "normal" tests. We can do this here if it's simple or on another ticket/PR if it's going to be more work.

Gamecock:
This is already in the README.

```Create just the functions you need to satisfy any compiler errors and get the
test to fail. Then write just enough code to get the test to pass. Once you've
done that, move onto the next test.```

Between that and the comment on the first line of every file,
`   TEST_IGNORE();               // delete this line to run test`
 I think it's enough documentation and this is ready to merge.

Issue #184, opened 2017-09-30
all: Naming style inconstencies


NobbZ:
I'm randomly doing some of the C exercises and I realised, that some do use camelCase, others snake_case for the functions to implement.

Is there a certain reason for this inconsistency or is a PR uniforming those welcome? If yes, should it favor snake_case or camelCase?

ryanplusplus:
A PR would absolutely be welcome. We're trying to adhere to the [Linux kernel coding style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html) which uses snake case, so snake case should be favored.

sjwarner-bp:
I'd be interested in trying to do this. I might just pick up one exercise with the inconsistency and issue a PR just to make sure that I've got the right idea, then do a few more if that is ok?

wolf99:
@ryanplusplus this would probably be a good candidate for the "good first commit" label

sjwarner-bp:
Please ignore that reference and pull request! I will try again.

sjwarner-bp:
I think that PR was ok. Please let me know if I have done anything wrong, otherwise I will look to fix up a few more of these exercises. Sorry for the clutter above!

wolf99:
@sjwarner-bp , this issue as currently stated is for 

> the functions to implement

That is, the functions that the student see's in the interface implied by the tests (and possibly the header), i.e. "public" functions. In the PR #185 you have changed the name for a `static` function. While it might be good style to have *all* functions named consistently, I'm not sure that this is what this issue was aimed at. Maybe @NobbZ could clarify on that point?

Further it would maybe be better to have this work for all exercises done under a single PR, rather than a PR per exercise... might wait for word from @ryanplusplus  or @patricksjackson though 😃

NobbZ:
The name of the FUTs is probably the first name in each suites that a student sees. And I found different naming schemes a bit confusing.

In #185 only the example implementation was changed, while the tests would still try call `getVerse`. So basically, things that needs to get changed are stubs (for exercises which have them), tests, and the example code.

For me, beeing consistent with the names of the FUTs is the most important part. If though actual test names follow the same conventions that would be a plus.

---

FUT: Function under test

sjwarner-bp:
I agree, though they were static methods it still probably wants to be consistent to avoid confusion.

[...]

PR #197, opened 2017-10-12
Update palindrome product exercise to snake case

wolf99:
Update palindrome product exercise to snake case per #184 

Also took the chance to make the type names more relevant to it's actual contents and add new lines between the test cases and the `UnityBegin()` & `UnityEnd()` calls as per other exercises


wolf99:
Forgot to rebase first!

PR #199, opened 2017-10-13
Update robot simulator exercise to snake case

wolf99:
Update robot simulator exercise to snake case per #184 

Am unsure about enum members casing. IIRC the Linux kernel style is to use all-cap snake case, like:

```c
enum {
        LIKE_THIS
}
```

But from memory I don't think any of the other exercises use this so have left all lower for now.

### DIFF
--- a/exercises/robot-simulator/src/example.c
+++ b/exercises/robot-simulator/src/example.c
@@ -1,48 +1,48 @@
 #include <string.h>
 #include "robot_simulator.h"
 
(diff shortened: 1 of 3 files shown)


Gamecock:
I think we should adhere to the coding style and update all constants and enum labels to ALL_CAPS in a separate PR, but could be easily talked out of that position with an argument deeper than `too much work`

ryanplusplus:
I agree with @Gamecock -- let's do this in other PRs.

PR #205, opened 2017-10-14
Add setUp and tearDown to be compatible with MSVC

### LINKED ISSUE #174: Tests don't compile with MSVC (Windows)
**TLDR** - There are issues with doing the exercises under Visual Studio (and likely other non-GCC/Clang compilers like ICC). This can be fixed by adding the following function defs to all of the `test\test_*.c` files.

```
void setUp(void) { }
void tearDown(void) { }
```

This is because Unity uses a non-standard method to support optional `setUp` and `tearDown` methods. GCC and Clang methods are implemented in Unity (they support `__attribute__((weak))` and/or `#pragma weak` I think). Other compilers may do the same, but MSVC doesn't support those as far as I can tell.

Someone could fix the issue for us by patching Unity with the correct method for MSVC (you can pass comments to the linker using a pragma), but that doesn't fix it for other compilers that don't support the weak linking either.

It does work fine using the various unix implementations on Windows (WSL and MSYS) since they just use GCC.

ryanplusplus:
Fixes #174

### DIFF
--- a/exercises/acronym/test/test_acronym.c
+++ b/exercises/acronym/test/test_acronym.c
@@ -3,6 +3,14 @@
 #include <stdlib.h>
 #include <string.h>
 
+void setUp(void)
+{
+}
+
+void tearDown(void)
+{
+}
+
 void test_abbreviation(char *phrase, char *expected)
 {
    char *actual = abbreviate(phrase);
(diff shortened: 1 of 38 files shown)


ryanplusplus:
@patricksjackson can you do a quick check on at least one exercise to make sure that this actually works on MSVC?

patricksjackson:
Works fine on MSVC. I didn't exhaustively check everything, but it LGTM

PR #208, opened 2017-10-16
Update beer song exercise to snake case

wolf99:
Update beer song exercise to snake case per #184


wolf99:
Had to remove a new line in example.c before indent picked up that changes were needed. Not sure if this is down to my platform (bash on WSL) or something with indent. 

I ran indent individually on each file before committing initially, then ran it and the indent.sh script again after Travis reported the failure - all times it reported no changes needed.

Removed a new line already in place on a longer line (function signature) , then ran indent again on the file and it picked up both the changed line and the `for()` line for changes. Done now anyway 😄

PR #178, opened 2017-09-23
Simplify Hello World and sync it with canonical JSON

ryanplusplus:
Closes #177

PR #210, opened 2017-10-16
Update snippet to follow hello-world exercise

wolf99:
hello-world exercise changed with #178 to match canonical json


patricksjackson:
At some point we could update this to be something other than Hello World, but matching this with the actual solution for now is a good thing. Thanks.

PR #211, opened 2017-10-16
Fix TEST_IGNORE on allergies

Gamecock:
TEST_IGNORE was applied to all tests and one helper function, so the first time `make test` was run it ran 0 tests.

### DIFF
--- a/exercises/allergies/test/test_allergies.c
+++ b/exercises/allergies/test/test_allergies.c
@@ -10,12 +10,13 @@ void tearDown(void)
 {
 }
 
-void test_list_count_is(int count, allergen_list_t * list)
+//helper functions
+void list_count_is(int count, allergen_list_t * list)
 {
    TEST_ASSERT_EQUAL(count, list->count);
 }
 
-void test_list_contains(allergen_t allergen, allergen_list_t * list)
+void list_contains(allergen_t allergen, allergen_list_t * list)
 {
    bool allergen_found = false;
 
@@ -65,7 +66,7 @@ void test_no_allergies_at_all(void)
 
    get_allergens(score, &list);
 
-   test_list_count_is(0, &list);
+   list_count_is(0, &list);
 
    free(list.allergens);
 }
@@ -78,8 +79,8 @@ void test_allergic_to_just_eggs(void)
 
    get_allergens(score, &list);
 
-   test_list_count_is(1, &list);
-   test_list_contains(ALLERGEN_EGGS, &list);
+   list_count_is(1, &list);
+   list_contains(ALLERGEN_EGGS, &list);
 
    free(list.allergens);
 }
@@ -92,8 +93,8 @@ void test_allergic_to_just_peanuts(void)
 
    get_allergens(score, &list);
 
-   test_list_count_is(1, &list);
-   test_list_contains(ALLERGEN_PEANUTS, &list);
+   list_count_is(1, &list);
+   list_contains(ALLERGEN_PEANUTS, &list);
 
    free(list.allergens);
 }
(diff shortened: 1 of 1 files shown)


patricksjackson:
So there was some confusion about which of the functions were helper functions? It looks like these helpers at least were named with the same `test_` prefix as actual tests, I think we should have some way of more easily distinguishing the actual tests from the helper functions.

Any ideas about how we could make them more distinguishable? Comment blocks?

wolf99:
a `helper_` prefix maybe?

ryanplusplus:
I like `helper_` -- we can eliminate confusion without resorting to an apologetic comment

Gamecock:
`//Tests start here`
This was already in some of the test cases .  To minimize confusion for us and the users I think we should forward declare the helper functions, write the tests, run them, and then have the implementation.  I didn't show this with the prefix change, but we could do that also, or just delete the prefix `list_count_is`.

```//Declare helpers 
void test_list_count_is(int count, allergen_list_t * list);
void test_list_contains(allergen_t allergen, allergen_list_t * list);

void test_no_allergies_means_not_allergic(void)
{
   int score = 0;

   TEST_ASSERT_FALSE(is_allergic_to(allergen_peanuts, score));
   TEST_ASSERT_FALSE(is_allergic_to(allergen_cats, score));
   TEST_ASSERT_FALSE(is_allergic_to(allergen_strawberries, score));
}

void test_is_allergic_to_eggs(void)
{
   TEST_IGNORE();               // delete this line to run test
   int score = 1;

   TEST_ASSERT_TRUE(is_allergic_to(allergen_eggs, score));
}

... other test ....

int main(void)
{
   UnityBegin("test/test_allergies.c");

   RUN_TEST(test_no_allergies_means_not_allergic);
   RUN_TEST(test_is_allergic_to_eggs);
... other tests ...

   UnityEnd();
   return 0;
}

// helper function implementations
void test_list_count_is(int count, allergen_list_t * list)
{
   TEST_ASSERT_EQUAL(count, list->count);
}

void test_list_contains(allergen_t allergen, allergen_list_t * list)
{
   bool allergen_found = false;

   for (int i = 0; i < list->count; i++) {
      if (list->allergens[i] == allergen) {
         allergen_found = true;
         break;
      }
   }

   TEST_ASSERT_TRUE(allergen_found);
}

```

wolf99:
WRT the `helper_` prefix idea I had initially thought this would result in something like `helper_do_foo()`, i.e. replacing any other prefix. But I can see from the gamecock's example above that it is logical in some cases to still also retain a `test_` prefix, such as in the `test_list_contains()` case. This means that adopting the `helper_` prefix could make some of the function names become overly long and may start to cause problems with `indent` overly mutilating lines that such functions are called on.

Gamecock:
Looking at some of our other test cases the helper functions just have a function name w/o any prefix.
Example from robot simulator:
```// Test Helper Function
void confirm_position(RobotGridStatus_t * expected, RobotGridStatus_t * actual)
{
   TEST_ASSERT_EQUAL(expected->bearing, actual->bearing);
   TEST_ASSERT_EQUAL(expected->grid.x_position, actual->grid.x_position);
   TEST_ASSERT_EQUAL(expected->grid.y_position, actual->grid.y_position);
}
```
Do we want to add `helper_` and forward declare all, or just remove `test_` from this one?

ryanplusplus:
For consistency's sake my vote is to just go with removing `test_`

ryanplusplus:
Sorry I didn't catch this before, but we already covered enums in Allergies: https://github.com/exercism/c/blob/master/exercises/allergies/src/allergies.h. If you merge `master` into your branch you'll get these changes.

Gamecock:
I got a little rebase practice, but it's good now.

wolf99:
Just noticed that this does not seem to have the `setUp()` and `tearDown()` functions, unless 'm missing something.

Gamecock:
@wolf99 thanks, got removed by mistake when fixing merge conflicts.

ryanplusplus:
LGTM

patricksjackson:
Awesome :)

PR #217, opened 2017-10-17
Create C style guide

wolf99:
Initial attempt at the style guide.

Am still unsure on if the capitalization for constants applies only to globals or not.

I intend to add further commits with a short description of the formatting as set by the parameters we pass to `indent` (line length, brace placement, etc)

### DIFF
--- a/docs/C_STYLE_GUIDE.md
+++ b/docs/C_STYLE_GUIDE.md
@@ -12,7 +12,7 @@ In general, the names of types, variables, structures and files should use snake
 void some_foo_function(char * input_string, int input_string_length);
 ```
 
-## Enums
+### Enums
 
 Enum names should usually be singular rather than plural.
 
@@ -27,7 +27,7 @@ enum heading {
 }
 ```
 
-## Macro names
+### Macros
 
 Macros defining constants are named using snake case with all letters in upper case.
 
@@ -37,7 +37,7 @@ Macros defining constants are named using snake case with all letters in upper c
 
 Upper-case snake case for macro names is appreciated, but macros resembling functions may be named in lower-case snake case.
 
-## Constant names
+### Constants
 
 Names for variables defined as `const` (global or otherwise) should use snake case with all letters in upper case.
 
@@ -45,7 +45,7 @@ Names for variables defined as `const` (global or otherwise) should use snake ca
 const int LIMIT = 22;
 ```
 
-## Type Names
+### Types
 
 Types defined by a program using `typedef` should be named with a `_t` suffix. 
 While aware that POSIX reserves the `_t` suffix, the track does not currently utilise POSIX.
--- a/docs/C_STYLE_GUIDE.md
+++ b/docs/C_STYLE_GUIDE.md
@@ -55,3 +55,55 @@ const int LIMIT = 22;
 
 Types defined by a program using `typedef` should have a `_t` suffix. While aware 
 that POSIX reserves the `_t` suffix, the track does not currently utilise POSIX.
+
+
+## Test file layout
+
+Each exercise's test file follows a particular layout. The layout is intended to 
+both help contibutors to identify parts of the file, and to help learners to 
+understand how the tests work
+
+Like most C source files, a test file begins with includes. 
+
+```c
+#include "vendor/unity.h"
+#include "../src/file_under_test.h"
+// any other includes here
+```
+The includes are then followed by any helper functions that are used by the test 
+functions. The names of such helper functions should not use the `test_` prefix.
+
+Next there are the `setUp()` and `tearDown()` functions. So far, no exercise uses 
+these (the function bodies are empty) yet they must remain in place to allow the 
+test files to compile correctly across all platforms.
+
+```c
+void setUp(void)
+{
+}
+
+void tearDown(void)
+{
+}
+```
+
+Next are the test functions themselves. The names of these functions use a `test_` 
+prefix.
+
+Lastly is the `main()` function. The funciton body of main follows a particular 
+layout itself. First a call to `UnityBegin()` followed by an empty line, then the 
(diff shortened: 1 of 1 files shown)


docs/C_STYLE_GUIDE.md
```diff
@@ -0,0 +1,57 @@
+# C Style Guide
+
+This guide describes the style intended for use with the C track exercise example 
+and test files.
+
+The rules within this guide are in addition to, or to call out, those in the 
+[Linux kernal coding style guide](https://www.kernel.org/doc/html/latest/process/coding-style.html).
```
ryanplusplus: `kernal` => `kernel`

docs/C_STYLE_GUIDE.md
```diff
@@ -0,0 +1,57 @@
...
+lower-case snake case, where underscores `_` are used to separate name 
+components.
+
+```c
+void some_foo_function(char * input_string, int input_string_length);
+```
+
+## Enums
+
+Enum names should usually be singluare rather than plural.
```
ryanplusplus: `singluare` => `singular`

docs/C_STYLE_GUIDE.md
```diff
@@ -0,0 +1,57 @@
...
+
+Enum labels should be capitalized.
```
ryanplusplus: Should we call this out as upper snake case?

docs/C_STYLE_GUIDE.md
```diff
@@ -0,0 +1,57 @@
...
+## Macro names
+
+Macros defining constants are capitalized.
+
+```c
+#define CONSTANT 0x4000
+```
+
+Capitalized macro names are appreciated, but macros resembling functions may be
+named in lower case.
```
ryanplusplus: lower snake case?

docs/C_STYLE_GUIDE.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Constant names
+
+Names for variables defined as `const` should be capitalized.
```
ryanplusplus: Upper snake case?

ryanplusplus:
If the kernel style guide only wants UPPER_SNAKE_CASE for globals, I think we should deviate from them on that point. It feels arbitrary and fiddly.

ryanplusplus:
Looks good so far. My only real suggestion is that we should be consistent in how we describe the style (upper snake case vs. just capital/uppercase).

wolf99:
@ryanplusplus thanks for those comments, I have updated with regard to those in f1c282e

I have also added some commentary on the layout of the test file - not sure if this exactly belongs in a style guide but it seems like handy information to have available

docs/C_STYLE_GUIDE.md
```diff
@@ -0,0 +1,112 @@
...
+snake case with all letters in lower case, where underscores `_` are used to 
+separate name components.
+
+```c
+void some_foo_function(char * input_string, int input_string_length);
+```
+
+## Enums
+
+Enum names should usually be singluar rather than plural.
```
ryanplusplus: `singluar` => `singular`

docs/C_STYLE_GUIDE.md
```diff
@@ -0,0 +1,112 @@
+# C Style Guide
+
+This guide describes the style intended for use with the C track exercise 
+example and test files.
```
ryanplusplus: A minor issue: Exercism-wide it was decided to not add manual breaks in lines in Markdown documents. In order to keep things fairly readable (and writable), each sentence should be on its own line. This makes it so that lines don't get forever long in editors that don't soft wrap while also making it so that minor updates to a Markdown document don't require lines to be manually re-wrapped.

This is a fairly recent thing and I'm personally okay with doing no manual wrapping, but if we are going to wrap, let's do it in the recommended way.

ryanplusplus:
@wolf99 I think the layout of the test file is definitely helpful. My vote is to keep it.

docs/C_STYLE_GUIDE.md
wolf99: Dammit! I type too fast and then read only what I expect to be there when I check it. Thanks for catching that.

docs/C_STYLE_GUIDE.md
wolf99: Ah, good to know, I'll adjust to suit

ryanplusplus:
Looks good so far. Are you still planning to add info about `indent`?

wolf99:
Yep, that should be next up

wolf99:
@ryanplusplus  I relised that if non-global constants are to be all-caps,  then that means capitalized names showing up in some unusual places... such as function parameters, e.g.

```c
void some_function_foo(const char *BUFFER, const int BUFFER_LENGTH);
```

To me that looks a bit odd TBH.

ryanplusplus:
@wolf99 -- @Gamecock checked the Linux kernel source and didn't find any non-`#define`d, non-enumerated upper snake case constants. So I think we're good to just leave them as they are and go with lower snake case.

wolf99:
@ryanplusplus Have removed the cont style info then. Now this should be good to go, unless you can think of anything else that could be added or adjusted?

Gamecock:
New question, but maybe should merge this and save for update two.  Which is correct:

`void some_function_foo(const char *buffer, const int buffer_length);`
or
`void some_function_foo(const char *, const int );`

wolf99:
@Gamecock Good point. At a quick glance the kernel seems to use the style where names are specified. I'd be inclined to stick with that for clarity's sake. Thoughts?

Gamecock:
I think that's clearer and we should explicitly require the names be the
same in the .c and .h.

ryanplusplus:
👍

PR #219, opened 2017-10-18
Simplified the implementation of the raindrops exercise and its tests.

mindejulian:
When attempting to solve the raindrops exercise in C, I found the method signature unnecessarily complex, and hard to understand.
Therefore I have here changed it to an implementation I personally believe to be easier to understand and solve the same problem.

### DIFF
--- a/exercises/raindrops/src/example.c
+++ b/exercises/raindrops/src/example.c
@@ -1,11 +1,12 @@
 #include "raindrops.h"
 #include "stdio.h"
 #include "string.h"
+#include <stdlib.h>
 
 char *convert(int drops)
 {
-   static char result[100];
-   memset(result, '\0', 100);
+   // strlen("PlingPlangPlong") == 15, and max length of int is 12.
+   char *result = calloc(sizeof(char), 15);
 
    if (drops % 3 == 0) {
       strcat(result, "Pling");
--- a/exercises/raindrops/src/example.c
+++ b/exercises/raindrops/src/example.c
@@ -2,28 +2,24 @@
 #include "stdio.h"
 #include "string.h"
 
-#define PLING_FACTOR 3
-#define PLANG_FACTOR 5
-#define PLONG_FACTOR 7
-
 char *convert(int drops)
 {
-   static char result[100];
-   memset(result,'\0',100);
-  
-   if(drops % 3 == 0) {
+   static char     result[100];
+   memset(result, '\0', 100);
+
+   if (drops % 3 == 0) {
       strcat(result, "Pling");
    }
-   if(drops % 5 == 0) {
+   if (drops % 5 == 0) {
       strcat(result, "Plang");
    }
-   if(drops % 7 == 0) {
+   if (drops % 7 == 0) {
       strcat(result, "Plong");
    }
-   if(strlen(result) == 0) {
+   if (strlen(result) == 0) {
       char drops_string[12] = "";
       sprintf(drops_string, "%d", drops);
       strcat(result, drops_string);
    }
-   return(result);
+   return (result);
 }
--- a/exercises/raindrops/src/example.c
+++ b/exercises/raindrops/src/example.c
@@ -6,20 +6,24 @@
 #define PLANG_FACTOR 5
 #define PLONG_FACTOR 7
 
-char *convert(char *buffer, int buffer_length, int drops)
+char *convert(int drops)
 {
-   memset(buffer, '\0', sizeof(char) * buffer_length);
-   if ((drops % PLING_FACTOR != 0) &&
-       (drops % PLANG_FACTOR != 0) && (drops % PLONG_FACTOR != 0)) {
-      snprintf(buffer, buffer_length, "%d", drops);
-   } else {
-      snprintf
-          (buffer,
-           buffer_length,
-           "%s%s%s",
-           drops % PLING_FACTOR == 0 ? "Pling" : "",
-           drops % PLANG_FACTOR == 0 ? "Plang" : "",
-           drops % PLONG_FACTOR == 0 ? "Plong" : "");
+   static char result[100];
+   memset(result,'\0',100);
+  
+   if(drops % 3 == 0) {
+      strcat(result, "Pling");
    }
-   return buffer;
+   if(drops % 5 == 0) {
+      strcat(result, "Plang");
+   }
+   if(drops % 7 == 0) {
+      strcat(result, "Plong");
+   }
+   if(strlen(result) == 0) {
+      char drops_string[12] = "";
+      sprintf(drops_string, "%d", drops);
+      strcat(result, drops_string);
+   }
+   return(result);
 }
--- a/exercises/raindrops/src/example.c
+++ b/exercises/raindrops/src/example.c
@@ -19,5 +19,5 @@ char *convert(char result[], int drops)
       sprintf(drops_string, "%d", drops);
       strcat(result, drops_string);
    }
-   return (result);
+   return result;
 }
--- a/exercises/raindrops/src/example.c
+++ b/exercises/raindrops/src/example.c
@@ -3,11 +3,8 @@
 #include "string.h"
 #include <stdlib.h>
 
-char *convert(int drops)
+char *convert(char result[], int drops)
 {
-   // strlen("PlingPlangPlong") == 15, and max length of int is 12.
-   char *result = calloc(sizeof(char), 15);
-
    if (drops % 3 == 0) {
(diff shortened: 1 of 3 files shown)


ryanplusplus:
@mindejulian I like the drive towards simplifying this, but I think there's a fairly serious issue with your implementation: the string being returned is allocated on the stack so you're technically pointing to "freed" memory by returning a pointer to it.

I think we can simplify things a bit by getting rid of the buffer length, but I think the buffer does need to be provided by the client.

Gamecock:
if he malloc'd the buffer and freed it in the test would that repeat a pattern we have in a lot of the tests.

mindejulian:
Thanks so much for reviewing and your comments. I understand your points, and memory allocation is not something I am used to work with. 
I will submit an updated pr where the variable is freed in the test.

One question though, does this malloc/free approach only apply to arrays and larger variables, or should all variables be treated this way? Specifically, if the convert function returned an integer value, would my approach be ok?

exercises/raindrops/test/test_raindrops.c
```diff
@@ -11,99 +12,106 @@ void tearDown(void)
 {
 }
 
+void test_convert_drops(int drops, char *expected)
+{
+   char *result = convert(drops);
+   TEST_ASSERT_EQUAL_STRING(expected, result);
+   free(result);
```
wolf99: This free should be in the same function that allocates the memory. This means moving the allocation (`malloc()`, `calloc()` or otherwise) into the _test_ rather than the implementation. Or just remove duynamic memory allocation altogether. Something like the following psuedo-code

```c
    char result[MAX_STRING_LENGTH] = {0};
    convert(result, drops);
    TEST_ASSERT_EQUAL_STRING(expected, result);

File: exercises/raindrops/test/test_raindrops.c [hunk 2]
```diff
@@ -11,99 +12,106 @@ void tearDown(void)
 {
 }
 
+void test_convert_drops(int drops, char *expected)
```
wolf99: We are trying to change how helper functions are named in test files, it would be great if you could remove the 'test_' prefix here

exercises/raindrops/test/test_raindrops.c
```diff
@@ -11,99 +12,106 @@ void tearDown(void)
...
 {
-   TEST_ASSERT_EQUAL_STRING("1", convert(1));
+   test_convert_drops(1, "1");
 }
 
 void test_three_yields_pling(void)
 {
-   TEST_IGNORE();               // delete this line to run test
-   TEST_ASSERT_EQUAL_STRING("Pling", convert(3));
+   //TEST_IGNORE();               // delete this line to run test
```
wolf99: These need to remain _un_-commented in the test file here in the repository

exercises/raindrops/src/example.c
```diff
@@ -1,25 +1,26 @@
...
-   } else {
-      snprintf
-          (buffer,
-           buffer_length,
-           "%s%s%s",
-           drops % PLING_FACTOR == 0 ? "Pling" : "",
-           drops % PLANG_FACTOR == 0 ? "Plang" : "",
-           drops % PLONG_FACTOR == 0 ? "Plong" : "");
+   // strlen("PlingPlangPlong") == 15, and max length of int is 12.
+   char *result = calloc(sizeof(char), 15);
```
wolf99: The tests fail in Travis because the maximum length of the string is _16_ not 15. there are 15 letters, plus the string terminator (the `'\0'` character).

exercises/raindrops/test/test_raindrops.c
mindejulian: Sure, that sounds like a better solution.

exercises/raindrops/test/test_raindrops.c
mindejulian: 👍

exercises/raindrops/test/test_raindrops.c
mindejulian: 🤦‍♂️  👍

exercises/raindrops/test/test_raindrops.c
```diff
@@ -161,8 +134,6 @@ int main(void)
    RUN_TEST(test_fifty_two_yields_itself);
    RUN_TEST(test_one_hundred_five_yields_plingplangplong);
    RUN_TEST(test_big_prime_yields_itself);
-   RUN_TEST(test_no_buffer_overflow_for_plingplangplong);
-   RUN_TEST(test_no_buffer_overflow_for_number);
```
wolf99: Could probably add these two back in now 😃 . They help people learn to be careful with putting data into arrays.

exercises/raindrops/test/test_raindrops.c
```diff
@@ -11,134 +12,106 @@ void tearDown(void)
 {
 }
 
+void convert_drops(int drops, char *expected)
+{
+   char result[16] = { 0 };
```
wolf99: Might be worth replacing `16` with `#define BUFFER_LENGTH 16`

exercises/raindrops/test/test_raindrops.c
```diff
@@ -1,5 +1,6 @@
 #include "vendor/unity.h"
 #include "../src/raindrops.h"
+#include <stdlib.h>
 
 #define BUFFER_LENGTH 100
```
wolf99: Don't think this is used anymore? Value could be changed to `16` and used in array definition in `convert_drops()`

exercises/raindrops/test/test_raindrops.c
mindejulian: Hehe, good point. I will also fix my example so that it passes the tests you mention. 😊

exercises/raindrops/test/test_raindrops.c
wolf99: Can then change this to `char result[BUFFER_LENGTH] = { 0 };`

wolf99:
Couple more small little points there, but looking good overall!

Gamecock:
@mindejulian back to your earlier question.  You do not have to worry about the memory of a return value being freed when it is a basic type: char, int, uint64_t ... because C passes everything by value. 
https://stackoverflow.com/questions/2229498/passing-by-reference-in-c
Where you get tripped up is when you return a string.  C passes the value of the memory location of the first character in the string, which is why it's a `char *`.  
Arrays are weird, even if your signature if `doSomething(int foo[])` it gets interpreted as `doSomething(int* foo)`
https://stackoverflow.com/questions/6567742/passing-an-array-as-an-argument-to-a-function-in-c

Structures can be passed by value, or by a pointer.
https://stackoverflow.com/questions/30980759/when-should-i-pass-or-return-a-struct-by-value

exercises/raindrops/test/test_raindrops.c
Gamecock: I'm confused, how does `raindrops.c` know the size of the array `test_raindrops.c` instantiated.  If you give it too small a buffer it's going to overwrite it.  `sizeof(result)` will not work.  I think it's OK to use with maybe a comment in the header that minimum buffer size is 16. Since the maximum return size is known and the purpose of this exercise is to teach strings and mod.

mindejulian: I am also a bit confused here. In order to make the overflow tests pass, the `convert` function needs to get the size of `result` as an argument. And if I add the size as an argument the signature is back to what it was initially, and there is no point with this pr.
So, imho, either the lesson of buffer overflow should be postponed to a lesson further down the line, or this pr should be rejected. 😌

exercises/raindrops/test/test_raindrops.c
mindejulian: 👍

mindejulian:
Thank you very much for the explanation @Gamecock 😃

exercises/raindrops/test/test_raindrops.c
wolf99: Ah my bad, didn't fully think through how these tests work! They should indeed be left out for the implementation in this PR.

exercises/raindrops/test/test_raindrops.c
Gamecock: Might as well change buffer length to 16, since it cannot be any longer, might need those other 84 bytes someday.

Gamecock:
LGTM, we can wait on buffer overruns for a later issue.

ryanplusplus:
Looks good! Sorry for the delay (had a stomach bug go around my house).

PR #221, opened 2017-10-20
Add Collatz Conjecture exercise

RealBarrettBrown:
Okay, so it's been a year and I need help with `config.json`.  What's this `uuid`?

### DIFF
--- a/exercises/collatz-conjecture/src/example.c
+++ b/exercises/collatz-conjecture/src/example.c
@@ -1,0 +1,33 @@
+#include "collatz_conjecture.h"
+#include <stdbool.h>
+
+static bool is_even(int number)
+{
+    if(number % 2) {
+        return false;
+    }
+
+    return true;
+}
+
+int steps(int start)
+{
+    int step = 0;
+    int number = start;
+
+    if(start <= 0) {
+        return ERROR_VALUE;
+    }
+
+    while(number != 1) {
+        step++;
+
+        if(is_even(number)) {
+            number /= 2;
+        }
+        else {
+            number = (number * 3) + 1;
+        }
+    }
+    return step;
+}
--- a/exercises/collatz-conjecture/src/example.c
+++ b/exercises/collatz-conjecture/src/example.c
@@ -3,31 +3,30 @@
 
 static bool is_even(int number)
 {
-    if(number % 2) {
-        return false;
-    }
+   if (number % 2) {
+      return false;
+   }
 
-    return true;
+   return true;
 }
 
 int steps(int start)
 {
-    int step = 0;
-    int number = start;
+   int step = 0;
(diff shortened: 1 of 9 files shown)


patricksjackson:
I'm not sure exactly what the `uuid` is used for, but it needs to be a unique identifier for the exercise. It looks like you already used the `configlet` tool to generate the README (?) you can also use it to generate a `uuid` to add in as well.

If that makes no sense somebody can explain further. I don't think `configlet` is documented very much, if at all, so it's not exactly obvious what to do.

exercises/collatz-conjecture/test/test_collatz_conjecture.c
```diff
@@ -0,0 +1,51 @@
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
```
patricksjackson: Do you need all of these include files?

exercises/collatz-conjecture/src/example.c
```diff
@@ -0,0 +1,32 @@
...
+   int number = start;
+
+   if (start <= 0) {
+      return ERROR_VALUE;
+   }
+
+   while (number != 1) {
+      step++;
+
+      if (is_even(number)) {
```
patricksjackson: As a nitpick on the solution - it seems unnecessary to have a separate function for testing if a number is even.

ryanplusplus:
@RealBarrettBrown I've been using this: https://www.uuidgenerator.net/

exercises/collatz-conjecture/src/example.c
```diff
@@ -0,0 +1,32 @@
...
+#include <stdbool.h>
+
+static bool is_even(int number)
+{
+   if (number % 2) {
+      return false;
+   }
+
+   return true;
+}
```
ryanplusplus: `return `number % 2` == 0` (if this function even survives)

exercises/collatz-conjecture/test/test_collatz_conjecture.c
ryanplusplus: As far as I can tell, none are necessary

RealBarrettBrown: Nope, they're not necessary.  Copied from another exercise.

exercises/collatz-conjecture/src/example.c
RealBarrettBrown: I agree

ryanplusplus:
LGTM 👍

patricksjackson:
👍

patricksjackson:
I knew I missed something - this doesn't have the TEST_IGNORE() macros on the tests.

PR #222, opened 2017-10-24
Add buffer overrun tests to word_count

### LINKED ISSUE #207: Additional Test Cases Word Count
```
#define MAX_WORDS 20            // at most MAX_WORDS will be found in the test input string
#define MAX_WORD_LENGTH 50      // no individual word will exceed this length

// results structure
typedef struct WordCount_Word {
   char text[MAX_WORD_LENGTH];
   int count;
} WordCount_Word_t;
```

The test cases for word_count don't send more than 20 words or longer words.  The test cases also clears the structure before giving it to word_count.c.  I think we should write some test cases to encourage developers to be more defensive.   Thoughts.

Gamecock:
Added tests to make sure that people don't overrun buffers, since this has two arrays it should be a good exercise for it.
Closes #207

### DIFF
--- a/exercises/word-count/src/example.c
+++ b/exercises/word-count/src/example.c
@@ -20,7 +20,7 @@ static int word_match(char *testWord, int current_word_count,
    return match_index;
 }
 
-int word_count(char *input_text, word_count_word_t words[MAX_WORDS])
+int word_count(const char *input_text, word_count_word_t* words)
 {
    int index;
    int unique_words = 0;
--- a/exercises/word-count/src/example.c
+++ b/exercises/word-count/src/example.c
@@ -20,7 +20,7 @@ static int word_match(char *testWord, int current_word_count,
    return match_index;
 }
 
-int word_count(const char *input_text, word_count_word_t* words)
+int word_count(const char *input_text, word_count_word_t * words)
 {
    int index;
    int unique_words = 0;
--- a/exercises/word-count/src/example.c
+++ b/exercises/word-count/src/example.c
@@ -20,7 +20,7 @@ static int word_match(char *testWord, int current_word_count,
    return match_index;
 }
 
-int word_count(const char *input_text, word_count_word_t * words)
+int word_count(const char *input_text, word_count_word_t *words)
 {
    int index;
    int unique_words = 0;
--- a/exercises/word-count/src/example.c
+++ b/exercises/word-count/src/example.c
@@ -51,10 +51,14 @@ int word_count(char *input_text, word_count_word_t words[MAX_WORDS])
       // add if not yet counted else increment count.
       if (-1 == index) {
          // reject anything that would overrun buffers
-         if (MAX_WORD_LENGTH < strlen(test_word))
-            return EXCESSIVE_LENGTH_WORD;
-         if (MAX_WORDS == unique_words)
-            return EXCESSIVE_NUMBER_OF_WORDS;
+         if (MAX_WORD_LENGTH < strlen(test_word)) {
+            unique_words = EXCESSIVE_LENGTH_WORD;
+            break;
+         }
+         if (MAX_WORDS == unique_words) {
+            unique_words = EXCESSIVE_NUMBER_OF_WORDS;
+            break;
+         }
          words[unique_words].count = 1;
          strncpy(words[unique_words].text, test_word, MAX_WORD_LENGTH);
          unique_words++;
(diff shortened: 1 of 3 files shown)


exercises/word-count/src/example.c
```diff
@@ -36,7 +36,7 @@ int word_count(char *input_text, word_count_word_t words[MAX_WORDS])
    lower_case_input_text[index] = '\0';
 
    // start with known results...
-   memset(words, 0, sizeof(word_count_word_t));
+   memset(words, 0, sizeof(word_count_word_t) * MAX_WORDS);
```
Gamecock: Found this fixing the tests.

exercises/word-count/test/test_word_count.c
```diff
@@ -14,36 +14,31 @@ void tearDown(void)
...
+                                    MAX_WORD_LENGTH);
    }
 }
 
-static void reset_actual_solution(void)
-{
-   memset(actual_solution, 0, sizeof(actual_solution));
-}
-
 // Tests Start here
```
Gamecock: Took this out to check if the student did.

exercises/word-count/src/example.c
```diff
@@ -50,8 +50,13 @@ int word_count(char *input_text, word_count_word_t words[MAX_WORDS])
 
       // add if not yet counted else increment count.
       if (-1 == index) {
+         // reject anything that would overrun buffers
+         if (MAX_WORD_LENGTH < strlen(test_word))
+            return EXCESSIVE_LENGTH_WORD;
+         if (MAX_WORDS == unique_words)
+            return EXCESSIVE_NUMBER_OF_WORDS;
```
Gamecock: leak is right here lower_case_input_text does not get freed.  I'll update tomorrow.

exercises/word-count/src/example.c
```diff
@@ -50,8 +50,17 @@ int word_count(char *input_text, word_count_word_t words[MAX_WORDS])
...
       if (-1 == index) {
+         // reject anything that would overrun buffers
+         if (MAX_WORD_LENGTH < strlen(test_word)) {
+            unique_words = EXCESSIVE_LENGTH_WORD;
+            break;
+         }
+         if (MAX_WORDS == unique_words) {
+            unique_words = EXCESSIVE_NUMBER_OF_WORDS;
+            break;
+         }
```
Gamecock: Important lesson for me in why some coding standards require one exit from a function.  Needed to clean up after myself.

exercises/word-count/src/word_count.h
```diff
@@ -1,22 +1,26 @@
...
 // inputs:
-//    input_text =  string no longer than will take (MAX_WORDS * MAX_WORD_LENGTH)
-//    words = allocated structure to record the words fourn and their frequency
+//    input_text =  a null-terminated string containing that is analyzed
 //
 // outputs:
+//    words = allocated structure to record the words fourn and their frequency
 //    uniqueWords - number of words in the words structure
+//           returns a negative number if an error. words will contain the results up to that point.
 int word_count(char *input_text, word_count_word_t words[MAX_WORDS]);
```
Gamecock: would it be more idiomatic C to change this to `int word_count(char* input_text, word_count_word_t* words)` or is that more change than necessary?

ryanplusplus: I more often seen a pointer, but I can't say that I know whether it's more idiomatic.

The only real issue I see with this prototype is that the input is not marked `const`.

Gamecock:
I have both of those changes in my draft, should be ready for review this
weekend.

Mike

exercises/word-count/src/example.c
patricksjackson: Good catch, that's why if possible I usually just use Cpp and [RAII](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rr-raii). In C, a lot of code ends up using `goto exit;` to cleanup resources if the control flow is too complicated, which always looks messy to me.

Gamecock: I didn't catch it, valgrind did.

ryanplusplus:
Sorry @Gamecock, I'm not sure why that angered the formatter. It's inconsistent even on the same line so I figured it was allowing either, but there's no use fighting the formatter.

ryanplusplus:
LGTM

Gamecock:
If only I cared enough to dig into the indent flags and see why it cared.

wolf99:
I think this is just the default behaviour of indent. AFAICT there is no way of altering this with basic GNU indent.

Gamecock:
Is there something else left to do on this?

wolf99:
LGTM

PR #223, opened 2017-10-24
Add queen-attack exercise

RealBarrettBrown:
Problem specification: https://github.com/exercism/problem-specifications/tree/master/exercises/queen-attack

### DIFF
--- a/exercises/queen-attack/src/queen_attack.h
+++ b/exercises/queen-attack/src/queen_attack.h
@@ -1,0 +1,19 @@
+#ifndef QUEEN_ATTACK_H
+#define QUEEN_ATTACK_H
+
+#include <stdint.h>
+
+enum result {
+    CAN_NOT_ATTACK,
+    CAN_ATTACK,
+    INVALID_POSITION
+};
+
+typedef struct {
+    uint8_t row;
+    uint8_t column;
+} queen_position_t;
+
+uint8_t can_attack(queen_position_t queen_1, queen_position_t queen_2);
+
+#endif
--- a/exercises/queen-attack/src/queen_attack.h
+++ b/exercises/queen-attack/src/queen_attack.h
@@ -3,17 +3,17 @@
 
 #include <stdint.h>
 
-enum result {
+typedef enum {
    CAN_NOT_ATTACK,
    CAN_ATTACK,
    INVALID_POSITION
-};
+} attack_status_t;
 
 typedef struct {
    uint8_t row;
    uint8_t column;
-} queen_position_t;
+} position_t;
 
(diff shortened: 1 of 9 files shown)


exercises/queen-attack/src/queen_attack.h
```diff
@@ -0,0 +1,19 @@
...
+   CAN_ATTACK,
+   INVALID_POSITION
+};
+
+typedef struct {
+   uint8_t row;
+   uint8_t column;
+} queen_position_t;
+
+uint8_t can_attack(queen_position_t queen_1, queen_position_t queen_2);
```
patricksjackson: The return value here should be the enum type (which I'd suggest renaming). Maybe `attack_status`?

patricksjackson:
Oops, see #179, we need to set this up to ignore all but the first test by default.

wolf99:
Does the standard exercise README need to contain information on removing the TEST_IGNORE()'s?

exercises/queen-attack/src/queen_attack.h
RealBarrettBrown: I agree with both 👍

wolf99:
LGTM 👍

patricksjackson:
@wolf99 If there's a place in the guides where we give a list of instructions for adding an exercise, TEST_IGNORE() info should be there.

Or did you mean in the README's we give to students? I am not sure if we need it (it's already explained in comments in the test case file) but it may be good just to have a short description, something like: `By default only the first test is run. To check your solution against other tests cases, just delete the "TEST_IGNORE()" macro in the test you'd like to run.`

exercises/queen-attack/src/queen_attack.h
```diff
@@ -0,0 +1,19 @@
...
+   CAN_ATTACK,
+   INVALID_POSITION
+} attack_status_t;
+
+typedef struct {
+   uint8_t row;
+   uint8_t column;
+} position_t;
+
+attack_status_t can_attack(position_t queen_1, position_t queen_2);
```
patricksjackson: If we're switching to `white_/black_queen` you should also switch it here.

patricksjackson:
Looks good to me except for the nit on the queen variable names 👍

ryanplusplus:
LGTM

wolf99:
@patricksjackson I'd forgotten there should be a comment on the first `TEST_IGNORE()` occurrence. Looks like there isn't one for this, I can open another quick PR for that. Unless @RealBarrettBrown is able to add it to this PR? (see line 23 in https://github.com/exercism/c/blob/master/exercises/acronym/test/test_acronym.c for example)

PR #224, opened 2017-10-25
Add sublist exercise

RealBarrettBrown:
Problem specification: https://github.com/exercism/problem-specifications/tree/master/exercises/sublist

### DIFF
--- a/config.json
+++ b/config.json
@@ -442,6 +442,19 @@
       ],
       "unlocked_by": "atbash-cipher",
       "uuid": "55b87cc2-c3a0-438a-b2b6-116ad6356802"
+    },
+    {
+      "core": true,
+      "difficulty": 4,
+      "slug": "sublist",
+      "topics": [
+        "pointers",
+        "arrays",
+        "searching",
+        "control_flow_if_statements",
+        "control_flow_loops"
+      ],
+      "uuid": "928e9860-0df9-5180-8529-7024c991b3616adc21a"
     }
   ],
   "foregone": [],
--- a/config.json
+++ b/config.json
@@ -444,7 +444,7 @@
       "uuid": "55b87cc2-c3a0-438a-b2b6-116ad6356802"
     },
     {
-      "core": true,
+      "core": false,
       "difficulty": 4,
       "slug": "sublist",
       "topics": [
@@ -454,6 +454,7 @@
         "control_flow_if_statements",
         "control_flow_loops"
       ],
+      "unlocked_by": "palindrome-products",
       "uuid": "928e9860-0df9-5180-8529-7024c991b3616adc21a"
     },
     {
--- a/exercises/sublist/src/example.c
+++ b/exercises/sublist/src/example.c
@@ -2,56 +2,61 @@
 #include <stdbool.h>
 #include <string.h>
 
-static bool is_equal(int *listOne, int *listTwo, size_t listOneLength, size_t listTwoLength)
+static bool is_equal(int *listOne, int *listTwo, size_t listOneLength,
+                     size_t listTwoLength)
 {
-    return (listOneLength == listTwoLength && !memcmp(listOne, listTwo, listOneLength));
+   return (listOneLength == listTwoLength
+           && !memcmp(listOne, listTwo, listOneLength));
 }
 
-static bool is_sublist(int *listOne, int *listTwo, size_t listOneLength, size_t listTwoLength)
+static bool is_sublist(int *listOne, int *listTwo, size_t listOneLength,
+                       size_t listTwoLength)
 {
-    int i;
-    int loop_end = (listTwoLength - listOneLength) / sizeof(*listOne);
+   int i;
+   int loop_end = (listTwoLength - listOneLength) / sizeof(*listOne);
 
-    if (listOneLength > listTwoLength) {
-        return false;
-    }
+   if (listOneLength > listTwoLength) {
+      return false;
+   }
 
-    for (i = 0; i <= loop_end; i++) {
-        if (!memcmp(listOne, &listTwo[i], listOneLength)) {
-            return true;
-        }
-    }
+   for (i = 0; i <= loop_end; i++) {
+      if (!memcmp(listOne, &listTwo[i], listOneLength)) {
+         return true;
+      }
+   }
 
-    return false;
+   return false;
 }
 
-static bool is_superlist(int *listOne, int *listTwo, size_t listOneLength, size_t listTwoLength)
+static bool is_superlist(int *listOne, int *listTwo, size_t listOneLength,
+                         size_t listTwoLength)
 {
-    int i;
-    int loop_end = (listOneLength - listTwoLength) / sizeof(*listOne);
+   int i;
+   int loop_end = (listOneLength - listTwoLength) / sizeof(*listOne);
 
-    if (listTwoLength > listOneLength) {
-        return false;
-    }
+   if (listTwoLength > listOneLength) {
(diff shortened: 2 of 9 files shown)


exercises/sublist/src/example.c
```diff
@@ -0,0 +1,62 @@
+#include "sublist.h"
+#include <stdbool.h>
+#include <string.h>
+
+static bool is_equal(int *listOne, int *listTwo, size_t listOneLength,
+                     size_t listTwoLength)
+{
+   return (listOneLength == listTwoLength
+           && !memcmp(listOne, listTwo, listOneLength));
```
patricksjackson: Passing null pointers to memcmp works to an extent, but it's undefined.

exercises/sublist/src/example.c
```diff
@@ -0,0 +1,62 @@
...
+}
+
+int check_lists(int *listOne, int *listTwo, size_t listOneLength,
+                size_t listTwoLength)
+{
+   if (is_equal(listOne, listTwo, listOneLength, listTwoLength)) {
+      return EQUAL;
+   } else if (is_sublist(listOne, listTwo, listOneLength, listTwoLength)) {
+      return SUBLIST;
+   } else if (is_superlist(listOne, listTwo, listOneLength, listTwoLength)) {
```
patricksjackson: `is_superlist(One, Two, OneLength, TwoLength)` = `is_sublist(Two, One, TwoLength, OneLength)`

exercises/sublist/src/sublist.h
```diff
@@ -0,0 +1,16 @@
...
+#include <stddef.h>
+
+enum {
+   EQUAL,
+   UNEQUAL,
+   SUBLIST,
+   SUPERLIST
+};
+
+int check_lists(int *listOne, int *listTwo, size_t listOneLength,
```
patricksjackson: The return value should be the enum type.

exercises/sublist/test/test_sublist.c
```diff
@@ -0,0 +1,198 @@
...
+void test_empty_lists(void)
+{
+   TEST_ASSERT_EQUAL(EQUAL, check_lists(NULL, NULL, 0, 0));
+}
+
+void test_empty_list_within_non_empty_list(void)
+{
+   int listTwo[] = { 1, 2, 3 };
+
+   TEST_ASSERT_EQUAL(SUBLIST, check_lists(NULL, listTwo, 0, sizeof(listTwo)));
```
patricksjackson: Passing the length argument as the size in bytes is pretty weird, and I don't think we do it in other array exercises. The length argument should instead be the number of elements.

patricksjackson:
Also remember that this needs the TEST_IGNORE() things too. #179

exercises/sublist/src/example.c
```diff
@@ -0,0 +1,62 @@
+#include "sublist.h"
+#include <stdbool.h>
+#include <string.h>
+
+static bool is_equal(int *listOne, int *listTwo, size_t listOneLength,
```
wolf99: Parameter names should also be snake_case rather than camelCase 🐫

exercises/sublist/src/example.c
RealBarrettBrown: I think a more explicit check for null is in order then.

exercises/sublist/src/example.c
RealBarrettBrown: Yes, yes it does 😄

RealBarrettBrown: I think keeping the helper `is_superlist` and just having it call `is_sublist` with reversed params is still clearer though.  Thoughts?

exercises/sublist/src/sublist.h
RealBarrettBrown: 👍

exercises/sublist/test/test_sublist.c
RealBarrettBrown: Yeah, I initially had it passing the number of elements and then I somehow convinced myself that it was weird, so I changed it.  But, now I agree that it was right the first time.

exercises/sublist/src/example.c
```diff
@@ -0,0 +1,62 @@
...
+
+comparison_result_t check_lists(int *list_one, int *list_two,
+                                size_t list_one_element_count,
+                                size_t list_two_element_count)
+{
+   if (is_equal
+       (list_one, list_two, list_one_element_count, list_two_element_count)) {
+      return EQUAL;
+   } else
+       if (is_sublist
```
ryanplusplus: Did `indent` put `else` and `if` on two lines?

exercises/sublist/src/sublist.h
```diff
@@ -0,0 +1,16 @@
...
+
+typedef enum {
+   EQUAL,
+   UNEQUAL,
+   SUBLIST,
+   SUPERLIST
+} comparison_result_t;
+
+comparison_result_t check_lists(int *listOne, int *listTwo,
+                                size_t listOneLength, size_t listTwoLength);
```
ryanplusplus: Make sure these variable names are lower_snake_case

ryanplusplus:
I don't have a great suggestion for new names, but the current names of the arguments makes it difficult for me to tell which list is being checked against which. Is list 1 being checked against list 2 or vice-versa? It would be great if we could find names that make that clear without having to infer it from the tests.

exercises/sublist/src/example.c
RealBarrettBrown: Yeah, it definitely did.  But now it doesn't seem to want to do it anymore.

RealBarrettBrown:
`base_list` & `list_to_compare`?

ryanplusplus:
LGTM

config.json
```diff
@@ -444,6 +444,19 @@
...
+      "difficulty": 4,
+      "slug": "sublist",
+      "topics": [
+        "pointers",
+        "arrays",
+        "searching",
+        "control_flow_if_statements",
+        "control_flow_loops"
+      ],
+      "uuid": "928e9860-0df9-5180-8529-7024c991b3616adc21a"
```
Gamecock: does this need an "unlocked_by" field?

ryanplusplus: Let's go with non-core. They want to have very few core exercises (I think 6-8?) so better to not burn them without a good reason.

exercises/sublist/src/example.c
```diff
@@ -0,0 +1,63 @@
...
+          (list_to_compare, &base_list[i],
+           list_to_compare_element_count * sizeof(int))) {
+         return true;
+      }
+   }
+
+   return false;
+}
+
+static bool is_superlist(int *list_to_compare, int *base_list,
```
patricksjackson: Is this function really necessary? I assumed that the calls to this would be replaced by calls to is_sublist, not that this function would be implemented as is_sublist.

patricksjackson:
This looks fine to me except for setting it to a non-core exercise (I think that needs a change to the config.json)

config.json
RealBarrettBrown: Ok, `palindrome-products` seems like a reasonable exercise to unlock this one, in my opinion

wolf99: Yes, `palindrome-products` seems like a good choice to unlock this IMO

wolf99:
This PR is blocking #227 and #226, it would be great to get those closed 😃
@RealBarrettBrown are you OK to make the change to the `config.json`?

RealBarrettBrown:
Oops... I made the change a few days ago but forgot to commit.

PR #225, opened 2017-10-26
Add a contributing guide

wolf99:
Start a contributing guide by collating information from the README and links to other relevant information. This allows us to extend the information beyond the scope of the README. Item in point, the CI information.

speaking of which, I don't actually know a huge amount about what Travis does (other than run the tests and indent check) so maybe someone else could contribute on that point in a more concise manner than I?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,7 @@ To get an idea of the general status in relation to the launch of the track, che
 
 ## Contributing
 
-Contributions are very welcome! Please see the [contributing guide](https://github.com/exercism/c/blob/master/docs/CONTRIBUTING.md) to get started.
-
-## Coding Style
-
-Please see the [style guide](https://github.com/exercism/c/blob/master/docs/C_STYLE_GUIDE.md) for information on style and the [contributing guide](https://github.com/exercism/c/blob/master/docs/CONTRIBUTING.md#coding-style) for information on installing and using `indent` when making changes to exercises.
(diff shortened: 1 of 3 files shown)


wolf99:
BTW, the link to the `install indent guide` is a dud for now as that has not been written yet. I intend to add a PR for that soon-ish, but if anyone else has experience with platforms other than Ubuntu for this then feel free to have at it yourself 😄

patricksjackson:
I have not closely read the whole thing, but a couple quick notes:

Travis can either be run locally or you can set it up on your own fork. (The latter may be easier for most people) Or you can just open a PR and mark it as [WIP] if that's either. I would also mention the `bin\run-tests` more prominently (I only saw it at the very bottom) as that's the simplest way to test.

As an example [here](https://github.com/fish-shell/fish-shell/blob/master/CONTRIBUTING.md#travis-ci-build-and-test) is a quick explanation of how to setup Travis that should be sufficient.

I would not be too concerned about talking about how to install indent, it probably doesn't need a different file. Just tell Mac folks they can get it from Homebrew, and Linux folks that it's probably available from their package manager. You can safely ignore Windows.

wolf99:
thanks for the info @patricksjackson I've made the changes for the indent install information as mentioned. The CI stuff is a bit more lengthy so I'll add that when I next have time.

ryanplusplus:
@wolf99 I think that the anybody who is really interested in running Travis locally will likely dig that information up on their own. I think we should simply mention the scripts that are run by Travis and note that they can be run on your own machine before opening a pull request. Any more than that and I think you run the risk of scaring people off.

ryanplusplus:
This is still marked WIP, but it looks good to me.

wolf99:
@ryanplusplus My bad, forgot I had `[WIP]` left on it! Think it's good to go.

PR #227, opened 2017-10-27
Change UUIDs to "longer" version generated by configlet

### LINKED ISSUE #226: Configlet lint errors out due to UUID
This is kind of odd, but running `./configlet lint .` locally gives errors about every exercise UUID saying 

> `The following UUID was found in multiple Exercism tracks. Each exercise UUID must be unique across tracks.`

`configlet uuid` returns a longer string that the ones currently in `config.json`.  I don't claim to know anything about UUIDs or what they're used for in this context.  I do know, however, that replacing the UUIDs in `config.json` with the longer versions generated by configlet causes the errors to go away.

RealBarrettBrown:
Fixes #226 

I'll update this once #223 and #224 get merged, but I gotta open this now so I can get my shirt!

### DIFF
--- a/config.json
+++ b/config.json
@@ -11,7 +11,7 @@
         "optional_values",
         "text_formatting"
       ],
-      "uuid": "a0278e53-728d-4c04-9a0e-d6f7fe7ce98f"
+      "uuid": "adf43ee2-0730-7980-b2d9-931c91510d78a9dbd19"
     },
     {
       "core": true,
@@ -23,7 +23,7 @@
         "control_flow_loops",
         "strings"
       ],
-      "uuid": "7167e43c-04bd-4c6c-a446-a37384d42626"
+      "uuid": "c0bf4029-06d6-db80-4447-464a1248f1ed63b8964"
     },
     {
       "core": false,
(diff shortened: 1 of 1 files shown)


ryanplusplus:
LGTM

patricksjackson:
Does anybody know if changing the UUIDs breaks something centrally? I don't know enough about how they're used.

We should check on that before merging.

ryanplusplus:
@patricksjackson the UUIDs are for nextercism/exercism v2 which is only live as a preview right now. So no harm in changing the UUIDs right now.

patricksjackson:
Cool, that works then :+1:

config.json
```diff
@@ -463,7 +463,8 @@
         "functions",
         "mathematics"
       ],
-      "uuid": "1f68f837-6432-4f9e-afae-0cb17f8e0153"
+      "unlocked_by": "triangle",
```
RealBarrettBrown: This was missing an `unlocked_by`.  If anyone has a better exercise to unlock this one, let me know.

ryanplusplus: This looks good to me

wolf99: `triangle` looks good, but maybe `pascals-triangle` might be better (which also unlocks `triangle`). Going by some of the other tracks it seems to be the case that a lot only use a single level of unlock depth. i.e. one core exercise unlocks multiple non-core exercises rather than one core exercise unlocks one non-core which unlocks another non-core, etc.

ryanplusplus: Oh, good catch. Yes, I think that although it is possible to use more than one level of unlocking, it's not recommended.

PR #228, opened 2017-10-30
Modified anagram to not require any memory management

### LINKED ISSUE #183: Anagram: Memory management
The current implementation of the exercise, does require the student to use `xalloc`, but the allocated memory is then `free`d in the testsuite.

This makes the student unaware of the problems a `xalloc` can bring with it. I'd suggest therefore to change the test in a way that memory is allocated **and** freed by the student, by requiring him to have “constructors”, “destructors”, and “accessors” for the `Vector`.

If this is considered beyond the scope of this exercise, we should provide those functions and remove the label “memory management".

Gamecock:
Took more changes than I thought to require no memory management out of anagram.  Closes #183

### DIFF
--- a/exercises/anagram/src/anagram.h
+++ b/exercises/anagram/src/anagram.h
@@ -1,6 +1,8 @@
 #ifndef ANAGRAM_H
 #define ANAGRAM_H
 
+#define MAX_STR_LEN 20
+
 enum anagram_status {
    UNCHECKED = -1,
    NOT_ANAGRAM,
@@ -9,7 +11,7 @@ enum anagram_status {
 
 struct candidate {
    enum anagram_status is_anagram;
-   char *candidate;
+   const char * candidate;
 };
 
 struct candidates {
@@ -21,6 +23,6 @@ struct candidates {
  * @description - determines if any of the words in candidate are anagrams
  *                for word.  Word buffer and candidate structures may be modified.
  */
-void anagrams_for(char word[], struct candidates *candidates);
+void anagrams_for(const char * word, struct candidates *candidates);
 
 #endif
--- a/exercises/anagram/src/anagram.h
+++ b/exercises/anagram/src/anagram.h
@@ -1,13 +1,33 @@
 #ifndef ANAGRAM_H
 #define ANAGRAM_H
 
-#define MAX_STR_LEN 20
-
-struct vector {
-   char (*vec)[MAX_STR_LEN];
-   int size;
+enum anagramStatus {
+   UNCHECKED = -1,
+   NOT_ANAGRAM,
+   IS_ANAGRAM
 };
 
-struct vector anagrams_for(char *, struct vector);
+struct candidate {
+   enum anagramStatus isAnagram;
+   char *candidate;
+};
+
+struct candidates {
+   struct candidate *candidate;
+   int size;
+};
+/**
+ * @description - determines if any of the words in candidate are anagrams
+ *                for word.  Word buffer and candidate structures may be modified.
+ *
+ * @param - word, buffer containing a word used to look for an anagram in  candidates.
+ *          may be modified before function returns.
+ *
+ * @param - candidates, list of potential anagrams as an array of struct candidate.
+ *          Words in candidate structure may be modifies before function returns.
+ *          candidate.isAnagram will be set to IS_ANAGRAM if the candidate is an
+ *          anagram for word.
+ */
+void anagrams_for(char word[], struct candidates *candidates);
 
 #endif
--- a/exercises/anagram/src/anagram.h
+++ b/exercises/anagram/src/anagram.h
@@ -1,32 +1,25 @@
 #ifndef ANAGRAM_H
 #define ANAGRAM_H
 
-enum anagramStatus {
+enum anagram_status {
    UNCHECKED = -1,
    NOT_ANAGRAM,
    IS_ANAGRAM
 };
 
 struct candidate {
-   enum anagramStatus isAnagram;
+   enum anagram_status is_anagram;
    char *candidate;
 };
 
 struct candidates {
    struct candidate *candidate;
-   int size;
+   size_t count;
 };
+
 /**
  * @description - determines if any of the words in candidate are anagrams
  *                for word.  Word buffer and candidate structures may be modified.
- *
- * @param - word, buffer containing a word used to look for an anagram in  candidates.
- *          may be modified before function returns.
- *
- * @param - candidates, list of potential anagrams as an array of struct candidate.
- *          Words in candidate structure may be modifies before function returns.
- *          candidate.isAnagram will be set to IS_ANAGRAM if the candidate is an
- *          anagram for word.
  */
 void anagrams_for(char word[], struct candidates *candidates);
 
--- a/exercises/anagram/src/anagram.h
+++ b/exercises/anagram/src/anagram.h
@@ -11,7 +11,7 @@ enum anagram_status {
 
 struct candidate {
    enum anagram_status is_anagram;
-   const char * candidate;
+   const char *candidate;
 };
 
 struct candidates {
@@ -23,6 +23,6 @@ struct candidates {
  * @description - determines if any of the words in candidate are anagrams
  *                for word.  Word buffer and candidate structures may be modified.
  */
-void anagrams_for(const char * word, struct candidates *candidates);
+void anagrams_for(const char *word, struct candidates *candidates);
 
 #endif
--- a/exercises/anagram/src/example.c
+++ b/exercises/anagram/src/example.c
@@ -19,21 +19,25 @@ static int compare(const void *a, const void *b)
    return *(const char *)a - *(const char *)b;
 }
 
-void anagrams_for(const char * word, struct candidates *candidates)
+void anagrams_for(const char *word, struct candidates *candidates)
 {
-   to_lower_case(word);
+   char lower_word[MAX_STR_LEN] = { 0 };
+   strncpy(&lower_word[0], word, MAX_STR_LEN);
+   to_lower_case(&lower_word[0]);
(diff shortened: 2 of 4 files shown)


exercises/anagram/src/anagram.h
```diff
@@ -1,13 +1,33 @@
...
+   enum anagramStatus isAnagram;
+   char *candidate;
 };
 
-struct vector anagrams_for(char *, struct vector);
+struct candidates {
+   struct candidate *candidate;
+   int size;
+};
+/**
```
ryanplusplus: Can you add whitespace before this comment block?

exercises/anagram/src/anagram.h
```diff
@@ -1,13 +1,33 @@
...
+ *                for word.  Word buffer and candidate structures may be modified.
+ *
+ * @param - word, buffer containing a word used to look for an anagram in  candidates.
+ *          may be modified before function returns.
+ *
+ * @param - candidates, list of potential anagrams as an array of struct candidate.
+ *          Words in candidate structure may be modifies before function returns.
+ *          candidate.isAnagram will be set to IS_ANAGRAM if the candidate is an
+ *          anagram for word.
+ */
```
ryanplusplus: Actually, I don't think we've really done these big comment blocks before. Does anybody have any thoughts on this? I'm tempted to say that the current signature is pretty clear and that this could be omitted in the name of consistency.

exercises/anagram/src/anagram.h
```diff
@@ -1,13 +1,33 @@
...
-   int size;
+struct candidate {
+   enum anagramStatus isAnagram;
+   char *candidate;
 };
 
-struct vector anagrams_for(char *, struct vector);
+struct candidates {
+   struct candidate *candidate;
+   int size;
```
ryanplusplus: I find that `count` is usually more clear because `size` could be in bytes or in number of elements.

exercises/anagram/src/anagram.h
```diff
@@ -1,13 +1,33 @@
...
+   UNCHECKED = -1,
+   NOT_ANAGRAM,
+   IS_ANAGRAM
+};
 
-struct vector {
-   char (*vec)[MAX_STR_LEN];
-   int size;
+struct candidate {
+   enum anagramStatus isAnagram;
```
ryanplusplus: `isAnagram` => `is_anagram`

exercises/anagram/src/anagram.h
ryanplusplus: I'd also go with `size_t` or some other unsigned type since you probably aren't going to want to specify that you provided -4 candidates.

exercises/anagram/src/example.c
```diff
@@ -5,56 +5,39 @@
 
 #include "anagram.h"
 
-char *lower_case(char *p)
+void to_lower_case(char p[])
```
ryanplusplus: This can be `static`, right?

exercises/anagram/src/example.c
```diff
@@ -5,56 +5,39 @@
...
-      char *input_lower = lower_case((char *)vecp);
-      if (strcmp(input_lower, lower) != 0) {
-         char *input_sorted = input_lower;
-         qsort(input_sorted, strlen(input_sorted), 1, compare);
-         if (strcmp(sorted, input_sorted) == 0) {
-            strcpy(*vout_vecp, *vecp);
-            vout_vecp++;
-            vout.size++;
+   to_lower_case(word);
+   for (int x = 0; x < candidates->size; x++) {
```
ryanplusplus: How about using `i` here and in the function below? It's a bit more idiomatic.

exercises/anagram/src/anagram.h
wolf99: Also `anagramStatus` => `anagram_status`

exercises/anagram/src/anagram.h
wolf99: Between the signature and the tests the usage is pretty clear IMO.

exercises/anagram/src/example.c
wolf99: While this is being changed anyway, `str` might be more idiomatic than `p` for an argument to a function that does basic operations on a string (see string.h for example)

exercises/anagram/src/example.c
```diff
@@ -5,56 +5,39 @@
...
+   char *tmp = p;
+   while (*tmp) {
+      *tmp = tolower(*tmp);
+      tmp++;
    }
-
-   return lower;
 }
 
 int compare(const void *a, const void *b)
```
wolf99: Could also become `static`

exercises/anagram/test/test_anagram.c
```diff
@@ -18,24 +14,26 @@ void tearDown(void)
 {
 }
 
-static int stringcmp(const void *a, const void *b)
+static struct candidates buildCandidates(char *inputs, size_t size)
```
wolf99: `buildCandidates` => `build_candidates`

wolf99:
FYI, think you may have missed the `buildCanditates()` => `build_candidates()` on that last commit 659bfbb, good work otherwise 👍

Gamecock:
Damm Camels

Gamecock:
@wolf99 I think the goal was to have the student to not need to use dynamic memory.  Then you have to search for the word before sorting.  There is a specific test for that.  Same reason word is a char[] instead of const char *.

wolf99:
@Gamecock Ah, I see. Other option is to do as follows, but there's not much difference in it at that point! 😄 

```c
void anagrams_for(char word[], struct candidates *candidates)
{
   to_lower_case(word);

   size_t word_length = strlen(word);
   char unsorted_word[word_length + 1];
   strcpy(unsorted_word, word);

   qsort(word, word_length, 1, compare);

   for (int i = 0; i < (int)candidates->count; ++i) {
      to_lower_case(candidates->candidate[i].candidate);
      if (strcmp(unsorted_word, candidates->candidate[i].candidate) == 0) {
         candidates->candidate[i].is_anagram = NOT_ANAGRAM;
      } else {
         qsort(candidates->candidate[i].candidate,
               strlen(candidates->candidate[i].candidate), 1, compare);
         if (strcmp(word, candidates->candidate[i].candidate) == 0) {
            candidates->candidate[i].is_anagram = IS_ANAGRAM;
         } else {
            candidates->candidate[i].is_anagram = NOT_ANAGRAM;
         }
      }
   }
}
```

Gamecock:
@toby, I like this.  Might be the first time someone sees a C array length defined at runtime.  I've only seen them defined at compile time.  I'll make the change.

wolf99:
@Gamecock Cool 😎. TBH, I rarely think of VLA's myself even though they've been part of C since C99.

ryanplusplus:
@wolf99 VLA is rarely supported in embedded toolchains that otherwise support C99 so the C11 change to optional support may be intended to allow those sorts of toolchains to actually meet the standard. I haven't seen any desktop compilers forego VLA support, but I wouldn't say I've done an exhaustive search.

ryanplusplus:
I'm not sure what the best way to achieve this is, but it would be nice if we didn't have to modify the candidates in place when checking them. Right now it's kind of odd that the result no longer maintains the actual word.

My first thought is to just allocate two VLAs and re-use the second one for the candidates.

ryanplusplus:
Also, with respect to VLAs, there's a pretty simple workaround for anybody with a compiler that doesn't support them: just allocate fixed-size buffers of a sufficient length. Given this, I think it's fine to keep using them in the sample solution and to keep the function signature as-is.

wolf99:
Good point @ryanplusplus, although the second probably doesn't need to be a VLA. Come to think of it, neither does `word`... the max word length is fixed by `MAX_STR_LEN`, right? 

So just make both of these temporary arrays `MAX_STR_LEN` length and make sure to clear the candidate one between candidates, or at least that each candidate is properly NULL terminated when copied into it and it should be fine.

Gamecock:
I'm confused, the goal was to make it so the student did not have to think about memory management.  As it is now they can do that 1 of 3 ways.
1) Sort in place.
2) Use VLAs
3) Malloc buffers to sort as needed

I can write any of these solutions, does it matter which one.

wolf99:
```c
void anagrams_for(char word[], struct candidates *candidates)
{
   to_lower_case(word);

   char unsorted_word[MAX_STR_LEN] = {0};
   strcpy(unsorted_word, word);

   qsort(word, strlen(word), 1, compare);

   for (int i = 0; i < (int)candidates->count; ++i) {
      char current_candidate[MAX_STR_LEN] = {0};
      strcpy(current_candidate, candidates->candidate[i].candidate);
      to_lower_case(current_candidate);
      if (strcmp(unsorted_word, current_candidate) == 0) {
         candidates->candidate[i].is_anagram = NOT_ANAGRAM;
      } else {
         qsort(current_candidate, strlen(current_candidate), 1, compare);
         if (strcmp(word, current_candidate) == 0) {
            candidates->candidate[i].is_anagram = IS_ANAGRAM;
         } else {
            candidates->candidate[i].is_anagram = NOT_ANAGRAM;
         }
      }
   }
}
```

wolf99:
Sorry @Gamecock! Kind of exploring options as I write 😳 . The above uses regular plain old fixed-size arrays and a single loop. I haven't tested it though.

ryanplusplus:
@Gamecock what I was trying to get at (but didn't actually say) is that each candidate should be a pointer to const data (that is, read only). This would affect your solution as well as constrain all solutions to not modify the candidates in the process of checking them.

Gamecock:
Got it.  I cam update to meet that constraint.

Gamecock:
Is this the signature you want for the header?

ryanplusplus:
@Gamecock yep, thanks

wolf99:
LGTM

PR #232, opened 2017-11-02
Inline helper function test_phone_number

Gamecock:
Another instance where a helper function had me confused on which one was the first test.  The helper function was so simple I just inlined.  Net increase in 4 LOC, but IMHO improved readability.


ryanplusplus:
LGTM 👍

wolf99:
Looks good. Think this is possibly the first occasion of use of the `tearDown()`, well spotted 👍

PR #234, opened 2017-11-02
Triangle: change parameter from pointer to a struct

wolf99:
The triangle exercise uses a pointer to a struct as the parameter to the functions that are to be implemented.

AFAICT there is no need for the parameter to be a pointer rather than the struct itself. As the exercise relates to structs rather than pointers, it seems an easy change to pass the structs by value to make it simpler to implement.

### DIFF
--- a/exercises/triangle/src/example.c
+++ b/exercises/triangle/src/example.c
@@ -1,30 +1,30 @@
 #include "triangle.h"
 
-bool triangle_equality(triangle_t * input)
+bool triangle_equality(triangle_t input)
 {
-   return (input->a < (input->b + input->c)
-           && input->b < (input->a + input->c)
-           && input->c < (input->a + input->b)
+   return (input.a < (input.b + input.c)
+           && input.b < (input.a + input.c)
+           && input.c < (input.a + input.b)
        );
 }
 
-bool is_equilateral(triangle_t * input)
+bool is_equilateral(triangle_t input)
 {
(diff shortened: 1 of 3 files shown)


ryanplusplus:
In general, the advantage to doing this is that a pointer to a struct is generally smaller than the struct itself and so there is a stack usage (more generally, a RAM usage) advantage to passing a struct by pointer instead of value. However, I don't have strong feelings about this in the context of this exercise so I'm happy with this change unless someone else has strong feelings. This may even be good to give people experience passing a struct by value because, believe it or not, I've encountered C developers who were not aware that structs could be passed by value or assigned by value.

If we decide not to pass by value, we should at least mark the pointer as a pointer to const to indicate that the implementation only needs the value and is not making any changes to the contents of the struct.

Gamecock:
LGTM, the memory savings is trivial, probably not worth the indirection.

wolf99:
Good point @ryanplusplus. I pass structs by value rarely enough that I had to double check if I had the syntax correct. On the flip side though it might be easier for learners to grasp passing a struct by value, just the same as any other value type, than to attempt struct pointer usage for an exercise where the problem description doesn't seem to indicate any need for such.

Doing a quick search I realize there are exercises that use lots of variants of struct type parameters (struct, struct pointer, arrays of structs, etc) so either way is good 😃

ryanplusplus:
Okay, let's go with this then

PR #237, opened 2017-11-04
Add default maintainers file

wolf99:
Seems we missed out on https://github.com/exercism/meta/issues/20.
I'm unsure who else is (or was, alumnus are included too) actually officially a maintainer for the C track.

The issue mentions that maintainers should each maintain their own entry so I'll submit this one and the relevant people can add/subtract entries and details as they like.

Documentation on it is here https://github.com/exercism/docs/blob/master/maintaining-a-track/maintainer-configuration.md

### DIFF
--- a/config/maintainers.json
+++ b/config/maintainers.json
@@ -1,0 +1,15 @@
+{
+  "maintainers": [
+    {
+      "github_username": "ryanplusplus",
+      "show_on_website": false,
+      "alumnus": false,
+      "name": null,
+      "bio": null,
+      "link_text": null,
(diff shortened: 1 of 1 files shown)


kytrinyx:
Current maintainer list:

- @ryanplusplus 
- @wolf99 
- @patricksjackson 
- @RealBarrettBrown 

If anyone wants to be moved to "alumni", now's the time to :-)

ryanplusplus:
Looks good. If anybody wants to be moved to alumni, let's do that as follow-on work.

PR #238, opened 2017-11-04
Adding memory safety to phone_number tests.

patricksjackson:
Because blindly freeing a global pointer just screams scary to me. It's not really in this case, but having a dangling `free()` with no check and no zero'ing of the pointer might not be something we want the users to see.

### DIFF
--- a/exercises/phone-number/test/test_phone_number.c
+++ b/exercises/phone-number/test/test_phone_number.c
@@ -2,7 +2,7 @@
 #include "../src/phone_number.h"
 #include <stdlib.h>
 
-char *result;
+char *result = NULL;
 
 void setUp(void)
 {
@@ -10,7 +10,10 @@ void setUp(void)
 
 void tearDown(void)
 {
-   free(result);
+   if (result) {
+      free(result);
+   }
+   result = NULL;
 }
 
 void test_cleans_parens_dashes_and_spaces_from_the_number(void)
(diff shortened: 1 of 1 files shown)


exercises/phone-number/test/test_phone_number.c
```diff
@@ -2,15 +2,18 @@
 #include "../src/phone_number.h"
 #include <stdlib.h>
 
-char *result;
+char *result = NULL;
```
ryanplusplus: How about making this guy `static`?

ryanplusplus:
Yes, `free`ing `NULL` is safe. @patricksjackson is making the pointer `NULL` sufficient to address your concerns?

exercises/phone-number/test/test_phone_number.c
Gamecock: Do we care that this would mask the compiler error for uninitilized variables?

Gamecock: By this I meant the = NULL and not the static.

patricksjackson: So the addition of the `= NULL` actually does nothing, since globals are always zero initialized. I usually write it myself just to be clear. I don't *think* either way causes a compiler error though.

It does need to be static to match our conventions though, you're right @ryanplusplus

patricksjackson:
Wow, I had no idea `free`ing NULL was safe. I guess that makes sense though...

Clearing the pointer after freeing it though is probably still a good idea though.

wolf99:
LGTM

Issue #169, opened 2017-09-15
Possible memory leak not detected


Gamecock:
This issue was raised by [byelipk](http://exercism.io/byelipk) in his [exercise](http://exercism.io/submissions/22984132b96343c98e4f7e7f8b2cbbb3).  I don't think it applies in that case, but ran again using malloc and no error, I'm not used to Unity yet, but I think there should be one for a malloc without free.   Creating the issue until I can dig deeper.

Code I ran to test:
```
#include "raindrops.h"
#include "stdio.h"
#include "string.h"
#include <stdlib.h>

#define PLING_FACTOR 3
#define PLANG_FACTOR 5
#define PLONG_FACTOR 7

char *convert(char *buffer, int buffer_length, int drops)
{
   memset(buffer, '\0', sizeof(char) * buffer_length);
   if ((drops % PLING_FACTOR != 0) &&
       (drops % PLANG_FACTOR != 0) && (drops % PLONG_FACTOR != 0)) {
      snprintf(buffer, buffer_length, "%d", drops);
   } else {
      snprintf
          (buffer,
           buffer_length,
           "%s%s%s",
           drops % PLING_FACTOR == 0 ? "Pling" : "",
           drops % PLANG_FACTOR == 0 ? "Plang" : "",
           drops % PLONG_FACTOR == 0 ? "Plong" : "");
   }
    char* newbuffer = malloc(sizeof(char) * buffer_length);
    memcpy(newbuffer,buffer,buffer_length);
   return newbuffer;
}
```

Completely unrelated to the title, but raindrops_test.c needs to be updated to check the value in the buffer, not just the return value.  They should match.  I can address that at the same time or with a different patch.  However you want.

patricksjackson:
The Unity test suite doesn't check for memory leaks, so there shouldn't be any warning about that.

In this repo we run the tests twice, the second time using `valgrind` to check for memory leaks. But that only has an effect on code checked into this repo and not on the user's solutions.

ryanplusplus:
Right, Unity doesn't support this. CppUTest _does_ support it, but breaks frequently enough with updated toolchains that I would rather not go this route.

@patricksjackson what do you think about possibly making `valgrind` an optional install that our makefiles can use if it is available on the user's system?

patricksjackson:
I think that's a pretty good idea :) Were you thinking about adding it as a separate target in the makefile or adding it to `make test` if it's on the system?

[...]

PR #241, opened 2017-11-10
hello-world: Add memcheck make target

wolf99:
per #169

### DIFF
--- a/exercises/hello-world/makefile
+++ b/exercises/hello-world/makefile
@@ -15,7 +15,7 @@ test: tests.out
 memcheck: tests.out
 	@echo "Running memory check"
 	@valgrind $(VFLAGS) ./tests.out >> /dev/null
-	@echo "Memory check complete"
+	@echo "Memory check passed"
 
 clean:
 	rm -f *.o *.out
--- a/exercises/hello-world/makefile
+++ b/exercises/hello-world/makefile
@@ -1,4 +1,5 @@
 CFLAGS  = -std=c99
+CFLAGS += -g
 CFLAGS += -Wall
 CFLAGS += -Wextra
 CFLAGS += -pedantic
@@ -13,8 +14,7 @@ test: tests.out
 	@./tests.out
 
 memcheck: tests.out
-	@echo "Running memory check"
-	@valgrind $(VFLAGS) ./tests.out >> /dev/null
+	@valgrind $(VFLAGS) ./tests.out
 	@echo "Memory check passed"
 
 clean:
--- a/exercises/hello-world/makefile
+++ b/exercises/hello-world/makefile
@@ -4,9 +4,16 @@ CFLAGS += -Wextra
 CFLAGS += -pedantic
 CFLAGS += -Werror
 
+VFLAGS  = --quiet
+VFLAGS += --tool=memcheck
+VFLAGS += --leak-check=full
+VFLAGS += --error-exitcode=1
 
 test: tests.out
 	@./tests.out
+
+memcheck: tests.out
+	@valgrind $(VFLAGS) ./tests.out >> /dev/null
 
 clean:
 	rm -f *.o *.out
--- a/exercises/hello-world/makefile
+++ b/exercises/hello-world/makefile
@@ -13,7 +13,9 @@ test: tests.out
 	@./tests.out
 
 memcheck: tests.out
+	@echo "Running memory check"
 	@valgrind $(VFLAGS) ./tests.out >> /dev/null
+	@echo "Memory check complete"
 
 clean:
 	rm -f *.o *.out


exercises/hello-world/makefile
```diff
@@ -4,10 +4,17 @@ CFLAGS += -Wextra
 CFLAGS += -pedantic
 CFLAGS += -Werror
 
+VFLAGS  = --quiet
+VFLAGS += --tool=memcheck
+VFLAGS += --leak-check=full
+VFLAGS += --error-exitcode=1
```
Gamecock: This seems like something only needed for a script, does not help the user.

exercises/hello-world/makefile
```diff
@@ -4,10 +4,17 @@ CFLAGS += -Wextra
 CFLAGS += -pedantic
 CFLAGS += -Werror
 
+VFLAGS  = --quiet
+VFLAGS += --tool=memcheck
```
Gamecock: This is the default, not needed, but does not hurt.

exercises/hello-world/makefile
```diff
@@ -4,10 +4,17 @@ CFLAGS += -Wextra
...
+VFLAGS += --leak-check=full
+VFLAGS += --error-exitcode=1
 
 test: tests.out
 	@./tests.out
 
+memcheck: tests.out
+	@valgrind $(VFLAGS) ./tests.out >> /dev/null
```
Gamecock: Should we echo something hhere so users knows something happened.  Otherwise with --quiet it just returns.

exercises/hello-world/makefile
wolf99: Thought this might be useful for potential future change to `run-tests` to change from explicitly calling `valgrind` to instead using this `make` target. This would allow flexibility in the case that an exercise requires special consideration WRT `valgrind`. Is there any negative to using it?

exercises/hello-world/makefile
wolf99: I like to be explicit with options for stuff students use. Saying that I've no strong feelings on it and would be happy to take it out if you like?

exercises/hello-world/makefile
wolf99: With the added two `echo` statements this now also means that the second doesn't get output in case of failure.

Gamecock: LGTM, if the second only gets printed on a success then you could say `Memory check passed.`
I'm fine with the other flags, just wanted the discussion.  No strong feeling.

exercises/hello-world/makefile
ryanplusplus: My vote is to leave it in

ryanplusplus:
This looks good to me. I think that @patricksjackson has some `valgrind` experience so hopefully he'll have an opportunity to look as well.

patricksjackson:
The most important thing is to add `-g` to CFLAGS so the debug info is included in the binary. That will let Valgrind figure out the offending line numbers. Otherwise this is exactly how I was hoping this would look :)

One suggestion though, if you don't redirect stdout to null (and maybe delete the first echo) then the flow for people who want to use Valgrind can just be to use `make memcheck` all the time. If it's clean, there'd be a passing note, and if not we'd see the Valgrind output. That would probably make the documentation cleaner too, as you wouldn't need to explain running it twice.

That would also let us run the tests only once in CI too (I guess we could've done this before.... oh well).

I did a quick check and it looks like the valgrind call in bin/run-tests could be replaced by `make memcheck`, though I didn't check in a Travis environment.

Here's an example with debug info, I just added a malloc call into the hello function.
Without `-g` it looks like this:
```
Compiling tests.out
Running memory check
==3368== 100 bytes in 1 blocks are definitely lost in loss record 1 of 1
==3368==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==3368==    by 0x400683: hello (in /mnt/c/Users/Patrick/code/xc/exercises/hello-world/tests.out)
==3368==    by 0x402402: test_hello (in /mnt/c/Users/Patrick/code/xc/exercises/hello-world/tests.out)
==3368==    by 0x4022BF: UnityDefaultTestRun (in /mnt/c/Users/Patrick/code/xc/exercises/hello-world/tests.out)
==3368==    by 0x40243E: main (in /mnt/c/Users/Patrick/code/xc/exercises/hello-world/tests.out)
==3368==
makefile:16: recipe for target 'memcheck' failed
make: *** [memcheck] Error 1
```

And after adding it:
```
Compiling tests.out
Running memory check
==3487== 100 bytes in 1 blocks are definitely lost in loss record 1 of 1
==3487==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==3487==    by 0x400683: hello (hello_world.c:6)
==3487==    by 0x402402: test_hello (test_hello_world.c:15)
==3487==    by 0x4022BF: UnityDefaultTestRun (unity.c:1250)
==3487==    by 0x40243E: main (test_hello_world.c:22)
==3487==
makefile:17: recipe for target 'memcheck' failed
make: *** [memcheck] Error 1
```

wolf99:
Whoops! totally had `-g` and an explicit `-O0` in my local testing of this - promise! 😄 
Thanks for catching that and the other suggestion @patricksjackson

patricksjackson:
LGTM :+1:

ryanplusplus:
:+1:

PR #242, opened 2017-11-11
hello-world: remove unneeded includes

wolf99:
Seems the `hello_world.c`  stub still has two includes left in it from the previous implementation. This PR removes them. They are unneeded and could cause confusion or students that are very new to C. They also deviate from what is used in the `example.c` solution.


wolf99:
Shoot - I've accidentally done #241 on my master branch and now there'll be conflicts with this >.<

PR #243, opened 2017-11-15
hello-world: remove unneeded includes

wolf99:
The includes are left from the old implementation and are not needed for the current one


Gamecock:
LGTM

PR #244, opened 2017-11-15
Add memcheck make target

wolf99:
Per #169 and #241

### DIFF
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -28,12 +28,11 @@ for D in exercises/*; do
         { cd ${D};
           echo "Running tests for ${EXERCISE_NAME}";
           make clean >> /dev/null;
-          if make | grep FAIL: ; then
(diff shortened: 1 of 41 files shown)


wolf99:
Should be good to go now I think.

ryanplusplus:
Can we update the Travis script to execute this now? This would be a strong check for whether this hits every exercise.

wolf99:
Added to CI - seems to work fine (after fixing missing pesky quote mark!).
I have tested that it does produce a valid fail on my own machine by adding a `malloc(10)` call to the `example.c` for `hello-world` and running `$ ./bin/run-tests`. Would be great if someone could check that also to ensure I'm not missing something silly 😄

ryanplusplus:
LGTM :+1:

wolf99:
Closes #169

Issue #19, opened 2016-01-18
Makefiles missing source dependencies


McFadge:
Hi all,

New to this, so I'm sorry if I'm missing something obvious, but it seems that all the make files are missing source dependencies? Is it expected for the user to add them?

Also, it seems that it might be useful to include blank source files, especially since there are references to the required files in the Makefiles and test files. I feel like both of these would just be one less step for setting up.

kytrinyx:
I'm not quite sure what the status of this is. This track isn't active, and I don't think that it has been under active development over the past few months.

sunzenshen:
My impression was that there hasn't been a consensus on which build system to use:
https://github.com/exercism/xc/issues/4

With the tests as is, the user is expected to read the test #include(s) in order to figure out which source files to create. An alternative compromise could be to provide a stub header file, and let the participant consider breaking off code into an implementation file.

Either way, if you want to try experimenting with how to provide the build environment, now is a good time as any.

kytrinyx:
A stub header file would work, since any files not named /[Ee]xample/ are delivered to the user when they fetch the exercise.

McFadge:
I began creating stub .h and .c files and including them in the makefile dependencies. However, doing so made me wonder how much we expect to leave to the user. Should the program compile with no work? Should the program compile but pass no tests? Some tests? Should the user be deciding on datatypes?

Just including the dependencies without the source files means you get something like this when running `make`

`make: *** No rule to make target `src/leap.c', needed by `tests.out'.  Stop.`

Which people might be able to interpret correctly as "Hey, you should create this file".

Even including the .h files with just the appropriate function invites some decision of what datatypes to use, or the names of function arguments. But maybe including too much, like a dummy variable to return, would put the user off from modifying existing code.

Personally, I think providing stub files with a blank template for the required function is beneficial. It lets the user focus on solving the problem than having them consider whether they want an int or an unsigned int. Though, this being my first track, I'm not sure what is expected of exercism exercises.

kytrinyx:
In many languages we start the user off with nothing--it doesn't compile, we don't give them hints, and they have to piece together what to define by looking at the error messages and the tests.

We do, however, make sure that the test dependencies are all in place.

I think it's beneficial to have to think about whether you want an int or an unsigned int.

In the Go track we provide lots of hints and a very complete stub file for the first exercise, then we back off little by little.

sunzenshen:
[...]

PR #246, opened 2017-11-20
Fix small static analysis items

wolf99:
While testing CppCheck static analysis I see small items that could be fixed in several exercises.
This PR fixes those. Each commit is in relation to items in a different exercise.

### DIFF
--- a/exercises/meetup/src/example.c
+++ b/exercises/meetup/src/example.c
@@ -108,16 +108,15 @@ int meetup_day_of_month(unsigned int year, unsigned int month, const char *week,
                         const char *day_of_week)
 {
    int day_of_month = BAD_DATE_REQUESTED;
-   int baseline_day_of_week;
    int days_in_test_month = days_in_month(year, month);
-   int day_offset;
-   int reference_day = 1;       // used to calculate a baseline day of week given a target date.
 
    int which_week_of_month = get_week_of_month(week);
    int target_day_in_week = get_requested_day_of_week(day_of_week);
 
    // check for valid lookup...
    if ((which_week_of_month >= 0) && (target_day_in_week >= 0)) {
+      int day_offset;
+      int reference_day = 1;       // used to calculate a baseline day of week given a target date.
       // first - fifth Xday of month
       if (which_week_of_month <= 5) {
          day_offset = 1 + ((which_week_of_month - 1) * 7);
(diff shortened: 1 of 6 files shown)


exercises/meetup/src/example.c
```diff
@@ -108,16 +108,15 @@ int meetup_day_of_month(unsigned int year, unsigned int month, const char *week,
...
-   int day_offset;
-   int reference_day = 1;       // used to calculate a baseline day of week given a target date.
 
    int which_week_of_month = get_week_of_month(week);
    int target_day_in_week = get_requested_day_of_week(day_of_week);
 
    // check for valid lookup...
    if ((which_week_of_month >= 0) && (target_day_in_week >= 0)) {
+      int day_offset;
+      int reference_day = 1;    // used to calculate a baseline day of week given a target date.
```
ryanplusplus: 👍 👍

exercises/raindrops/src/example.c
```diff
@@ -15,7 +15,7 @@ char *convert(char result[], int drops)
       strcat(result, "Plong");
    }
    if (strlen(result) == 0) {
-      char drops_string[12] = "";
+      char drops_string[12] = "\0";
```
ryanplusplus: Isn't this equivalent? What was the warning that this fixes?

wolf99: The error given is for ln #19 (`strcat(result, drops_string)`). It reads `(error) Buffer is accessed out of bounds`.

Of course the buffer is _not_ accessed out of bounds by the example code. This then means that either the standard library functions used with the buffer directly reference some out of bounds index or this is a false positive. I find the former hard to credit. While searching online on this I came across [this sourcforge bug report](https://sourceforge.net/p/cppcheck/discussion/general/thread/2c33dfc5/) that mentions a similar false positive in the initialization. The fix for that made it into CppCheck 1.81, but the version in the Ubuntu PPA is 1.72. Getting a later version requires building from source so I've not been able to see if a later version fixes this due to time constraints (I have actually had a stab at this but it turned into dependency-fest).

While `= ""` should work, `= "\0";`, `= { '\0' }` or `= { 0 };` are all equally as valid. I'm not sure on the C spec for initialization of empty strings but I do know that for initializer *lists* it mentions something like 

> An initializer-list must have at least one initializer

and also 

> If there are fewer initializers in a brace-enclosed list than there are elements or members of an aggregate, or fewer characters in a string literal used to initialize an array of known size than there are elements in the array, the remainder of the aggregate shall be initialized implicitly the same as objects that have static storage duration

`"\0"` seems to fit the bill with the least amount of change.

ryanplusplus: Ah, sounds good. Thanks for the in-depth explanation!

ryanplusplus:
LGTM

PR #247, opened 2017-11-26
Remove *.dSYM folders in 'clean' rule

bcc32:
On macOS (using the Xcode compiler), compiling with debug symbols (`-g`) causes the creation of a _*.dSYM_ folder, e.g., _test.out.dSYM_.

This PR adds these folders to the makefile `clean` rules.


ryanplusplus:
Good catch, thanks

PR #248, opened 2017-11-30
Call missing tests for hamming exercise

bcc32:
Two of the tests aren't being called at all.


wolf99:
Good catch, LGTM.

ryanplusplus:
:+1:

Issue #250, opened 2017-12-13
Anagram is nested too deep in config.json


wolf99:
In #227 it was [decided](https://github.com/exercism/c/pull/227#discussion_r150229995) that more than one level of locking in the `config.json` was not a desirable thing. Looking at the visualization of the current `config.json` file we can see that the `anagram` exercise has this undesired multiple levels of lock.

![image](https://user-images.githubusercontent.com/281700/33931050-cdb4091a-dfe6-11e7-8619-c96f985217a2.png)

Where would be the best place to move the exercise in the ordering such that it does not suffer double-locking? Obviously it is thematically related to `acronym` and `pangram` but the current difficulty level of `5` suggests that it should be later in the track, possibly somewhere near `allergies`?

PS: Also note that there are two floating exercises -- this may also need rectifying.

PPS: FYI the config visualizer is at https://jonmcalder.shinyapps.io/exercism-config-viz/

ryanplusplus:
I think that we could probably drop `anagram` in as a child of `isogram` because the "side quest" exercises are optional and don't need to be completed to progress.

However, we have quite a few exercises that don't open up side exercises, so maybe we should add `leap` to `grains`, `raindrops` to `beer-song`, and drop `anagram` in at `allergies` or `binary-search`?

wolf99:
Closed by #251

PR #251, opened 2017-12-14
Fix exercise nesting

wolf99:
Per #250, this changes the nesting of some exercises that were either nested too deep or not linked to the core of the track at all (i.e. floating).

I also took the opportunity to reorder the `config.json` listing to approximate the exercise ordering as it was starting to get a little hard to find related exercises.

The visualisation of the `config.json` now appears like so (see the linked issue for the visualisation as it was previously):

image

### DIFF
--- a/config.json
+++ b/config.json
@@ -27,13 +27,39 @@
     {
       "core": false,
       "difficulty": 1,
-      "slug": "leap",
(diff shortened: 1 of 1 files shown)


ryanplusplus:
LGTM 👍

PR #254, opened 2017-12-29
Implement exercise 'complex-numbers'

clementi:


config.json
```diff
@@ -477,6 +477,17 @@
         "enums"
       ],
       "uuid": "b0a152e9-5a45-41f9-bda0-427111d9a56c"
+    },
+    {
+      "core": true,
```
ryanplusplus: Can you make this non-core and make it unlocked by something topical, maybe `difference-of-squares`?

exercises/complex-numbers/src/example.c
```diff
@@ -0,0 +1,70 @@
+#include "complex_numbers.h"
+#include <math.h>
+
+double square(double x) {
+    return x * x;
+}
```
ryanplusplus: Since this function is only used locally, can you declare it as `static` so that it is private to this module?

exercises/complex-numbers/src/complex_numbers.h
```diff
@@ -0,0 +1,19 @@
+#ifndef _COMPLEX_NUMBERS_H_
+#define _COMPLEX_NUMBERS_H_
+
+typedef struct Complex {
+    double real;
+    double imag;
+} Complex;
```
ryanplusplus: To be consistent in style, this should be named `complex_t`

exercises/complex-numbers/src/complex_numbers.h
```diff
@@ -0,0 +1,19 @@
...
+
+Complex c_add(Complex* a, Complex* b);
+Complex c_sub(Complex* a, Complex* b);
+Complex c_mul(Complex* a, Complex* b);
+Complex c_div(Complex* a, Complex* b);
+double c_abs(Complex* x);
+Complex c_conjugate(Complex *x);
+double c_real(Complex* x);
+double c_imag(Complex* x);
+Complex c_exp(Complex* x);
```
ryanplusplus: Since none of these should modify the provided number(s), can you mark them as `const`?

exercises/complex-numbers/test/test_complex_numbers.c
```diff
@@ -0,0 +1,192 @@
...
+
+    unsigned int i;
+    for (i = 0; i < sizeof(cases) / sizeof(Complex); i++) {
+        double result = c_abs(&cases[i]);
+
+        TEST_ASSERT_EQUAL(5.0, result);
+    }
+}
+
+// conjugate
```
ryanplusplus: These comments don't seem to add anything that the test name doesn't already say. Can they be removed?

exercises/complex-numbers/test/test_complex_numbers.c
```diff
@@ -0,0 +1,192 @@
...
+        { { .real = 1.0, .imag = 2.0 }, { .real = 3.0, .imag = 4.0 }, { .real = 4.0, .imag = 6.0 } }
+    };
+
+    unsigned int i;
+    for (i = 0; i < sizeof(cases) / sizeof(Complex[3]); i++) {
+        Complex result = c_add(&cases[i][0], &cases[i][1]);
+        
+        TEST_ASSERT_EQUAL(result.real, cases[i][2].real);
+        TEST_ASSERT_EQUAL(result.imag, cases[i][2].imag);
+    }
```
ryanplusplus: @wolf99 @patricksjackson what are your thoughts on using loops like this here? It definitely makes things more compact, but it 1) loses the meaningful names for the sub-tests (ie: add only real, add only imaginary), 2) makes it more difficult for someone to figure out which sub-test is failing, and 3) isn't really consistent with how the other test suites are written.

ryanplusplus:
Be sure to run the indent script to ensure that the formatting matches the expected formatting. See https://github.com/exercism/c/blob/master/docs/CONTRIBUTING.md#coding-style for details about formatting and style.

exercises/complex-numbers/test/test_complex_numbers.c
```diff
@@ -0,0 +1,192 @@
...
+        Complex result = c_exp(&cases[i][0]);
+
+        TEST_ASSERT_EQUAL(result.real, cases[i][1].real);
+        TEST_ASSERT_EQUAL(result.imag, cases[i][1].imag);
+    }
+}
+
+int main(void)
+{
+   UnityBegin("complex_numbers.c");
```
patricksjackson: @wolf99 @ryanplusplus Did we ever standardize on what to actually use here?

exercises/complex-numbers/test/test_complex_numbers.c
patricksjackson: I'm fine with it as long as it's clear what's getting tested while reading this source, and that the error messages are clear on failure.

Here's my suggestion. I'd check the following two methods to see which gives the better error messages

```
void compare_complex(complex_t lhs, complex_t rhs) {
  TEST_ASSERT_EQUAL_DOUBLE(lhs.real, rhs.real);
  TEST_ASSERT_EQUAL_DOUBLE(lhs.imag, rhs.imag);
}

void test_addition(void) {
  compare_complex(c_add({.real = 1.0, .imag = 0.0}, {.real = 2.0, .imag = 0.0}), {.real = 3.0, .imag = 0.0});
  // etc
}
```
Or
```
bool compare_complex(complex_t lhs, complex_t rhs) {
  return (lhs.real == rhs.real) && (lhs.imag == rhs.imag);
}
void test_addition(void) {
  TEST_ASSERT(compare_complex(c_add({.real = 1.0, .imag = 0.0}, {.real = 2.0, .imag = 0.0}), {.real = 3.0, .imag = 0.0}));
  // etc
}
```

exercises/complex-numbers/src/complex_numbers.h
patricksjackson: Why don't we just pass them by value instead? The pointer is already half the size of the struct so it's not like there's a lot of overhead, and it simplifies the interface.

exercises/complex-numbers/test/test_complex_numbers.c
clementi: I'll get rid of them.

exercises/complex-numbers/src/complex_numbers.h
clementi: I had considered that. I'll pass them by value.

exercises/complex-numbers/src/complex_numbers.h
clementi: Will do.

exercises/complex-numbers/src/example.c
clementi: You bet.

config.json
clementi: I'll change that.

exercises/complex-numbers/test/test_complex_numbers.c
clementi: How about I just make a separate function for each test case? It's not a big deal to do that.

clementi:
Well, I keep running `./indent.sh` and it keeps finding an indentation error.

patricksjackson:
There must be a difference with your version of indent and the one on Travis (?). Those lines could definitely be some weird edge case.

exercises/complex-numbers/test/test_complex_numbers.c
```diff
@@ -0,0 +1,365 @@
...
+#include "vendor/unity.h"
+
+#include <math.h>
+
+#define PI acos(-1)
+#define E exp(1)
+
+void compare_complex(complex_t lhs, complex_t rhs)
+{
+   TEST_ASSERT_EQUAL(lhs.real, rhs.real);
```
patricksjackson: Use `TEST_ASSERT_EQUAL_DOUBLE` to compare doubles, this I believe boils down to just an '==' which shouldn't be used.

exercises/complex-numbers/test/test_complex_numbers.c
```diff
@@ -0,0 +1,365 @@
...
+
+   complex_t expected = {.real = -1.0,.imag = 0.0 };
+   complex_t result = c_mul(z, z);
+
+   compare_complex(result, expected);
+}
+
+void test_add_purely_real_numbers(void)
+{
+   complex_t z1 = {.real = 1.0,.imag = 0.0 };
```
patricksjackson: Can you add the `TEST_IGNORE()` line to all tests but the first? See #179.

exercises/complex-numbers/src/complex_numbers.h
```diff
@@ -0,0 +1,19 @@
+#ifndef _COMPLEX_NUMBERS_H_
+#define _COMPLEX_NUMBERS_H_
+
+typedef struct complex_t {
```
patricksjackson: Here you only need `typedef struct { double real; double imag; } complex_t;`, you don't need to write `complex_t` twice.

See [this stackoverflow post](https://stackoverflow.com/a/612350)

clementi:
After switching to `TEST_ASSERT_EQUAL_DOUBLE`, I'm getting all kinds of unused variable warnings as errors. I just need to do `TEST_ASSERT_EQUAL_DOUBLE(expected, actual)`, correct? What am I missing?

exercises/complex-numbers/test/test_complex_numbers.c
```diff
@@ -0,0 +1,395 @@
...
+
+   complex_t expected = {.real = -1.0,.imag = 0.0 };
+   complex_t actual = c_mul(z, z);
+
+   compare_complex(expected, actual);
+}
+
+void test_add_purely_real_numbers(void)
+{
+   TEST_IGNORE()
```
patricksjackson: All these need a semicolon.

patricksjackson:
On testing doubles: According to Unity's docs that should just work. According to `unity.h` you also need to `#define UNITY_INCLUDE_DOUBLE` before including `unity.h`, but _that_ doesn't work either.

What you _can_ do instead is use `TEST_ASSERT_EQUAL_FLOAT` instead of `TEST_ASSERT_EQUAL_DOUBLE`, and add `#define UNITY_FLOAT_TYPE double` at the top of the test file. That will just use Unity's normal float comparison and redefine the type name to be double.

(Sorry, I did not know it was that bad when I suggested it, but that's the cause of the unused variable warnings.)

clementi:
@patricksjackson But using `TEST_ASSERT_EQUAL_FLOAT` with `#define UNITY_FLOAT_TYPE double` causes this error (because of `-Werror`):

    test/test_complex_numbers.c:10:0: error: "UNITY_FLOAT_TYPE" redefined [-Werror]

patricksjackson:
You need to put the #define above the unity.h include. If the unity header doesn't see it it will define it as 'float'.

clementi:
How do we look now?

exercises/complex-numbers/src/complex_numbers.c
```diff
@@ -0,0 +1,46 @@
+#include "complex_numbers.h"
+
+complex_t c_add(complex_t a, complex_t b)
```
wolf99: I think the parameters should be `const` here as well as in the header file, right?

exercises/complex-numbers/test/test_complex_numbers.c
wolf99: This should be the test file name - every other usage so far is `UnityBegin("test/test_complex_numbers.c")`. The [examples given in the Unity repo](https://github.com/ThrowTheSwitch/Unity/blob/master/examples/example_1/test/test_runners/TestProductionCode_Runner.c) indicate that this is correct

wolf99: This is actually shown in the example in https://github.com/exercism/c/blob/master/docs/C_STYLE_GUIDE.md

wolf99:
I've never looked into why, but the UNITY_INCLUDE_DOUBLE does work if defined in the makefile with the `-D` flag to GCC. However, the best way of setting Unity options is to add and use the `unity_config.h` file (from the unity repo).

exercises/complex-numbers/src/complex_numbers.c
ryanplusplus: Since these aren't passed by reference, it doesn't really matter, right? It only really matters in the C file (and only because it constrains your implementation).

ryanplusplus: Oh, I see now that the included header includes the `const`s. My preference would be non-`const` everywhere so that we don't lead anyone to believe that the `const` makes a difference, but as long as we're consistent I could be happy either way.

wolf99: Ah, of course; yes I'd agree, although either way is good my preference is to remove `const`, but whichever is chosen, all definitions should be consistent.

ryanplusplus:
Thanks @clementi!

PR #260, opened 2018-01-26
roman-numerals: added test case "49" as per updated canonical-data.json

Zarimax:
Aligns test_roman_numerals.c with new test case specified in https://github.com/exercism/problem-specifications/pull/1139


ryanplusplus:
Looks great, thanks!

Issue #262, opened 2018-02-02
pangram: misleading description about non-ASCII symbols


Vitorvgc:
The description states:

> The alphabet used consists of ASCII letters a to z, inclusive, and is case insensitive. **Input will not contain non-ASCII symbols.**

Although, the two last test cases involve non-ASCII symbols:

```c
void test_non_ascii_characters(void)
{
   TEST_IGNORE();
   const char sentence[] =
       "Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich.";

   TEST_ASSERT_TRUE(is_pangram(sentence));
}

void test_pangram_in_alphabet_other_than_ascii(void)
{
   TEST_IGNORE();
   const char sentence[] =
       "Широкая электрификация южных губерний даст мощный толчок подъёму сельского хозяйства.";

   TEST_ASSERT_FALSE(is_pangram(sentence));
}
```

I think that either the description should clarify that input can contain non-ASCII symbols or these test cases should be removed.

wolf99:
I think this is a case where the implementation and tests here have not been updated to the latest version of the problem specification. The current problem specification does not include these inputs. 

Furthermore there may be other exercise that exhibit similar issues due to the same problem. There is ongoing discussion in #261 and there has been discussion in the problem-specifications repo about this, however the issue may be disappearing in nextercism (I am unclear on if an how this is actually happening).

For this particular problem @Vitorvgc, would you be bale to submit a pull request with changes to make the pangram exercise tests match the current problem specification's canonical data? ([https://github.com/exercism/problem-specifications/tree/master/exercises/pangram](https://github.com/exercism/problem-specifications/tree/master/exercises/pangram))

Vitorvgc:
Yes, I can work on it. I'll submit a PR soon.

Vitorvgc:
Fixed by #263

PR #263, opened 2018-02-03
pangram: Update test cases to conform to canonical data

Vitorvgc:
Solves #262 

Changed the tests to match the canonical data (https://github.com/exercism/problem-specifications/tree/master/exercises/pangram).

### DIFF
--- a/exercises/pangram/test/test_pangram.c
+++ b/exercises/pangram/test/test_pangram.c
@@ -9,31 +9,25 @@ void tearDown(void)
 {
 }
 
-void test_null(void)
-{
-   TEST_ASSERT_FALSE(is_pangram(NULL));
-}
-
 void test_empty_sentence(void)
 {
-   TEST_IGNORE();               // delete this line to run test
    const char sentence[] = "";
 
    TEST_ASSERT_FALSE(is_pangram(sentence));
+}
+
+void test_perfect_lowercase_pangram(void)
+{
+   TEST_IGNORE();               // delete this line to run test
+   const char sentence[] = "abcdefghijklmnopqrstuvwxyz";
+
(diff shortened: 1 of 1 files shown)


exercises/pangram/test/test_pangram.c
```diff
@@ -9,31 +9,25 @@ void tearDown(void)
 {
 }
 
-void test_null(void)
```
ryanplusplus: There's a special case for this test in the example. Can we remove that special case now that this test is not included?

Vitorvgc: I was in doubt between removing and keeping this test case. Although it is not explicitly in the canonical data, the majority of the exercises that I solved have tests to check null input, and it may be relevant to remember.

wolf99: I think there was a decision on the problem-specifications repo that null tests were unnecessary... I'll see if I can find the convo before I'll say for sure though

wolf99: Heres an issue that summarizes the current stance https://github.com/exercism/problem-specifications/issues/1071

Basically, we can check for invalid inputs such as `null` as an edge case argument for a string parameter in a per-track manner. But where languages don't allow for it, attempting to input integer arguments in for a string parameter would not make sense in languages that don't facilitate this kind of thing.

In this case that means, IMO, that it is basically up to the track maintainers if invalid cases should be tested. For my two c, if there is a test, and a check for that case in the implementation as-is, then leave it in.

Vitorvgc: Thanks for the clarification. I added back the null test.

wolf99:
LGTM - thanks for updating this @Vitorvgc

ryanplusplus:
Thanks!

PR #268, opened 2018-02-12
Complete palindrome_products to fully match the problem description

vlzware:
- check for the factors of the returned palindromes;
- more test cases.

### DIFF
--- a/exercises/palindrome-products/src/example.c
+++ b/exercises/palindrome-products/src/example.c
@@ -3,15 +3,17 @@
 #include <stdio.h>
 #include <limits.h>
 
-int palindrome(int n);
-void addfactors(factor_t ** p, int i, int k);
-void check_alloc(void *p);
-void swap(int *i, int *k);
+static int palindrome(int n);
+static int addfactors(factor_t ** p, int i, int k);
+static void free_ll(struct factors *p);
 
 product_t *get_palindrome_product(int from, int to)
 {
-   product_t *res = (product_t *) malloc(sizeof(product_t));
-   check_alloc(res);
+   product_t *res = malloc(sizeof(product_t));
+   if (res == NULL) {
+      fprintf(stderr, "Memory error!\n");
+      return NULL;
+   }
 
    res->error[MAXERR - 1] = '\0';
    res->smallest = INT_MAX;
@@ -26,15 +28,20 @@ product_t *get_palindrome_product(int from, int to)
    }
 
    int i, k, n;
+   int err = 0;
    for (i = from; i <= to; i++)
       for (k = i; k <= to; k++)
          if (palindrome(n = i * k)) {
             if (n <= res->smallest) {
                res->smallest = n;
-               addfactors(&res->factors_sm, i, k);
+               err = addfactors(&res->factors_sm, i, k);
             } else if (n >= res->largest) {
                res->largest = n;
-               addfactors(&res->factors_lg, i, k);
+               err = addfactors(&res->factors_lg, i, k);
+            }
+            if (err) {
+               free(res);
+               return NULL;
             }
          }
 
@@ -46,7 +53,16 @@ product_t *get_palindrome_product(int from, int to)
    return res;
 }
 
-void free_ll(struct factors *p)
+void free_product(product_t * p)
+{
+   if (p == NULL)
+      return;
+   free_ll(p->factors_lg);
+   free_ll(p->factors_sm);
+   free(p);
+}
+
+static void free_ll(struct factors *p)
 {
    if (p == NULL)
       return;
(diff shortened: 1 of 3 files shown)


ryanplusplus:
Hey @vlzware, there is already a list of canonical tests for this exercise at https://github.com/exercism/problem-specifications/blob/master/exercises/palindrome-products/canonical-data.json. The reason that we don't currently follow this is that the exercise was created well before the exercise was defined to the extent that it is now. Given that there now _is_ a list of canonical tests, can you update your test suite to match the common test specification?

exercises/palindrome-products/test/test_palindrome_products.c
```diff
@@ -1,6 +1,8 @@
 #include "vendor/unity.h"
 #include "../src/palindrome_products.h"
 
+product_t *product;
```
ryanplusplus: This is a globally visible symbol. Can you either make this `static` so that it's only visible within this file or simply push its declaration into each test? I don't know that there's a compelling reason to share this declaration.

exercises/palindrome-products/src/example.c
```diff
@@ -1,36 +1,104 @@
 #include "palindrome_products.h"
-#include <stdbool.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <limits.h>
 
-/* Return true if the product of a and b is a palindrome, false otherwise */
-bool is_palindrome_product(int a, int b)
+int palindrome(int n);
```
wolf99: File-scope functions should probably be marked with `static`

exercises/palindrome-products/src/example.c
```diff
@@ -1,36 +1,104 @@
...
+void swap(int *i, int *k);
+
+product_t *get_palindrome_product(int from, int to)
 {
-   int ret = 0, prod = a * b;
-   while (prod) {
-      ret *= 10;
-      ret += prod % 10;
-      prod /= 10;
+   product_t *res = (product_t *) malloc(sizeof(product_t));
```
wolf99: It is good practice _not_ to cast the value returned from `malloc()`

exercises/palindrome-products/src/example.c
```diff
@@ -1,36 +1,104 @@
...
+   }
+
+   return (nn == r);
+}
+
+void check_alloc(void *p)
+{
+   if (p == NULL) {
+      fprintf(stderr, "Memory error!\n");
+      exit(1);
```
wolf99: Factoring out the allocation check probably saves some typing but I would tend to shy away from using `exit()`... what do others think on this point?

A pattern that i often used to keep `malloc` checking code to a  vertical minimum is to do something like:

```c
product_t *res = NULL;

if ( (res = malloc(sizeof(product_t)) == NULL )
{
    // print out some message ...
    return 1;
}
```

exercises/palindrome-products/src/example.c
```diff
@@ -1,36 +1,104 @@
...
+
+void addfactors(factor_t ** p, int i, int k)
+{
+   int n = i * k;
+   if ((*p == NULL) || (((*p)->factor_a) * (*p)->factor_b != n)) {
+      free_ll(*p);
+      *p = NULL;
+   }
+
+   factor_t *tmp = (factor_t *) malloc(sizeof(factor_t));
```
wolf99: Same again for the casting here 😃

wolf99:
I've made a few stylistic comments, that could change, but otherwise good work 👍

vlzware:
Thanks for the review and the suggestions @ryanplusplus and @wolf99

exercises/palindrome-products/src/example.c
ryanplusplus: Most of my work is done without a heap, but when there's an unrecoverable error of this sort, I tend to think that getting out as quickly as possible is often a good choice. So I'm not opposed to using `exit()`, but I think that either solution works in this environment.

exercises/palindrome-products/src/example.c
```diff
@@ -1,36 +1,109 @@
...
-#include <stdbool.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <limits.h>
 
-/* Return true if the product of a and b is a palindrome, false otherwise */
-bool is_palindrome_product(int a, int b)
+static int palindrome(int n);
+static int addfactors(factor_t ** p, int i, int k);
+static void free_ll(struct factors *p);
```
ryanplusplus: I don't think there's any need to change, but something that helps to keep your code DRY (https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) is to avoid static prototypes by reorganizing your code so that helpers are always above the functions that use them. This means that if you ever refactor a name or a signature you only have to change it in one place, not two.

exercises/palindrome-products/src/example.h
```diff
@@ -1,12 +1,27 @@
...
-typedef struct {
+#define MAXERR 100
+
+struct factors {
+   int factor_a;
+   int factor_b;
+   struct factors *next;
+};
+
+typedef struct factors factor_t;
```
ryanplusplus: No need to change, but you can do this more simply with:

```c
typedef struct factors {
   int factor_a;
   int factor_b;
   struct factors *next;
} factor_t;
```

And if you don't intend to ever declare one of these as `struct factors` (and instead plan to only use `factor_t`), you can simplify further to:

```c
typedef struct {
   int factor_a;
   int factor_b;
   struct factors *next;
} factor_t;
```

ryanplusplus:
Thanks @vlzware for the extra effort and thanks @wolf99 for the review!

vlzware:
Thanks for the review @ryanplusplus 
I will gladly improve my coding style.

PR #269, opened 2018-02-17
Implement exercise 'etl'

vlzware:


exercises/etl/test/test_etl.c
```diff
@@ -0,0 +1,121 @@
...
+      return (b == NULL) ? 0 : 1;
+   if (b == NULL)
+      return 1;
+   while (--len >= 0)
+      if ((a[len].key != b[len].key) || (a[len].value != b[len].value))
+         return 1;
+   return !(len == -1);
+}
+
+//Tests start here
```
wolf99: This comment is not needed. Comments that mark sections may seem like a good idea but are generally considered to be against the ideas of clean coding (see Uncle Bob, etc)

exercises/etl/src/example.c
```diff
@@ -0,0 +1,47 @@
...
+}
+
+int convert(legacy_map * input, int input_len, new_map ** output)
+{
+   if (input == NULL || input_len <= 0)
+      return 0;
+
+   int len = 0;
+   int i;
+   for (i = 0; i < input_len; i++)
```
wolf99: This and the later `for` loop could probably become `for (size_t i = 0; ...`
Additionally I would be for adding braces to wrap the single operation in the `for` block. I'd be interested in thoughts on this though as it is an almost purely stylistic concern. Whichever is decided could be added to the style guide maybe.
EDIT: Just realized also that if the indexing variable declaration is moved into the initialization condition of the `for`, then the `for` will _have_ to have braces to create the necessary block scope.

exercises/etl/src/example.c
```diff
@@ -0,0 +1,47 @@
...
+#include <stdio.h>
+
+static int compare(const void *a, const void *b)
+{
+   new_map *p = (new_map *) a;
+   new_map *q = (new_map *) b;
+   return (p->key - q->key);
+}
+
+int convert(legacy_map * input, int input_len, new_map ** output)
```
wolf99: Is it possible to convert the parameter types so that the interface is simpler: `int convert legacy_map input[], int input_len, new_map * output[]);`.

wolf99:
Couple of very minor review comments there @vlzware, good work 👍

exercises/etl/src/example.c
ryanplusplus: +1 for moving the declaration into the `for`

exercises/etl/src/example.h
```diff
@@ -0,0 +1,16 @@
...
+   int value;
+   const char *keys;
+} legacy_map;
+
+typedef struct {
+   char key;
+   int value;
+} new_map;
+
+int convert(legacy_map * input, int input_len, new_map ** output);
```
ryanplusplus: How about adding a `const` to `legacy_map` since it shouldn't be touched?

exercises/etl/src/example.c
```diff
@@ -0,0 +1,47 @@
...
+      len += strlen(input[i].keys);
+
+   *output = malloc(len * sizeof(new_map));
+   if (*output == NULL) {
+      fprintf(stderr, "Memory error!\n");
+      return -1;
+   }
+
+   int pos = 0;
+   char *tmp;
```
ryanplusplus: Can you move the declaration of `tmp` into the body of the `for`? This makes it clear that it's value is not used outside of the `for`.

exercises/etl/src/example.c
```diff
@@ -0,0 +1,47 @@
...
+         (*output)[pos].key = tolower(*tmp);
+         (*output)[pos].value = input[i].value;
+         tmp++;
+         pos++;
+      }
+   }
+
+   /* full alphabet */
+   if (len == 26)
+      qsort(*output, len, sizeof(new_map), compare);
```
ryanplusplus: Why do we only sort for full alphabet? I think that the problem specifications were written assuming that the language has an associative array data type and so the order given there isn't really important. For the C track I agree that order is important (otherwise it's a huge pain to compare the output), but I think that means that we should always sort to reduce confusion and (seemingly) arbitrary requirements.

ryanplusplus:
A couple more comments. Looks great overall. Thanks!

vlzware:
Thanks again @wolf99  and @ryanplusplus for the review and the suggestions.

exercises/etl/src/example.c
vlzware: @wolf99 - Can you please elaborate a bit more about the "have to have braces". I followed you suggestion, bit still can't follow why this is a must.

exercises/etl/src/example.c
vlzware: Whoops, I clearly overlooked one test in the canonical data for this, which was the cause of this confusion.

exercises/etl/src/example.c
wolf99: @vlzware IIRC (and I very well may not, so advised to test, I don't have an environment that I can test this myself just ATM): if the index variable (`i`) declaration is moved into the initialization condition of the `for` loop as suggested, then the main idea is that the scope of that indexing variable is limited to the body of the `for`. If the `for`doesn't have braces it doesn't have a body _as far as scope is concerned_ (i.e. no block scope is created for it to exist within). Because you use `i` with the operation within the `for` loop instead of just within the for conditions only, then it needs to be in scope, thus you need braces to create the block scope.

exercises/etl/src/example.c
wolf99: Well caught folks!

exercises/etl/src/example.h
wolf99: Does it even need to be a pointer? The legacy map is a fixed length array (as far as `convert`is concerned I mean). This could `potentially change to `int convert (legacy_map input[], int input_len, new_map ** output);`. Depends how you feel about having both pointer type and array type parameters. Personally I don't feel it's a major concern except that it follows C's general idea of being as restrictive as possible  for types and makes the types explicit as to which parameter is the input and which is the output.

wolf99:
Little discussion for consideration on Ryan's first review comment, otherwise it LGTM

exercises/etl/src/example.c
vlzware: @wolf99 This does not seem to be true.
But as far as stylistic preferences go, I have no problem in using braces :)

exercises/etl/src/example.h
ryanplusplus: I don't think that I follow regarding it being a fixed length array. Isn't the actual length of the array unknown?

If you're not putting an array size, `legacy_map input[]` as a parameter is AFAIK identical to `legacy_map *input`. It doesn't pass by value and it doesn't even restrict the size of the array (any provided size is ignored). The only real difference with array types and pointer types as arguments is the behavior of `sizeof`. Only if you do something like `input[static n]` do you put any constraint on the passed array. C is bizarre sometimes.

ryanplusplus:
LGTM, thanks @vlzware!

PR #270, opened 2018-02-20
Implement exercise 'luhn'

vlzware:


exercises/luhn/src/example.c
```diff
@@ -0,0 +1,38 @@
...
+         char c = num[pos] - '0';
+         zeros += (c == 0);
+         if (every_second) {
+            c *= 2;
+            res += c > 9 ? c - 9 : c;
+         } else {
+            res += c;
+         }
+         every_second = !every_second;
+      } else if (num[pos] != ' ') {
```
wolf99: The readme states that spaces are allowed in input (but that they must be stripped before checking). Further, as the nature of allowed space is not specifically defined (to be 0x32), it may be better to use `isspace()` if checking for it.

exercises/luhn/test/test_luhn.c
```diff
@@ -0,0 +1,109 @@
...
+void test_a_simple_valid_SIN_that_becomes_invalid_if_reversed(void)
+{
+   TEST_IGNORE();
+   TEST_ASSERT_TRUE(luhn("59"));
+}
+
+void test_a_valid_Canadian_SIN(void)
+{
+   TEST_IGNORE();
+   TEST_ASSERT_TRUE(luhn("055 444 285"));
```
wolf99: Have I misunderstood something? How does this test pass if spaces cause the function to exit early and return false?

vlzware: @wolf99 Spaces are just skipped. ```false``` is returned from the loop if the current char is not digit and also not ```' '``` (space). ```isspace``` checks for much more, which I think should be an error. For example this:
```
123<tab>   <spaces>456\r\n
<tab>        789\r
``` 
would be considered okay by using ```isspace```.

exercises/luhn/src/example.c
ryanplusplus: @wolf99 I think that you misread this as `==` instead of `!=`. This is stating that if it's not a digit _and it's not a space_ then it's invalid.

ryanplusplus:
LGTM 👍

exercises/luhn/src/example.c
wolf99: @ryanplusplus You are correct, I read that incorrectly 🙄

exercises/luhn/test/test_luhn.c
wolf99: Yes, this was my point. The readme doesn't specify `' '` only "space", which is somewhat subjective... only a suggestion anyway. Otherwise everything LGTM

PR #271, opened 2018-02-20
Implement exercise 'prime-factors'

vlzware:


vlzware:
'luhn' files are not there (the solution is still not accepted), the first commit is somewhat misleading.

Anyway I am so happy to be able to learn so much here - about C, about git, workflow, getting feedback etc.

exercises/prime-factors/src/example.h
```diff
@@ -0,0 +1,11 @@
+#ifndef PRIME_FACTORS_H
+#define PRIME_FACTORS_H
+
+#include <stdint.h>
+#include <stddef.h>
+
+#define MAXFACTORS 10
+
+size_t find_factors(uint64_t n, uint64_t * factors);
```
ryanplusplus: I believe you can use `uint64_t factors[static MAXFACTORS]` to require that `factors` have at least 10 elements.

exercises/prime-factors/test/test_prime_factors.c
```diff
@@ -0,0 +1,96 @@
+#include "vendor/unity.h"
+#include "../src/prime_factors.h"
+#include <stdint.h>
+
+uint64_t buf[MAXFACTORS] = { 0 };
```
ryanplusplus: Be sure to mark this `static`

ryanplusplus:
@vlzware because of how git works, merging this after merging `luhn` will erase the `luhn` exercise. This is because they both contain the commit that adds `luhn` but this also includes a commit that removes it. So when they are both merged, you'll only have the single commit adding `luhn` and you'll have the commit from this PR that removes it.

There are some ways to resolve this, but they involve re-writing the git history for this branch. You can Google for some solutions to this, otherwise you can copy and paste your changes into a new branch and re-create the PR or you can clean up this branch after we merge `luhn`.

If this is not clear let me know and I can try to explain differently.

exercises/prime-factors/src/example.h
vlzware: This is also new to me. For the sake of understanding - what could have happen if the size was declared without ```static```?

vlzware:
Thanks again and sorry for the git mess. I'll give my best to fix it.

wolf99:
@ryanplusplus what is the policy regarding function names vs canonical-data "property"? For example, here vlzware has implemented the function under test with the reasonable name `find_factors()`. Should this though follow the canonical data `property` value and instead be `factors()`?

FYI @vlzware IMO `find_factors()` seems entirely reasonable, so if the policy is not to stick to the canonical-data rigidly in this regard, then it is fine as-is.

exercises/prime-factors/src/example.h
wolf99: AS the comments on that blog post and the second paragraph [here][1] remark A) this is not supported by many compilers (e.g. GCC doesn't), B) if the guarantee made by the programmer is not honored by the caller then the behaviour is undefined, and C) the optimization gains are only apparent on some architectures

On the other hand, using the `const` qualifier _does_ appear to work on GCC (>=v7.2.0 at least).

[1]: https://www.ibm.com/support/knowledgecenter/en/ssw_ibm_i_71/rzarg/static_array_index.htm

ryanplusplus:
@wolf99 I don't feel strongly about sticking to the canonical data `property` value so I'm fine with `find_factors()`.

If it wouldn't make maintenance harder, I'd go as far as to say that as long as we cover the same content as the tests in the canonical data, we could use better names there too.

ryanplusplus:
@vlzware no worries about the git stuff. Everyone makes a mistake like that when learning git because it works differently (I think better!) than older version control tools. Looks like everything cleaned up nicely.

exercises/prime-factors/src/example.h
ryanplusplus: @wolf99 thanks for the info. I've never actually used this before (just picked it up from someone well-versed in C standards trivia on Stack Overflow) so wasn't sure about the exact behavior. Sounds like this isn't terribly useful in practice. Thanks for doing some digging.

ryanplusplus:
This looks good to me. @wolf99 if you're happy then feel free to merge this.

vlzware:
I have to admit - I never paid attention of the 'property' value - still learning how everything works here.

Thanks for the feedback.

PR #272, opened 2018-02-20
Implement exercise 'bracket-push'

vlzware:


exercises/bracket-push/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+   return false;
+}
+
+bool bbp_okay(const char *input)
+{
+   if (input == NULL || *input == '\0') {
+      return true;
+   }
+
+   char stack[MAXBBP] = { 0 };
```
ryanplusplus: I don't know that it's worth it, but you could use the C call stack as your stack and not need to allocate a fixed length array that might not be big enough. This would use a lot more memory, though :)

exercises/bracket-push/src/example.c
```diff
@@ -0,0 +1,50 @@
+#include "bracket_push.h"
+#include <stdio.h>
+
+static bool is_bbp(char c)
```
ryanplusplus: `bbp` is bracket, brace, or parenthesis?

ryanplusplus:
Just a few comments from me

exercises/bracket-push/src/example.c
vlzware: I am afraid I can not follow - can you please explain a bit more? This sounds interesting

exercises/bracket-push/src/example.c
vlzware: Yes

exercises/bracket-push/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+bool bbp_okay(const char *input)
+{
+   if (input == NULL || *input == '\0') {
+      return true;
+   }
+
+   char stack[MAXBBP] = { 0 };
+   int top = 0;
+
+   char *tmp = (char *)input;
```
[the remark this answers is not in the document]
vlzware: Here again new and interesting stuff for me :)
How can I iterate over a ```const char*``` in this case - maybe using indexes? What happens then on the microcontroller when I iterate with ```char*``` over ```const char*```?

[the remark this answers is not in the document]
wolf99: Yes, array indexing is used instead. Try compiling with `-Wcast-qual` (with GCC) to see warnings about discarding const qualifiers.

exercises/bracket-push/src/example.c
```diff
@@ -0,0 +1,50 @@
...
+      return b == '{';
+   if (a == ']')
+      return b == '[';
+   if (a == ')')
+      return b == '(';
+
+   return false;
+}
+
+bool bbp_okay(const char *input)
```
wolf99: Unusual acronyms in public interfaces are often frowned upon - maybe go with `is_paired()` or something a little more verbose?

exercises/bracket-push/src/example.c
wolf99: might be better to make this `is_brace_bracket_paren()` or maybe split it into `is_brace()`, `is_bracket()` & `is_parenthesis()` (if going the latter route you could possibly add `inline` ).

wolf99:
A few small comments from me too but again, pretty good! 👍

exercises/bracket-push/src/example.c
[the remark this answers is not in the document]
ryanplusplus: @wolf99 thanks for the tip about `-Wcast-qual`. That is my new favorite warning.

exercises/bracket-push/src/example.c
ryanplusplus: Because I think this is interesting, I made a solution in Lua. If you don't know Lua but know JavaScript you will probably be able to understand the logic. You can view it here if you want to check it out:
http://exercism.io/submissions/b4bff072c1b84b7baba62e1e9c2de7ad

vlzware: The idea was indeed very interesting. Thanks.

I updated the example.c to use recursion. It is not mine solution (see link) but I like it better.

vlzware:
Thanks for the constructive feedback @wolf99 and @ryanplusplus

exercises/bracket-push/src/example.c
```diff
@@ -0,0 +1,52 @@
...
+{
+   return (c == '{') || (c == '[') || (c == '(');
+}
+
+static bool is_closing(const char c)
+{
+   return (c == '}') || (c == ']') || (c == ')');
+}
+
+/* https://stackoverflow.com/a/2718114/6049386 */
```
wolf99: Might be worth adding the date (year and month) here, and some reference to the MIT license SO use for contributions, which is compatible with the MIT license used here. SO can, and have changed licenses before; they used to use CC BY-SA, which (I'm not at all sure) I think is not directly compatible with MIT.

I'm sure this is probably overkill, but if we include attributions (we should, where needed) then we should do so precisely.

wolf99: `/* https://stackoverflow.com/a/2718114/6049386 , MIT,  Feb 2018*/`would probably be sufficient

exercises/bracket-push/src/example.c
```diff
@@ -0,0 +1,52 @@
...
+
+bool is_paired(const char *input)
+{
+   if (input == NULL) {
+      return true;
+   }
+
+   const char *res = match(input);
+   return (*res == '\0');
+}
```
ryanplusplus: This is a very elegant solution. I am in love.

ryanplusplus:
LGTM once the attribution is updated to @wolf99's suggestion.

PR #274, opened 2018-02-23
Implement exercise 'crypto-square'

vlzware:
Omitted error checking on 'calloc' to avoid over-bloating. The tests check already for 'NULL'. Interested to hear your thoughts about this (and about the whole solution of course).

### DIFF
--- a/exercises/crypto-square/makefile
+++ b/exercises/crypto-square/makefile
@@ -1,0 +1,25 @@
+CFLAGS  = -std=c99
+CFLAGS += -g
+CFLAGS += -Wall
+CFLAGS += -Wextra
+CFLAGS += -pedantic
+CFLAGS += -Werror
+
+VFLAGS  = --quiet
+VFLAGS += --tool=memcheck
+VFLAGS += --leak-check=full
+VFLAGS += --error-exitcode=1
+
+test: tests.out
+	@./tests.out
+
+memcheck: tests.out
+	@valgrind $(VFLAGS) ./tests.out
+	@echo "Memory check passed"
+
+clean:
+	rm -rf *.o *.out *.out.dSYM
+
+tests.out: test/test_crypto_square.c src/crypto_square.c src/crypto_square.h
+	@echo Compiling $@
+	@cc $(CFLAGS) src/crypto_square.c test/vendor/unity.c test/test_crypto_square.c -o tests.out -lm
--- a/exercises/crypto-square/src/example.c
+++ b/exercises/crypto-square/src/example.c
@@ -1,0 +1,92 @@
+#include "crypto_square.h"
+#include <stdlib.h>
+#include <ctype.h>
+#include <stdio.h>
+#include <math.h>
+
+/* mstrlen: get the length of input counting only alnums */
+static int mstrlen(const char *input)
+{
+   int i = 0;
+   while (*input) {
+      if (isalnum(*input))
+         i++;
+      input++;
+   }
+
+   return i;
+}
+
+/* get_cols_and_rows: calculate columns and rows from len */
+static void get_cols_and_rows(int len, int *cols, int *rows)
+{
+   *cols = sqrt(len);
+
+   if (*cols * *cols == len) {
+      *rows = *cols;
+   } else if (*cols * (*cols + 1) >= len) {
+      *rows = *cols;
+      (*cols)++;
+   } else {
+      (*cols)++;
+      *rows = *cols;
+      if ((*cols * *cols) < len)
+         (*cols)++;
+   }
+}
+
+/* get_pos_cipher: calculate the position in the cipher string */
+static int get_pos_cipher(const int pos_inp, const int rows, const int cols)
+{
+   return (pos_inp / cols) + (pos_inp % cols) * (rows + 1);
+}
+
+char *ciphertext(const char *input)
+{
+   if (input == NULL)
+      return NULL;
+   if (*input == '\0')
+      return calloc(1, 1);
+
+   int len_inp = mstrlen(input);
+   if (len_inp == 0)
+      return calloc(1, 1);
+
+   int cols, rows;
+   get_cols_and_rows(len_inp, &cols, &rows);
+
+   /* no need for extra row if single line */
+   int len_out = rows > 1 ? (rows + 1) * cols : rows * cols;
+
+   char *res = malloc(len_out + 1);
+   if (res == NULL) {
+      fprintf(stderr, "Memory error!\n");
+      return NULL;
+   }
+
+   int pos_inp = 0;
+   while (*input) {
+      if (isalnum(*input)) {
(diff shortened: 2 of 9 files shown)


vlzware:
Great... I need help on this.
Because of the length of the function names, running indent produces every time different result on the same file. Like this:
```
# indent.sh test_crypto_square.c
# diff test_crypto_square.c test_crypto_square.c~
67c67,70
< void test_8_character_plaintext_results_in_3_chunks_the_last_one_with_a_trailing_space(void) {
---
> void
> test_8_character_plaintext_results_in_3_chunks_the_last_one_with_a_trailing_space
> (void)
> {
77c80,83
< void test_54_character_plaintext_results_in_7_chunks_the_last_two_with_trailing_spaces(void) {
---
> void
> test_54_character_plaintext_results_in_7_chunks_the_last_two_with_trailing_spaces
> (void)
> {

# indent.sh test_crypto_square.c
# diff test_crypto_square.c test_crypto_square.c~
67,70c67
< void
< test_8_character_plaintext_results_in_3_chunks_the_last_one_with_a_trailing_space
< (void)
< {
---
> void test_8_character_plaintext_results_in_3_chunks_the_last_one_with_a_trailing_space(void) {
80,83c77
< void
< test_54_character_plaintext_results_in_7_chunks_the_last_two_with_trailing_spaces
< (void)
< {
---
> void test_54_character_plaintext_results_in_7_chunks_the_last_two_with_trailing_spaces(void) {
```

And this leads of course to Travis fail.
Maybe make an exception this time about function names? Or even better edit canonical_data to be not *that* verbose?

wolf99:
We can probably get a bit creative with the test function names here (this is a strategy used in past cases of this problem).

Maybe work down these suggestions, testing indent locally after each adjustment and stop following them when indent stops complaining.

- Seeing as there are no test function names that specifically mention _non_-plain-text, we can probably lose the `plaintext` part.
- `characters` can probably be shortened to `chars`.
- The `_the` and `_a` can probably be dropped from `the_last_n_with_a_trailing_space`
- `results_in`  can be replaced with `gives`.

If indent still complains, report back here.

The idea here being to convey the same meaning to a casual reader while still being apparent to someone comparing the names to the canonical data.

exercises/crypto-square/src/example.c
```diff
@@ -0,0 +1,92 @@
...
+#include <stdlib.h>
+#include <ctype.h>
+#include <stdio.h>
+#include <math.h>
+
+/* mstrlen: get the length of input counting only alnums */
+static int mstrlen(const char *input)
+{
+   int i = 0;
+   while (*input) {
```
patricksjackson: You have two of these loops along input. I think you should be explicit about what you're testing against just to be a bit clearer. `*input != '\0'`

Additionally, I personally feel that writing these as a for loop is clearer. No need to change that if you don't want to though.

```
for (; *input != '\0'; input++) {
  if (isalnum(*input)) {
    i++;
  }
}
```

exercises/crypto-square/src/example.c
```diff
@@ -0,0 +1,92 @@
...
+   if (len_inp == 0)
+      return calloc(1, 1);
+
+   int cols, rows;
+   get_cols_and_rows(len_inp, &cols, &rows);
+
+   /* no need for extra row if single line */
+   int len_out = rows > 1 ? (rows + 1) * cols : rows * cols;
+
+   char *res = malloc(len_out + 1);
```
patricksjackson: You may want to be explicit about the allocation size here, instead writing the malloc as `malloc(sizeof(char) * (len_out + 1))`. You're using it correctly, but misusing malloc or calloc calls by mixing up what the arguments should be is a very common problem.

exercises/crypto-square/test/test_crypto_square.c
```diff
@@ -0,0 +1,106 @@
...
+void tearDown(void)
+{
+}
+
+void test_empty_text_res_in_an_empty_ciphertext(void)
+{
+   const char *input = "";
+   const char *expected = "";
+   char *res = ciphertext(input);
+   TEST_ASSERT_NOT_NULL(res);
```
patricksjackson: You don't need these NULL checks by the way. `TEST_ASSERT_EQUAL_STRING` checks for that with a pretty good error message, and `free(NULL)` is a valid call.

The error message:
```
test/test_ciphertext.c:19:test_empty_text_res_in_an_empty_ciphertext:FAIL: Expected '' Was NULL
```

patricksjackson:
My comments on the code are fairly minor, I think this looks fine. All of my comments here are suggestions and/or personal preference, besides the first one about `size_t`.

I think your solution is fine. I couldn't think of a better way to simplify the logic in `get_cols_and_rows` and `get_pos_cipher`, but it was good that you pulled those out of the main function.

Comments:

- You should use `size_t` instead of `int` for your array indices / lengths / etc. That's generally better practice. (There's some disagreement about whether to use `size_t` or `ptrdiff_t`, but that's really not too important, and the rest of this repo uses `size_t`. See isocpp/CppCoreGuidelines#1115).
- Your comments and variable names were a bit confusing at first, as I mixed up the rows/cols in the setup with the rows/cols in the output. I'm not sure how that can be more clear though.
- You should probably use `//` instead of `/* */` for comments as well.
- You also may want to use some more parentheses to clear up your statements. e.g `if ((*cols) * (*cols) == len)` instead of `if (*cols * *cols == len)`

exercises/crypto-square/src/example.c
vlzware: I can not really follow here - was not ```sizeof(char)``` defined in the standard to be 1? Or you mean to write it explicit for the sake of readability?

exercises/crypto-square/test/test_crypto_square.c
vlzware: I never checked the implementation of ```TEST_ASSERT_EQUAL_STRING``` just followed my own assumptions.

Nice catch - thanks!

vlzware:
Thanks for the review @patricksjackson

ryanplusplus:
👍

exercises/crypto-square/makefile
```diff
@@ -0,0 +1,25 @@
...
+memcheck: tests.out
+	@valgrind $(VFLAGS) ./tests.out
+	@echo "Memory check passed"
+
+clean:
+	rm -rf *.o *.out *.out.dSYM
+
+tests.out: test/test_crypto_square.c src/crypto_square.c src/crypto_square.h
+	@echo Compiling $@
+	@cc $(CFLAGS) src/crypto_square.c test/vendor/unity.c test/test_crypto_square.c -o tests.out -lm
```
wolf99: Ideally the `-lm` should have used a `LDFLAGS` variable here. As it is merged is it probably OK as-is for now but maybe bear in mind in the future :smiley:

PR #275, opened 2018-02-24
Implement exercise 'say'

vlzware:


exercises/say/src/example.c
```diff
@@ -0,0 +1,148 @@
...
+   const int below_20_len[] = {
+      3, 6, 6, 8, 8,
+      7, 7, 9, 8, 8
+   };
+
+   const char *below_100[] = {
+      "", "", "twenty", "thirty", "forty",
+      "fifty", "sixty", "seventy", "eighty", "ninety"
+   };
+   const int below_100_len[] = {
```
wolf99: These arrays may be a good candidate for introducing a typedef'd struct that holds both the number and the word. What do you think?

It may even be a good opportunity for you to learn what X-macros are and what they are often used for, if you haven't come across them yet. Not suggesting that X-macros should be introduced to the C track here though as they would likely be an exercise (or several) in themselves 😃

wolf99: Also, the arrays could have file-scope, rather than being part of the function.

exercises/say/src/example.c
```diff
@@ -0,0 +1,148 @@
...
+      0, 8, 7, 7
+   };
+
+   /* below_20's are like two times below_10's so this will work */
+   const int widest = 8;
+
+   /* digits count */
+   int digits = floor(log10(input)) + 1;
+
+   /* take care of spaces/hyphens, thousends and termination */
```
wolf99: typo: `thousends` -> `thousands`, note the `a`

exercises/say/makefile
```diff
@@ -0,0 +1,25 @@
...
+memcheck: tests.out
+	@valgrind $(VFLAGS) ./tests.out
+	@echo "Memory check passed"
+
+clean:
+	rm -rf *.o *.out *.out.dSYM
+
+tests.out: test/test_say.c src/say.c src/say.h
+	@echo Compiling $@
+	@cc $(CFLAGS) src/say.c test/vendor/unity.c test/test_say.c -o tests.out -lm
```
wolf99: `-lm` should probably be added by the variable `LDFLAGS` (doesn't currently exist). See [https://stackoverflow.com/questions/13249610/how-to-use-ldflags-in-makefile](https://stackoverflow.com/questions/13249610/how-to-use-ldflags-in-makefile) for an example of `LDFLAGS`

exercises/say/src/example.c
wolf99: ### X macros info if interested:
- [The late, great, Dr. Dobbs](http://www.drdobbs.com/the-new-c-x-macros/184401387)
- [embedded.com part 1](https://www.embedded.com/design/programming-languages-and-tools/4403953/C-language-coding-errors-with-X-macros-Part-1)
- [embedded.con part 2](https://www.embedded.com/design/programming-languages-and-tools/4405283/Reduce-C--language-coding-errors-with-X-macros---Part-2)
- [embedded.com part 3](https://www.embedded.com/design/programming-languages-and-tools/4408127/Reduce-C-language-coding-errors-with-X-macros--Part-3)

vlzware: It is indeed a lot of clutter like this :/
I have to admit I never thought about struct on this one, which is obviously a better, more integrated choice.
Also never heard of the 'X-macros'. It looks interesting - thanks for the links! Always glad to learn new stuff.

vlzware:
Thanks @wolf99 for the review and the suggestions.

exercises/say/src/example.c
vlzware: For this exercise I decided to go the 'struct'-route.
As you said maybe would be nice to have exercises with X-macros. I personally will gladly experiment more with them.

exercises/say/src/example.h
```diff
@@ -1,6 +1,12 @@
 #ifndef SAY_H
 #define SAY_H
 #include <stdint.h>
+#include <stddef.h>
+
+typedef struct {
```
wolf99: I think this can go inside `example.c`, right? No need for it to be exposed to the caller if `map` is only used internally.

exercises/say/test/test_say.c
```diff
@@ -15,7 +15,6 @@ void test_zero(void)
    char *ans = NULL;
    int res = say(0, &ans);
    TEST_ASSERT_EQUAL_INT(0, res);
-   TEST_ASSERT_NOT_NULL(ans);
```
wolf99: Why remove the tests for `NULL`?

exercises/say/src/example.h
wolf99: Also, I think we've been following the convention of giving type names a suffix of `_t`, e.g. `map_t`.
This is common in non-POSIX environments to make it clear what is a type name and what is a variable name (because a lot of editors don't do syntax highlighting for custom types).

vlzware: Completely agree.

exercises/say/test/test_say.c
vlzware: ```TEST_ASSERT_EQUAL_STRING``` checks already for ```NULL``` (see unity.c:1014).

This was first brought to my attention by @patricksjackson in [this PR](https://github.com/exercism/c/pull/274#pullrequestreview-99124342)

exercises/say/src/example.h
wolf99: FYI, the reason I mention non-POSIX environment is that names ending with `_t` are reserved in POISX. everywhere else it is a common idiom though.

exercises/say/src/example.c
```diff
@@ -0,0 +1,141 @@
...
+      int64_t order = pow(1000, (digits - 1) / 3);
+      tmp = input / order;
+      input -= tmp * order;
+
+      while (tmp) {
+
+         if (tmp / 100) {
+            int i = tmp / 100;
+            strcpy(head, below_10[i].name);
+            head += below_10[i].size;
```
ryanplusplus: Instead of `.size`, how about using `strlen`? This will be less fragile.

exercises/say/src/example.c
```diff
@@ -0,0 +1,141 @@
...
+static const int widest = 8;
+
+int say(int64_t input, char **ans)
+{
+   if (input < 0 || input > 999999999999 || ans == NULL)
+      return -1;
+
+   /* 'zero' is a corner case */
+   if (input == 0) {
+      *ans = malloc(5);
```
ryanplusplus: How about `*ans = "zero";` instead?

ryanplusplus: I think this will require you to change to `const char **ans`

exercises/say/src/example.c
```diff
@@ -0,0 +1,141 @@
...
+      }
+      strcpy(*ans, "zero");
+      return 0;
+   }
+
+   /* digits count */
+   int digits = floor(log10(input)) + 1;
+
+   /* take care of spaces/hyphens, thousands and termination */
+   *ans = malloc((widest + 1) * (digits + (digits + 1) / 3) + 1);
```
ryanplusplus: I'm a bit lost on what some of these terms mean. Could you extract some enumerations, `#define`s, or constants to make it clear what these mean?

exercises/say/src/example.c
```diff
@@ -0,0 +1,141 @@
...
+            head += below_10[tmp].size;
+
+            tmp = 0;
+
+            digits--;
+
+         } else if (tmp < 20) {
+            tmp -= 10;
+            strcpy(head, below_20[tmp].name);
+            head += below_20[tmp].size;
```
ryanplusplus: These two lines (or something very similar) are used quite often. How about extracting a helper function to reduce the duplication and make the code easier to read?

exercises/say/src/example.c
ryanplusplus: I use X-macros a lot at work. They add some cognitive overhead where they are used, but if you use them at the right time they are incredibly powerful.

ryanplusplus:
Looks pretty good. This is a tough exercise. I think the most important improvements that can be made are ones that will make this a bit easier to read and understand since it is a dense algorithm.

exercises/say/src/example.c
vlzware: You mean renaming 'size' to 'strlen' or really using 'strlen' to calculate the size?
My point of using the string-length pairs was to spare the size calculation iterations over predefined ```const char*```'s. I know there are small number of strings, but it hurts somehow to calculate their length over and over again. Also they are not supposed to be changed, as long as the names in English for numbers remain the same :)

exercises/say/src/example.c
```diff
@@ -0,0 +1,141 @@
...
+   /* take care of spaces/hyphens, thousands and termination */
+   *ans = malloc((widest + 1) * (digits + (digits + 1) / 3) + 1);
+   if (*ans == NULL) {
+      fprintf(stderr, "Memory error!\n");
+      return -1;
+   }
+
+   char *head = *ans;
+   while (input) {
+
```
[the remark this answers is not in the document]
vlzware: Well that was me experimenting with formatting in my editor.

exercises/say/src/example.c
vlzware: How can I then build the result if it is ```const```?

exercises/say/src/example.c
```diff
@@ -0,0 +1,141 @@
...
+   {"forty", 5}, {"fifty", 5}, {"sixty", 5}, {"seventy", 7},
+   {"eighty", 6}, {"ninety", 6}
+};
+
+static const map_t thousands[] = {
+   {"", 0}, {"thousand", 8}, {"million", 7}, {"billion", 7}
+};
+
+/* below_20's are like two times below_10's so this will work */
+static const int widest = 8;
```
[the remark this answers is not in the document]
vlzware: 17 is in the order of 10^1, meaning it occupies two digits. But 17 also gets represented by a single word. So it is shorter or equal than everything that is not '-teen' on the same order of magnitude. One of the longest 10^1 order digit is for example 77 which makes together length of 12 ('seventy' + 'seven'), occupying 2*widest = 16 bytes, leaving 4 free.
The value of 8 is actually the length of 'thousand' which would be the longest, non-complex word.

exercises/say/src/example.c
vlzware: I'll try to improve this.

exercises/say/src/example.c
vlzware: Of course. A very good suggestion.

vlzware:
Thanks @ryanplusplus for the code review.

exercises/say/src/example.c
ryanplusplus: It is a pointer to const data and the compiler will replace a string literal with a `const void *` that points to the constant string data.

ryanplusplus: I think I understand now what you're saying: the client needs to be able to `free` this memory so we can't declare it as `const`. I agree with you.

ryanplusplus:
👍

exercises/say/src/example.c
vlzware: Good point.
My question was how to use ```strcpy``` when the destination string is a ```const```? Build the result in a modifiable char and then cast the return to ```const char```?

ryanplusplus: Well, in this case I was suggesting that you _not_ use `strcpy` but instead just set `*ans` to a string literal. This would work except that it would not be possible to free the string.

vlzware: Got it - thanks for the clarification!

PR #276, opened 2018-02-25
Implement exercise 'pig-latin'

vlzware:
Although the specification says nothing about it, I took the decision to just copy not just the spaces, but all non-words to the output; and also make everything lowercase.

Let me know what you guys think about this.

I am learning a lot around here, so don't judge me too hard if I forgot again something.

### DIFF
--- a/config.json
+++ b/config.json
@@ -544,6 +544,19 @@
       ],
       "unlocked_by": "atbash-cipher",
       "uuid": "23c2a169-eea2-4636-af62-4a77a4dc3d7b"
+     },
+     {
+      "core": false,
+      "difficulty": 4,
+      "slug": "pig-latin",
+      "topics": [
+        "algorithms",
+        "functions",
+        "strings",
+        "control_flow_if_statements"
+      ],
+      "unlocked_by": "hamming",
+      "uuid": "62f539e7-f682-412d-bf86-38503e7cebca"
     }
   ],
   "foregone": [],
--- a/exercises/pig-latin/src/example.c
+++ b/exercises/pig-latin/src/example.c
@@ -1,0 +1,121 @@
+#include "pig_latin.h"
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <ctype.h>
+#include <stdbool.h>
+
+static bool is_vowel(const char a)
+{
+   return a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u';
+}
+
+static bool starts_with_vowel(const char a, const char b)
+{
+   return is_vowel(a) || (a == 'x' && b == 'r') || (a == 'y' && b == 't');
+}
+
+static void count_words_and_chars(const char *input, size_t * words,
+                                  size_t * chars)
+{
+   bool in_word = false;
+   while (*input) {
+      (*chars)++;
+      if (isalpha(*input++)) {
+         *words += in_word ? 0 : 1;
+         in_word = true;
+      } else {
+         in_word = false;
+      }
+   }
+}
+
+static void mstrncpy(char *dst, const char *src, size_t n)
+{
+   while ((n--) && (*dst++ = tolower(*src++))) ;
+}
+
+static void append_ay(char *res)
+{
+   *res = 'a';
+   *(res + 1) = 'y';
+}
+
+static void process_pig_logic(char *dst, const char *src, size_t word_len)
+{
+   append_ay(dst + word_len);
+
+   const char a = tolower(*src);
+   const char b = a ? tolower(*(src + 1)) : 0;
+
+   if (starts_with_vowel(a, b)) {
+      mstrncpy(dst, src, word_len);
+      return;
+   }
+
+   size_t i = 0;
+   while (src[i]) {
+      const char a = tolower(src[i]);
+      const char b = tolower(src[i + 1]);
+
+      if (is_vowel(a))
+         break;
+
+      if (b == 'y') {
+         i++;
+         break;
+      }
+
+      if (a == 'q' && b == 'u') {
+         i += 2;
+         break;
+      }
+      i++;
+   }
+   mstrncpy(dst, src + i, word_len - i);
+   mstrncpy(dst + (word_len - i), src, i);
+}
+
+char *translate(const char *phrase)
+{
+   if (phrase == NULL)
+      return NULL;
+
+   size_t words = 0, chars = 0;
+   count_words_and_chars(phrase, &words, &chars);
+
+   if (chars == 0)
+      return NULL;
+
+   /* every word gets 'ay' at the end */
+   const size_t res_len = chars + words * 2;
+
+   char *res = malloc(res_len + 1);
+   if (res == NULL) {
+      fprintf(stderr, "Memory error!\n");
+      return NULL;
+   }
+
+   if (words == 0)
+      return strcpy(res, phrase);
+
+   char *tmp = res;
+   while (words--) {
+
+      /* just copy non-alphas */
+      while (*phrase && !isalpha(*phrase))
+         *tmp++ = *phrase++;
+
+      size_t word_len = 0;
+      while (isalpha(phrase[word_len]))
+         word_len++;
+
+      process_pig_logic(tmp, phrase, word_len);
+
+      tmp += word_len + 2;
+      phrase += word_len;
+   }
+   res[res_len] = '\0';
+
+   return res;
+}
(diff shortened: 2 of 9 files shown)


exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,121 @@
...
+{
+   return a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u';
+}
+
+static bool starts_with_vowel(const char a, const char b)
+{
+   return is_vowel(a) || (a == 'x' && b == 'r') || (a == 'y' && b == 't');
+}
+
+static void count_words_and_chars(const char *input, size_t * words,
```
wolf99: I'd say split this into `count_words()` and `count_chars()` - that would align better with the single responsibility principle and would make the parameters simpler. True, it would result in two iterations through the input but we're not expecting to handle massive strings.

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,121 @@
+#include "pig_latin.h"
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <ctype.h>
+#include <stdbool.h>
+
+static bool is_vowel(const char a)
```
wolf99: maybe use `letter` instead of `a` as the parameter name - C does prefer brevity and single letter parameter names are used in the standard library a lot, but A) the standard library is nearly _too_ brief (they are looking at efficiency of code size , as applications programmers we don't have that concern at this level) B) `a` is a confusing name when talking about vowels (we have to say ` a == 'a'` which works but can be confusing).

The main idea when deciding such things is to think what would be the least mental work for someone reading the code to understand when the don't already know what's going on.

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,121 @@
...
+{
+   return a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u';
+}
+
+static bool starts_with_vowel(const char a, const char b)
```
wolf99: Same suggestion for parameter names here, maybe `first_letter, second_letter` or something

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,121 @@
...
+         in_word = true;
+      } else {
+         in_word = false;
+      }
+   }
+}
+
+static void mstrncpy(char *dst, const char *src, size_t n)
+{
+   while ((n--) && (*dst++ = tolower(*src++))) ;
```
wolf99: Nice!

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,121 @@
...
+         i += 2;
+         break;
+      }
+      i++;
+   }
+   mstrncpy(dst, src + i, word_len - i);
+   mstrncpy(dst + (word_len - i), src, i);
+}
+
+char *translate(const char *phrase)
```
wolf99: IMHO, this should not need to allocate memory - instead just alter the input, maybe make the input parameter `char phrase[]`. The returned length is the same as the input length so just changing the input makes memory handling a lot simpler.

Personally I'm quite averse to making functions that allocate memory but rely on the caller to `free` it. In real life, the function documentation would mention that it altered the input and the caller could do the allocation and make a copy if it needs the original to remain intact. This again looks to the single responsibility principle, if you can, make your class/function/etc take care of one thing as long as the functionality doesn't _need_ memory allocation then leave it to the caller to handle (in fact the caller might want to call some other code or library to do that for it).

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,121 @@
...
+   *res = 'a';
+   *(res + 1) = 'y';
+}
+
+static void process_pig_logic(char *dst, const char *src, size_t word_len)
+{
+   append_ay(dst + word_len);
+
+   const char a = tolower(*src);
+   const char b = a ? tolower(*(src + 1)) : 0;
```
wolf99: These could also change to `first_letter, second_letter`

wolf99:
Looks good, just a few small suggestions on naming mainly

config.json
```diff
@@ -530,6 +530,18 @@
...
+    {
+      "core": false,
+      "difficulty": 4,
+      "slug": "pig-latin",
+      "topics": [
+        "algorithms",
+        "functions",
+        "strings",
+        "control_flow_if_statements"
+      ],
```
wolf99: This needs an `unlocked_by` value as `core` is `false`

exercises/pig-latin/src/example.c
vlzware: The whole point of this function was to spare one iteration through the input. If possible, I would prefer to keep it this way.

config.json
vlzware: Nice catch! It seems I made the same mistake with 'bracket-push' (I made a PR with correction already). It is sad, but I still cannot find a way to automate the generation/updating of 'config.json' and copy/paste errors occur.

exercises/pig-latin/src/example.c
vlzware: I suppose it is a matter of taste, because I personally find it easier for my eyes when functions/names are not too verbose. Anyway, I changed this to your suggestion.

vlzware:
Thanks @wolf99 for the code review!

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,123 @@
...
+      if (isalpha(*input++)) {
+         *words += in_word ? 0 : 1;
+         in_word = true;
+      } else {
+         in_word = false;
+      }
+   }
+}
+
+static void mstrncpy(char *dst, const char *src, size_t n)
```
ryanplusplus: What does the `m` mean?

exercises/pig-latin/src/example.c
ryanplusplus: I'm fine with the short variable name if you use an idiomatic `c` (for `char`).

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,123 @@
...
+#include <ctype.h>
+#include <stdbool.h>
+
+static bool is_vowel(const char letter)
+{
+   return letter == 'a' || letter == 'e' || letter == 'i' || letter == 'o'
+       || letter == 'u';
+}
+
+static bool starts_with_vowel(const char letter_a, const char letter_b)
```
ryanplusplus: I understand your desire to keep the variable names brief, but as-is this function is quite cryptic. It's not clear to me from the naming or context that `letter_a` and `letter_b` refer to the first and second letters in a string. I actually think it would be more clear to just pass a `const char *` and use indexing to get the first two letters.

It's also quite odd to see a function called `starts_with_vowel` that has three `true` paths where only one of those is because the first letter is a vowel. Without looking at tests I'm confused as to what this function really determines.

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,123 @@
...
+
+      if (is_vowel(letter_a))
+         break;
+
+      if (letter_b == 'y') {
+         i++;
+         break;
+      }
+
+      if (letter_a == 'q' && letter_b == 'u') {
```
ryanplusplus: This is a dense solution so any trick you can use to make this easier to read is appreciated. In this case, how about using a helper like `starts_with(&src[i], "qu")`? I think that this could be helpful in a few places.

exercises/pig-latin/src/example.c
vlzware: Named after strncpy because it does the same up to the 'tolower' part. The ```m``` (my) was to avoid confusion  with the "real" strncpy. I suppose now it is still a confusing name ... :/

exercises/pig-latin/src/example.c
vlzware: Completely agree.
Will give my best to make it more sane.

vlzware:
Thanks @ryanplusplus for the code review.
I tried to improve as much as I could.

exercises/pig-latin/src/example.c
```diff
@@ -0,0 +1,128 @@
...
+
+   size_t words = 0, chars = 0;
+   count_words_and_chars(phrase, &words, &chars);
+   if (chars == 0)
+      return NULL;
+
+   /* every word gets 'ay' at the end */
+   const size_t res_len = chars + words * 2;
+
+   char *res = malloc(res_len + 1);
```
wolf99: What does `res` mean? Can we use a more descriptive name in its place?

vlzware: Sure. This was a stupid name indeed.

wolf99:
LGTM 👍

PR #277, opened 2018-02-26
Update config.json

vlzware:
- add 'unlocked_by' in 'bracket-push'
- add 'unlocked_by' in 'luhn'
- thanks to @wolf99 for pointing this out

### DIFF
--- a/config.json
+++ b/config.json
@@ -518,6 +518,7 @@
         "strings",
         "algorithms"
       ],
+      "unlocked_by": "pascals-triangle",
       "uuid": "8c0762e6-702e-4733-a0a1-65099e2a46f7"
     },
     {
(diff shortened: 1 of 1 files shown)


wolf99:
Goo catch to realise this off the back of my comment on the other exercise. However `word-count` is probably not an ideal exercise to unlock this as `word-count` itself is also has `core` set to `false`. To date we have been trying to ensure that exercises only nest to one level. This would nest `bracket-push` at the second level. 

If this leaves you unsure where to put this exercise, do reply here and we can discuss it :smiley:

vlzware:
Thanks @wolf99 
In the mean time I spotted the same problem by 'luhn'. I will update the PR today.
The visualizer is cool, I'll remember to use it next time when I am updating the json.

wolf99:
Good catch on `luhn` :+1:

PR #278, opened 2018-02-27
Implement exercise 'run-length-encoding'

vlzware:
The code to get the length and to fill the result are almost the same for both encode and decode, so I decided to combine them to reduce redundancy.

### DIFF
--- a/exercises/secret-handshake/src/example.c
+++ b/exercises/secret-handshake/src/example.c
@@ -3,6 +3,8 @@
 #include <stdio.h>
 #include <stdbool.h>
 
+#define ARRAY_SIZE(arr) (sizeof(arr)/sizeof(arr[0]))
+
 typedef struct {
    const char *cmd;
    const size_t num;
--- a/exercises/secret-handshake/src/example.h
+++ b/exercises/secret-handshake/src/example.h
@@ -2,8 +2,6 @@
 #define SECRET_HANDSHAKE_H
 #include <stddef.h>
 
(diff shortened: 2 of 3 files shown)


exercises/run-length-encoding/src/example.c
```diff
@@ -0,0 +1,130 @@
...
+ */
+static size_t enc_helper(char *encoded, const char *text,
+                         const enum rw_mode mode)
+{
+   size_t enc_len = 0;
+   while (*text) {
+      if (isdigit(*text)) {
+         fprintf(stderr, "Digits in the plaintext are not supported!\n");
+         return 0;
+      }
```
ryanplusplus: Is there a test for this assertion?

ryanplusplus:
Overall this looks good to me. One way that you can make problems like this (where you need to get the size of the solution and then generate it) easier for yourself is by using `realloc`. For example, you could create a simple dynamic string helper that lets you add a character at a time and `realloc`s when it runs out of space.

exercises/run-length-encoding/src/example.c
vlzware: Nope. I added this for the sake of completeness. 

Following [your comment in the other PR](https://github.com/exercism/c/pull/281#pullrequestreview-101385711), will remove this too.

vlzware:
I always tried to stay away from 'realloc' fearing of some hypothetical overhead, which is only my own assumption, but maybe it is time to finally learn to work with it, or at least try it.

vlzware:
Great - screwed it up again with git

PR #279, opened 2018-02-28
Implement exercise 'secret-handshake'

vlzware:


wolf99:
Looks pretty good! 👍

exercises/secret-handshake/src/example.c
```diff
@@ -0,0 +1,58 @@
...
+};
+
+static const size_t CMD_COUNT = ARRAY_SIZE(cmds);
+static const size_t CMD_REVERSE = 16;
+
+/* bitcount: count the number of set bits in n */
+static size_t bitcount(size_t num)
+{
+   size_t count;
+   for (count = 0; num; num = num & (num - 1), count++) ;
```
wolf99: Won't this just return the total number of bits (`count++` is executed on every iteration) rather than the number of bits that are 1?

ryanplusplus:
👍

PR #280, opened 2018-03-02
Implement exercise 'flatten-array'

vlzware:
The actual difficulty in this exercise is not in the short recursion function, but in understanding the structures in the tests. So I tried to make our non-rectangular "arrays" as human readable as possible.

Therefore:
- used a predefined fixed-length rectangular array - this allows to define the "nested arrays" in human readable, not reversed sequence - 0,1,2... each pointing to the next (already existing) array member;
- used a set of macros to be able to represent the "brackets" and the "elements" so close as possible to a typical representation like: ```[0, 1, [2, [3]], 4]```

Some more notes:
- I skipped brackets-matching checking for simplicity;
- it seems that initializing array of structures gets incorrectly identified as warning by gcc, so I disabled this check (see also this [stackoverflow link](https://stackoverflow.com/questions/11551986/gcc-missing-braces-around-initializer)).

I had a lot of fun with this. Attached here a ddd visualization of the structure from the third test "test_flatten_5_level_nesting" which is ```[0, 2, [[2, 3], 8, 100, 4, [[[50]]]], -2]``` :

ddd_flatten_5_level_nesting

...
PS: I wanted once again to thank you all (but especially @wolf99 and @ryanplusplus) for spending so much of your time reviewing my code and improving my programming and representation skills. I've never been mentored by professional programmers before.


wolf99:
Bear with me, it might take a little longer to parse this one 😃. 

Firstly though, as the test file has things in it that rely on definitions in the header (`NONE`, `UNDEF`, etc), then they should be moved from `example.h` to `flatten_array.h` so that they are provided to the student (it does not necessarily need to include other items from `example.h`, such as the function declaration.

Also, you're welcome in regards to the reviewing. You contributions help continue track development and also further our code reading and reviewing skills, which are as important to programmers (most pro programmers end up having to read other peoples code lots of the time) as regular development skills are. So thank you 👍

exercises/flatten-array/src/example.h
```diff
@@ -0,0 +1,19 @@
...
+
+#define ARRAY_SIZE(arr) (sizeof(arr)/sizeof(arr[0]))
+#define NONE 0
+
+enum item_type { UNDEF, null, ELEMENT, ARRAY };
+struct item {
+   int value;
+   struct item *next;
+   struct item *array_elements;
+   enum item_type type;
```
wolf99: By convention, members of structs are often declared largest to smallest in terms of type bit-size. This allows the compiler to naturally lay out the struct in an in the most efficient way it can (without doing packing) due to memory alignment and word size. It is not important w.r.t. to functionality, but may be interesting for you. Following that idiom here would result in :

```c
struct item {
    struct item *next;
    struct item *array_elements;
    enum item_type type;
    int value;
};
```

exercises/flatten-array/src/example.h
```diff
@@ -0,0 +1,19 @@
...
+};
+
+typedef struct item item_t;
```
wolf99: If you put this before the struct definition then you can use `item_t *` in the struct instead of `struct item *`

exercises/flatten-array/src/example.c
```diff
@@ -0,0 +1,48 @@
...
+      return NULL;
+
+   flat_array_pos = 0;
+   parse_array(head, READ);
+   const size_t flat_array_len = flat_array_pos;
+
+   if (flat_array_len == 0)
+      return 0;
+
+   flat_array = malloc(sizeof(int) * flat_array_len);
```
wolf99: When using the `sizeof` operator it is generally better to apply it to a variable rather than directly to the type (e.g. `malloc(sizeof(*flat_array) * flat_aray_len);`) as this decouples the _implementation_ of `flatten()` from the types (i.e. there will be less places that code would need changing if the type of `flat_array` needed changing).

wolf99:
Can you explain what the macros at the top of test file are for, why they are necessary etc? I'm struggling to see exactly what they do.

wolf99:
I'm ... not super convinced that macros are the way to go with the arrays in the test file. I understand from your initial comment why you went that route but I'm not sure it makes things easier to understand (also, I have a personal aversion to non-trivial macros). It's late here so I don't have brain power to suggest alternatives at the moment but am interested in discussion on this.

exercises/flatten-array/makefile
```diff
@@ -0,0 +1,28 @@
+CFLAGS  = -std=c99
+CFLAGS += -g
+CFLAGS += -Wall
+CFLAGS += -Wextra
+CFLAGS += -pedantic
+CFLAGS += -Werror
+
+# https://stackoverflow.com/questions/11551986/gcc-missing-braces-around-initializer
```
wolf99: Date this and note the current SO license, if you could. Might also be worth including the bug report link  to cover the case that (gasp!) SO is gone/down at some point in the future.

vlzware: I did not forgot your suggestion from the 'bracket-push' commit, but in this case I did not copy anything from stackoverflow.

Will change the comment in the makefile to point directly to bugzilla.

vlzware:
Thanks @wolf99 for the code review!

vlzware:
About the approach in the testfile: it is all about making the representation of the nested structure as readable as possible.
- According to the problem description there are three cases for an element in the structure to be: an integer - I called this 'ELEMENT' (like in 'array element'), an array or 'null'. For the sake of completeness I added a fourth one: 'undefined'. This could be spared, but I think it adds to the readability;
- Declaring everything as a variable has the drawback that because of the pointing to each other either the declaring order should be reversed or the variables should be written two times - declaration and assignment , so I used the predefined ```nested_array``` to avoid this and write each element only one time and in the expected order 0,1,2..;
- Still this was not even remotely close to a representation like ```[1, [2, 3, [[[4]]], 5]]```. Therefore I added the macros. Sure, they look confusing at first time in the file head, but looking at the declarations using them later in the functions should convince one that they add *a lot* to the readability. Adding a new case is then as easy as describing almost in plain english what one sees in the problem description - like 'open bracket, element with value 1, open bracket, element with value 2...' and so on.

Descriptions:
- ```NESTED_ARRAY_SEQUENCE_START``` defines some variables for the current positions in each level and for error checking;
- ```NESTED_ARRAY_SEQUENCE_END``` is there basically just for the error checking;
- ```OPEN_BRACKET``` defines an array type and goes one level deeper;
- ```CLOSE_BRACKET``` goes back upper one level;
- ```ITEM(x)``` and ```NULL_ITEM``` declare respectively array element, setting its int-value and an pseudo-empty 'null' element;
- ```ADVANCE_POSITION``` (takes care of watching the column in the current row/level), ```NEXT_LEVEL``` (watching the current row/level), ```POINT_TO_NEXT``` (adjusting pointers) and ```NO_ELEMENTS``` (means the element is not an array) - are all shorthands to avoid code repetition later in the declarations.

ryanplusplus:
Sorry it has taken me so long to take a deep look at this.

I think the main problem (and the one that @wolf99 is also struggling with) is that by looking at the test file it is not clear what the representation is. The macros you've made make it very clear what you _intend_ to create, but in the test the student also needs to ascertain what the actual representation in C data structure is (and thus what they need to create). So while I think that they will largely be able to understand what sort of data you're representing, I don't think it will be easy for them to understand _how_ you're actually representing it.

For this reason I'm not sure that this is a good fit. I can imagine that a lot of people will reach this point in the track and either give up or skip this problem because it is not as enjoyable to try to understand the problem as it is to solve it.

@wolf99 -- What are your thoughts?

@vlzware -- Do you agree? Is there anything we can do to make this clearer and more concrete for someone trying to solve this problem?

wolf99:
I think @ryanplusplus has framed the issue well. Unfortunately I haven't had the time required to come back and look at this fully. The main part of my issue with it is not that the macros is in anyway faulty but that they make comprehension harder than it might ought to be. 

For a student, comprehending the test file is the way in which they understand what is required of the implementation's interface and how it should behave, at a more detailed level than is described in the README. Here the macros hinder that comprehension such that a student requires extra time is just to understand what they need to do.

Interested to hear your thoughts on this point @vlzware, and any alternative test structure you might consider.

PR #281, opened 2018-03-04
Implement exercise 'minesweeper'

vlzware:


exercises/minesweeper/test/test_minesweeper.c
```diff
@@ -0,0 +1,171 @@
...
+   const size_t rows = ARRAY_SIZE(expected);
+   char **actual = annotate(minefield, rows);
+   TEST_ASSERT_EQUAL_STRING_ARRAY(expected, actual, rows);
+   free_all(actual, rows);
+}
+
+void test_annotate_cross(void)
+{
+   TEST_IGNORE();
+   const char *minefield[] = { "  *  ", "  *  ", "*****", "  *  ", "  *  " };
```
wolf99: Can't recall if indent will allow, but for readability you could maybe format the array initialisations like:
```c
const char *minefield[] = {
        "  *  ", 
        "  *  ", 
        "*****", 
        "  *  ", 
        "  *  " 
};
```

exercises/minesweeper/test/test_minesweeper.c
```diff
@@ -0,0 +1,171 @@
...
+   char **actual = annotate(minefield, rows);
+   TEST_ASSERT_NULL(actual);
+}
+
+void test_annotate_no_columns(void)
+{
+   TEST_IGNORE();               // delete this line to run test
+   const char *minefield[] = { "" };
+   const char *expected[] = { "" };
+   const size_t rows = ARRAY_SIZE(expected);
```
wolf99: Just spotted this problem with using the array size macro in the test file (may apply to previous exercises also!): The macro is defined in the implementation header yet is not provided to the student. Generally I would think that things not directly related to the implementation (e.g. enums) that are used in the test file should be defined in the test file and let the student be concerned with the implementation rather than the instrumentation. This particular macro is used in both implementation and test, but only by coincidence for _this particular implementation_, students OTOH may find other implementations that do not require such a macro. To resolve this the macro should be defined in the test and separately in the implementation's source file.

wolf99: In fact, I just noticed, in this implementation the macro is not used at all in the source and so doesn't need to be defined there at all, just in the test file

exercises/minesweeper/test/test_minesweeper.c
vlzware: You are right, formatted looks much better and makes more sense as a 'minefield'.

exercises/minesweeper/test/test_minesweeper.c
vlzware: Fully agree, will check and fix them all.

exercises/minesweeper/src/example.c
```diff
@@ -0,0 +1,66 @@
...
+   if (annotated == NULL) {
+      fprintf(stderr, "Memory error!\n");
+      return NULL;
+   }
+
+   for (size_t i = 0; i < rows; i++) {
+      annotated[i] = malloc(cols + 1);
+      if (annotated[i] == NULL) {
+         fprintf(stderr, "Memory error!\n");
+         free(annotated);
```
ryanplusplus: Technically you have a memory leak here. If you have a `malloc` failure partway through here, you will only free the top level `annotated` array, but not all of the previous rows. One way to fix this without making your solution much more complicated is to allocate all of the rows with one `malloc` (all rows in one string) and then set the pointers from `annotated` for each row by indexing into the combined row string.

So the pointers in `annotated` would be set like this for a 3x3 mine field:

```
annotated = [ _ , _ , _ ]
                        |    \     \
                       /      |      \
                     /        |        \ 
all rows = [ |, _, _, |, _, _, |, _, _ ]
```

exercises/minesweeper/src/example.c
```diff
@@ -0,0 +1,66 @@
...
+         return NULL;
+      }
+      strcpy(annotated[i], minefield[i]);
+   }
+
+   for (size_t i = 0; i < rows; i++) {
+      for (size_t k = 0; k < cols; k++) {
+         if (minefield[i][k] == '*')
+            set_fields(annotated, rows, cols, i, k);
```
ryanplusplus: `set_fields` is not very clear for me. How about `increment_neighbors` or something?

exercises/minesweeper/src/example.c
```diff
@@ -0,0 +1,66 @@
...
+         else if (minefield[i][k] != ' ' && !isdigit(minefield[i][k]))
+            fprintf(stderr, "Invalid input: \"%c\" at "
+                    "row %ld column %ld.\n", minefield[i][k], i, k);
```
ryanplusplus: Is there a test for this? My preference would be to keep the solution to a minimum and not perform extra checks that add complexity.

In a production system you may need these checks (or mine field may come from a trusted input), but I think it is more important to keep the details related to the problem for this educational tool.

exercises/minesweeper/src/example.c
```diff
@@ -0,0 +1,66 @@
...
+{
+   if (minefield[row][col] == '*')
+      return;
+
+   const char c = minefield[row][col];
+   minefield[row][col] = (c == ' ') ? '1' : c + 1;
+}
+
+static void set_fields(char **minefield, const size_t all_rows,
+                       const size_t all_cols, const size_t row,
```
ryanplusplus: Would `all_rows` and `all_cols` be better named as `row_count` and `col_count`? I believe that's how they are used here.

exercises/minesweeper/src/example.c
vlzware: Lol! You are completely right - I overlooked this.
Thanks for the suggestion, I like it - will take care of this today.

exercises/minesweeper/src/example.c
vlzware: No tests for this. I tried to make it more 'bullet-proof' but you have right.

ryanplusplus:
Looks great 👍

PR #283, opened 2018-03-05
Implement exercise 'wordy'

vlzware:


exercises/wordy/src/example.c
```diff
@@ -0,0 +1,75 @@
...
+      (*str)++;
+   }
+   while (*(*str) && isdigit(*(*str))) {
+      int_len++;
+      (*i) = (*i) * 10 + *(*str) - '0';
+      (*str)++;
+   }
+   (*i) *= sign;
+   return (int_len != 0);
+}
```
ryanplusplus: Is there a reason to not use something like `sscanf` here?

exercises/wordy/src/example.c
```diff
@@ -0,0 +1,75 @@
+#include "wordy.h"
+#include <string.h>
+#include <ctype.h>
+
+typedef struct {
+   const char *str;
+   const size_t len;
+
```
ryanplusplus: Can you remove this blank line?

exercises/wordy/src/example.c
```diff
@@ -0,0 +1,75 @@
...
+#define CMD(a, b, c) \
+      if (strncmp(b, question, c) == 0) { \
+         question += c; \
+         if (!get_int(&question, &tmp)) \
+            return false; \
+         *result a##= tmp; \
+      } else
+      COMMANDS
+#undef CMD
```
ryanplusplus: Nice use of XMacros. Another option would be to select an via function pointer. Then you could grab the second operand and complete the operation by passing both operands when invoking the function pointed to by the function pointer. This way you wouldn't need to use XMacros to repeat the part where you grab the second operand. I don't know that this would be an improvement, though, since this is relatively straightforward.

ryanplusplus:
Looks great overall. I'm mostly just curious why you didn't use `sscanf` because it will make that part of your solution much simpler if you can.

exercises/wordy/src/example.c
vlzware: Well, no meaningful reason ... :)
1. It was fun to implement it myself.
2. Also, until now I did not know how to increment the string pointer after scanning, but your comment motivated me to search and learn about this.

vlzware:
Thanks @ryanplusplus for the code review!

ryanplusplus:
Looks great with `sscanf`, thanks for the update!

PR #284, opened 2018-03-06
Implement exercise 'run-length-encoding'

vlzware:


vlzware:
This is a recreation of [this PR](https://github.com/exercism/c/pull/278) with the [extra check removed](https://github.com/exercism/c/pull/278#pullrequestreview-101382586). Sorry for the git mess.

PR #285, opened 2018-03-07
Implement exercise 'diamond'

vlzware:


vlzware:
The canonical data states this can be also implemented using 'property-based' tests, which is also new and interesting concept for me. Would it make sense to implement such tests here without using external libraries? Or this would lead to over-bloated and confusing tests?
As far as I could understand the concept of property-based tests, I can define the properties according to the description and then test random chunks of the output for matching.

ryanplusplus:
@vlzware I'm not opposed to property-based testing if you can find a way to make it clean. I've only ever seen it done using 1) libraries that help and 2) languages that can be more concise/expressive than C, so it may not be easy.

exercises/diamond/test/test_diamond.c
```diff
@@ -0,0 +1,155 @@
...
+{
+}
+
+static void free_all(char **diamond)
+{
+   free(diamond[0]);
+   free(diamond);
+}
+
+void test__degenerate_case_with_a_single_a_row(void)
```
ryanplusplus: I noticed that in this test file all tests are prefixed with `test__` instead of `test_`. Can you update this?

ryanplusplus:
@vlzware sorry for the delay. Everything looks good except the small issue with test names 👍

exercises/diamond/test/test_diamond.c
vlzware: Indeed. Sorry about this.

ryanplusplus:
LGTM

wolf99:
LGTM too, thanks @vlzware

PR #287, opened 2018-03-08
Implement exercise 'binary-search-tree'

vlzware:


vlzware:
This time used the approach (suggested by @ryanplusplus) of one time parsing and using 'realloc' if needed.

Interested as always to hear your opinions.

exercises/binary-search-tree/test/test_binary_search_tree.c
```diff
@@ -0,0 +1,217 @@
...
+static void free_tree(node_t * tree)
+{
+   if (tree == NULL)
+      return;
+   free_tree(tree->left);
+   free_tree(tree->right);
+   free(tree);
+}
+
+void test_data_data_is_retained(void)
```
wolf99: Is purposely `data` repeated twice in this function name?

wolf99:
It only occurs to me now, but I know some of the other exercises stay away from typedef-ing struct types. This is something solely based on opinions of style and clarity so it does not affect this particular implementation but I wonder if we could add something to the style guide around this? @ryanplusplus,  @patricksjackson, @RealBarrettBrown @vlzware 

1. We could continue to go for any and all styles, or
2. We could decide a preference for one or more of the following

```c
/* option 2A - no typedef, struct usage is explicit at all occurances */
struct foo { };

/* option 2B - typedef'd with separate definition, struct may still be referenced directly */
struct bar { };
typedef struct bar bar_t;

/* option 2C - typedef defines struct, can only use baz_t */
typedef struct { } baz_t;

/* option 2D - only shown for completeness, is a bit redundant and stylistically unclear */
typedef struct qux { } qux_t;
```

@vlzware This should not affect this implementation, if discussion prefers something different to what you've added, then later changes to update this could be made (as ideal candidates for good-first-patch labeling).

exercises/binary-search-tree/test/test_binary_search_tree.c
vlzware: After looking at as many as possible canonical-data.json I decided to stick to the following format when generating the test-functions:
```
   "test_" + property + description
```
which in this case led to the repeating 'data'.

exercises/binary-search-tree/test/test_binary_search_tree.c
```diff
@@ -0,0 +1,217 @@
...
+
+void setUp(void)
+{
+}
+
+void tearDown(void)
+{
+}
+
+static void free_tree(node_t * tree)
```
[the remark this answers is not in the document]
vlzware: The already implemented exercises declare them after ```setUp()``` and ```tearDown()```

exercises/binary-search-tree/test/test_binary_search_tree.c
wolf99: Great, thanks for the confirmation :+1:

exercises/binary-search-tree/test/test_binary_search_tree.c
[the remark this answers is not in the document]
wolf99: Ah, fair enough. I may need to update the style guide to conform to this, rather than the other way around :smile:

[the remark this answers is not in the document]
vlzware: You are right - it does say the helpers comes before ```setUp()``` and ```tearDown()```. I read this a month ago or so, but then got carried away looking at the other exercises... Now, as your comment goes, not that bad after all :)

wolf99:
LGTM

exercises/binary-search-tree/src/example.c
```diff
@@ -0,0 +1,109 @@
...
+      return NULL;
+   }
+
+   mem_ok = true;
+   parsed_len = 0;
+   parse_tree(head);
+   if (!mem_ok)
+      return NULL;
+
+   qsort(parsed_tree, parsed_len, sizeof(int), compare);
```
ryanplusplus: This gives the correct answer, but I think the intent is to use the binary search tree to do the "sort". Basically, if you walk the tree using depth first search and always start with the left (lesser) branch then you will visit all of the nodes in descending order.

ryanplusplus: So I think that if you update `parse_tree` you will be able to do this without using `qsort`. I would probably also rename `parse_tree` to `enumerate_tree` or `walk_tree`.

ryanplusplus:
Looks good once the sorting is worked out 👍

exercises/binary-search-tree/src/example.c
vlzware: A very good point!

ryanplusplus:
:+1:

PR #288, opened 2018-03-15
Correct guide on order of test helper functions

wolf99:
Test helper functions should come *after* `setup()` and `tearDown()`.


ryanplusplus:
👍

PR #289, opened 2018-03-17
Update 'palindrome-products'

vlzware:
- improve the logic in the palindrome-checking function

### DIFF
--- a/exercises/palindrome-products/src/example.c
+++ b/exercises/palindrome-products/src/example.c
@@ -2,64 +2,34 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <limits.h>
+#include <stdbool.h>
 
-static int palindrome(int n);
-static int addfactors(factor_t ** p, int i, int k);
-static void free_ll(struct factors *p);
-
-product_t *get_palindrome_product(int from, int to)
+__attribute__ ((optimize("-O3")))
+static bool palindrome(int n)
 {
-   product_t *res = malloc(sizeof(product_t));
-   if (res == NULL) {
-      fprintf(stderr, "Memory error!\n");
-      return NULL;
+   /* 0 is palindrome */
+   if (n == 0)
+      return true;
+
+   /* negative can not be palindrome */
+   if (n < 0)
+      return false;
+
+   int tail;
+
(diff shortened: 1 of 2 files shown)


vlzware:
It troubled me since then, because this was my first commit and also the reason for the whole Travis build to be so painfully slow.
It looks that the palindrome-checking is the bottleneck (also confirmed with gprof), so here it goes:
- Checking the half of the number against the other (reversed) half improved the execution on my pc exactly twice;
- Declaring ```n``` and ```tail``` to be register variables improved even more - I read about them in K&R, but then future googling revealed I should not use ```register``` this modern days, but let the compiler do its job;
- Putting -O3 in the makefile gives the best performance, but to make the change as in-intrusive as possible I used optimization just on the 'palindrome' function.

Interested as always to hear your thoughts about this one.

ryanplusplus:
Changes look good to me.

Out of curiosity, how much of an improvement does adding `-O3` make?

vlzware:
On my laptop (i5 3320M / 2.6GHz):
- original: ~3.7 sec;
- with the new logic to check half/half: ~1.8 sec;
- new logic + register variables: ~1.4 sec;
- new logic + ```-O3``` only on the palindrome function: ~1 sec;
- new logic + ```-O3``` in the makefile: ~0.6 sec;

wolf99:
FYI:  the `register` keyword tells the compiler that the variable is not aliased with anything else in the program. One concrete thing that it _does_ do as part of this is prevent the programmer taking the address of the variable (`&` operator). 
However, because other aspects of the keyword are left up to the implementation (it is only a hint to the compiler, similarly to `inline`) and because lots of modern, system-application level compilers are able to detect variables that are not aliased without the use of the keyword, most people regard it as not worthwhile. 
Because of this its use is rare at system level, similarly to `restrict`. Like the other mentioned keywords that are also relatively rare at the system level, it does sometimes see more use in other areas, such as in embedded firmware - some embedded compiler designers chose to design their silicon and compilers such that the keywords are taken as more than a hint.

PS: if you haven't otherwise seen `restrict` before now it might be a good time to look it up :smiley:

PR #290, opened 2018-03-19
Use the CC implicit variable in exercise makefiles

h-3-0:
This would make it slightly more convenient to override the compiler used.  For example `make CC=clang` vs. having to edit the makefile or having to ensure that the `cc` in the environment is the compiler you want to use.


ryanplusplus:
👍

PR #296, opened 2018-05-23
Implemented exercise 'simple-linked-list'

joshi-b:


exercises/simple-linked-list/makefile
```diff
@@ -0,0 +1,25 @@
...
+memcheck: tests.out
+	@valgrind $(VFLAGS) ./tests.out
+	@echo "Memory check passed"
+
+clean:
+	rm -rf *.o *.out *.out.dSYM
+	
+tests.out: test/test_simple_linked_list.c src/simple_linked_list.c src/simple_linked_list.h
+	@echo Compiling $@
+	@$(CC) $(CFLAGS) src/simple_linked_list.c test/vendor/unity.c test/test_simple_linked_list.c -o tests.out
```
wolf99: There's something wrong with your line-ending encoding here and...

exercises/simple-linked-list/src/example.c
```diff
@@ -0,0 +1,127 @@
+#include "simple_linked_list.h"
+
+/*
+Initialize new node
+*/ 
+    node * new_node(int data) 
+{
+   node * startp = (node *) malloc(sizeof(node));
```
wolf99: ... on these lines also in the example.c, the example.h and the test_simple_linked_list.c
I'm not sure why the CI hasn't caught this.

wolf99: Also on this line - it is considered bad practice to cast the returned value of `malloc()`

exercises/simple-linked-list/src/example.c
```diff
@@ -0,0 +1,127 @@
...
return result;
+
}
+
+
 
+/*
+Insert data at the front of the list
+*/ 
+void insert_beg(node ** startp, int data_new)
```
wolf99: This function name isn't great, specifically the `beg` part. Can you use a different one? Maybe `head_insert()` or something similar

wolf99:
Thanks for contributing an exercise!
I've added a few review comments for you. It's kind of hard to tell fully with the `^M` but the rest looks good!

Usually the struct definition would be in the .c rather than the .h to keep the implementation details encapsulated, but I see that you have left them in the .h to access them in the tests. IIRC there's a better way to do this but I don't recall the details immediately (it's a while since I've looked at C at this point!). If I find the details I'll update here.

config.json
```diff
@@ -1,661 +1,674 @@
 {
+  "language": "C",
```
wolf99: Here you've made a whole bunch of changes to the config.json. Instead it would be best to update _just_ the part that relates to the exercise you are implementing, leaving the other changes to a PR specifically for updating the config.

PR #300, opened 2018-06-10
Word Count: Add a byte for null terminator

### LINKED ISSUE #299: word-count: the text of a word_count_word cannot be zero terminated if it is of maximum length
The tests require a solution to handle words of `MAX_WORD_LENGTH`, but if a word is that long, it cannot be stored in the `text` field of a `word_count_word` struct in the form of a null terminated string. While we could assume that `text` is either a null terminated string, or a string exactly `MAX_WORD_LENGTH` long, that is unpleasant data to work with.

I think the example solution is wrong in that it copies strings [here](https://github.com/exercism/c/blob/5ec4bb1f02a8fd98db39a64a8b8304f48f879ef5/exercises/word-count/src/example.c#L63) in a way that does not guarantee a null terminated string, while assuming null terminated strings with its use of `strcmp` [here](https://github.com/exercism/c/blob/5ec4bb1f02a8fd98db39a64a8b8304f48f879ef5/exercises/word-count/src/example.c#L15).

Assuming what I've said is correct, I think it would be better to have the `text` field be `MAX_WORD_LENGTH + 1` bytes in length.

ryanplusplus:
Closes #299

Thanks @h-3-0 for pointing this out 👍

### DIFF
--- a/exercises/word-count/src/word_count.h
+++ b/exercises/word-count/src/word_count.h
@@ -6,7 +6,7 @@
 
 // results structure
 typedef struct word_count_word {
-   char text[MAX_WORD_LENGTH];
+   char text[MAX_WORD_LENGTH + 1];      // allow for the string to be null-terminated
    int count;
 } word_count_word_t;
 


wolf99:
LGTM

PR #301, opened 2018-06-10
Add function for freeing minesweeper annotation

### LINKED ISSUE #297: Minesweeper tests don't properly free memory
The *free_all* method to free all the memory associated with the output is half-baked. The current method:
```c
static void free_all(char **field)
{
   free(field[0]);
   free(field);
}
```

If it isn't already apparent, we aren't freeing all the strings within the string array. My fix:
```c
static void free_all(char **field, const size_t rows)
{
   for (unsigned long i = 0; i < rows; i++)
      free(field[i]);
   free(field);
}
```

All occurrences of *free_all* would have to have the rows added as an argument, but that's pretty trivial.

ryanplusplus:
Closes #297 

Thanks @redsled84 for pointing this out 👍

### DIFF
--- a/exercises/minesweeper/src/example.c
+++ b/exercises/minesweeper/src/example.c
@@ -62,3 +62,9 @@ char **annotate(const char **minefield, const size_t rows)
 
    return annotated;
 }
+
+void free_annotation(char **annotation)
+{
+   free(annotation[0]);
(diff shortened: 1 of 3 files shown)


wolf99:
LGTM

PR #306, opened 2018-06-12
Changed SNIPPET.txt to be a small standalone program

jackhughesweb:
I've changed the code snippet to appear more like a standard "Hello, World!" program, which can be compiled and run. I've also included a for loop to make use of the available space in the snippet box.

[Issue #303]

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,6 +1,7 @@
-#include "hello_world.h"
+#include <stdio.h>
 
-const char *hello(void)
-{
-   return "Hello, World!";
+int main(int argc, char* argv[]) {
+
+	printf("Hello, World!\n");
+
 }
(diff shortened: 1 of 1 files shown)


docs/SNIPPET.txt
```diff
@@ -1,6 +1,9 @@
-#include "hello_world.h"
+#include <stdio.h>
+
+int main(int argc, char* argv[]) {
+
+	for (int i = 1; i <= 5; i++) {
```
wolf99: Can we get rid of the for loop? 
I'd thought the idea of the SNIPPET.md is to provide the simplest program as possible to provide a very basic example. Indeed, I'm assuming this is why you think it should be self-contained? Anyone with no programming experience at all would see a for loop as a pretty arcane piece of symbol garbage.

ryanplusplus:
Looks great, thanks!

PR #307, opened 2018-06-12
Added missing test information to later exercises

jackhughesweb:
I've added test information which was present in previous exercises to recently added exercises. 

[Issue #303]

PR #309, opened 2018-06-13
Run configlet generate to update READMEs

ryanplusplus:
This should cover the changes in #307 as well.


ryanplusplus:
Thanks @jackhughesweb for pointing out that the READMEs were messed up.

wolf99:
LGTM Thanks both!

wolf99:
Just to note that as this follows on from #307, then it is in aid of #303 :)

PR #313, opened 2018-07-10
Add linked-list exercise

wolf99:
WIP as I can't recall how to generate UUIDs for the config.json and it's late here 🌃. This is why Travis is failing for indent as I have a comment in the config so that it deffo won't pass until I get this done. Info or pointer to docs on this please! (I think I might right a contributing checklist for the docs or maybe as a PR template to aid with these kind of things 🤔, I often forget how to get and install configlet locally too which wastes a bunch of time when it comes to generating the readme). 

There's a couple different ways I could have varied the interface for the linked list but I went with what seemed to closest emulate a list in other languages that have collections, while still maintaining a relatively simple implementation.

Review comments welcome 😃

### DIFF
--- a/exercises/linked-list/src/example.h
+++ b/exercises/linked-list/src/example.h
@@ -7,28 +7,26 @@ typedef int ll_data_t;
 
 struct list_item;
 
-// constructs new item holding specified data
-struct list_item *new(ll_data_t);
+// constructs a new list of items
+struct list_item **new_list(void);
 
-// retrieves data from item
-ll_data_t item_data(struct list_item *item);
+// checks if the list is empty
+bool is_list_empty(struct list_item **list);
 
 // inserts item at back of list
-bool push(struct list_item *item);
+bool push(struct list_item **list, ll_data_t item_data);
 
 // removes item from back of list
-struct list_item *pop(void);
+ll_data_t pop(struct list_item **list);
 
 // removes item from front of list
-struct list_item *shift(void);
+ll_data_t shift(struct list_item **list);
 
 // inserts item at front of list
-bool unshift(struct list_item *item);
-
-// destroy a single item
-bool delete_item(struct list_item **item);
+bool unshift(struct list_item **list, ll_data_t item_data);
 
 // destroy the entire list
-bool delete_list(void);
+// list will be a dangling pointer after calling this method on it
+void delete_list(struct list_item **list);
 
 #endif
--- a/exercises/linked-list/test/test_linked_list.c
+++ b/exercises/linked-list/test/test_linked_list.c
@@ -3,128 +3,97 @@
 #include "vendor/unity.h"
 #include "../src/linked_list.h"
 
+struct list_item **list = NULL;
+
 void setUp(void)
 {
+   list = new_list();
 }
 
 void tearDown(void)
 {
-   delete_list();
+   if (list != NULL) {
+      delete_list(list);
+      list = NULL;
+   }
 }
 
-void test_new(void)
+void test_new_list(void)
 {
    TEST_IGNORE();               // delete this line to run test
-   ll_data_t data = 10;
-   struct list_item *item = new(data);
-   TEST_ASSERT_NOT_NULL(item);
-   delete_item(&item);
+   TEST_ASSERT_NOT_NULL(list);
 }
 
-void test_delete_item_with_invalid_item(void)
+void test_is_list_empty_when_empty(void)
 {
    TEST_IGNORE();
-   TEST_ASSERT_FALSE(delete_item(NULL));
+   TEST_ASSERT_TRUE(is_list_empty(list));
+   delete_list(list);
+   list = NULL;                 // stop list from dangling
+   TEST_ASSERT_TRUE(is_list_empty(list));
 }
 
-void test_delete_item_with_valid_item(void)
-{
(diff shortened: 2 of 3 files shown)


exercises/linked-list/src/example.h
```diff
@@ -0,0 +1,34 @@
+#ifndef LINKED_LIST_H
+#define LINKED_LIST_H
+
+#include <stdbool.h>
+
+typedef int ll_data_t;
```
ryanplusplus: Since this data type is fixed, is there any reason to expose `list_item` to the outside world? What if we the interface were `push(int)` and `int pop()`?

exercises/linked-list/src/example.h
```diff
@@ -0,0 +1,34 @@
...
+struct list_item;
+
+// constructs new item holding specified data
+struct list_item *new(ll_data_t);
+
+// retrieves data from item
+ll_data_t item_data(struct list_item *item);
+
+// inserts item at back list
+bool push(struct list_item *item);
```
ryanplusplus: It looks like this is a singleton. I haven't yet decided whether this makes sense for simplicity or whether this is a bit odd. Can you give your reasoning?

exercises/linked-list/test/test_linked_list.c
```diff
@@ -0,0 +1,151 @@
...
+void test_shift_with_empty_list(void)
+{
+   TEST_IGNORE();
+   TEST_ASSERT_FALSE(shift());
+}
+
+void test_shift_with_populated_list(void)
+{
+   TEST_IGNORE()
+       // pre=populate list
```
ryanplusplus: `pre=populate` => `pre-populate`

ryanplusplus:
Looks good overall. Just a few comments about the exact way in which we are defining the client API.

exercises/linked-list/src/example.h
wolf99: I did have `int` initially but wanted to expose the idea of "variable" type to indicate that the list could really hold anything. 
I tried pushing and popping `ll_data_t` (or `int`) instead of `struct node` to see how it looked, but it makes reaching the empty list state for popping and shifting hard to detect by the caller - this has particular import when destroying the entire list as, unless re-implementing popping in `destroy_list()`, the implementation needs to understand when `shift()` has reached the end of the list - it can't do this if `shift()` only returns an `int`.

Maybe there's another way around this that isn't too complex for an exercise? I'd be happy to adjust to suggestions.

exercises/linked-list/test/test_linked_list.c
wolf99: whoopsie! 👍

exercises/linked-list/src/example.h
```diff
@@ -0,0 +1,34 @@
+#ifndef LINKED_LIST_H
+#define LINKED_LIST_H
+
+#include <stdbool.h>
+
+typedef int ll_data_t;
+
+struct list_item;
```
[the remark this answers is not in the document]
wolf99: I did this for explicitness. This is one of the harder exercises on the track and it does not include the header, thus the type as used in the test file is the only information the student has about what this type is - if I used a `typedef` there would be even less information for the student to go by. 
Though, if you're of the opinion that this should more closely follow the Linux guide I'm happy to change as this is not something I felt strongly about 😃

exercises/linked-list/src/example.h
wolf99: BTW, it could be simpler in this regard if the caller took care of node allocation and freeing, but I wanted to ensure the implementation included these concerns... maybe this was wrong and the exercise should concentrate more exclusively on the linked list algorithm?

exercises/linked-list/src/example.h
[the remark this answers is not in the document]
ryanplusplus: Got it, that makes sense. I agree that's the best approach here if we don't want to confuse someone new to this.

exercises/linked-list/src/example.h
ryanplusplus: Sorry I was referring to the module overall.

exercises/linked-list/src/example.h
ryanplusplus: I think that makes sense. I think I'm approaching this from a "what would I do if this was production code" vs. "what's the best way to help someone learn".

ryanplusplus:
Looks good. The only outstanding comment is regarding the singleton nature of the module. I don't feel strongly about this, so if you're happy as-is then let me know and I'll go ahead and get this merged.

exercises/linked-list/src/example.h
wolf99: I struggle with this too! I rewrote the implementation a few times due to previous incantations being what I would prefer in production shoehorned into an exercise... It wasn't pretty. This is partly why I discarded the non-singleton idea early on - I'm still thinking if I could re-implement on this front though.

wolf99:
As mentioned in reply above, I'd like to see if I can do better on the singleton front - will tie it up by tomorrow evening (GMT) either way hopefully 👍 
Thanks for the review so far! 😃

wolf99:
So "tomorrow evening" was a little ambitious! I have adjusted the implementation to make it non-singleton and hopefully improve the interface some.

PR #314, opened 2018-07-14
Added test for duplicated letter past 'h'.

sesamemucho:
In the current test set, a submission will pass if it only checks for
duplicates in the first 8 letters.


wolf99:
Great @sesamemucho ! Thank you 😃

PR #318, opened 2018-07-16
Update TESTS.md to show `mkdir src`

### LINKED ISSUE #317: Tests.md instructions update (Build your code section)
After doing this for each exercise I think we should add a line to the build your code section of tests.md. Currently we say to issue the command touch src/bob.{h,c}. If you do that without first creating the src directory you get errors (at least on Fedora 28).

touch: cannot touch 'src/bob.h': No such file or directory
touch: cannot touch 'src/bob.c': No such file or directory

Maybe just add "mkdir src (from the exercise directory) to the instructions right before the touch command? I just automatically create the directory for each new exercise, but for someone new it may be nice to be in the instructions.

ryanplusplus:
Closes #317

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -34,6 +34,7 @@ On Windows 10, the Windows Subsystem for Linux (WSL) Bash is available.
 To do this, assuming the current exercise is `bob` and we're in the exercise folder:
 
 ```bash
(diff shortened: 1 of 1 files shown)


wolf99:
LGTM 👍

Issue #328, opened 2018-08-08
Format in readme and test don't agree


mikewalker:
Doesn't this test violate the format outlined in the readme? Its the second one that I am referring to. Since the format is 

```text
(NXX)-NXX-XXXX
```
where `N` is any digit from 2 through 9 and `X` is any digit from 0 through 9.

So in this case after taking the country code away the number 1234567890 would be invalid since the first digit should be in the range [2..9].

```
void test_valid_when_11_digits_and_first_digit_is_1(void)
{
   const char input[] = "11234567890";
   const char expected[] = "1234567890";

   result = phone_number_clean(input);

   TEST_ASSERT_EQUAL_STRING(expected, result);
}
```

ryanplusplus:
You're right, this was a mistake in the upstream test suite that was resolved in this commit: https://github.com/exercism/problem-specifications/commit/51d2475bba52b5ec7b02a4ff12ea42bc293888a5#diff-59f68d802e0ccde10042f386356a7c90. We need to update the tests for this track to reflect the current upstream test suite.

mikewalker:
Thanks for for confirming, I was just working through the exercise and ran into this since my clean method went a bit further than this test was expecting. :)

ryanplusplus:
No, thanks you. You must have great attention to detail to have found this!

ryanplusplus:
If you're interested in making a contribution, the current upstream test suite is detailed here: https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json. If not, @wolf99 or I will make the change as we get some time (hopefully in the next few days).

mikewalker:
I take it getting from the json file into the test suite is a manual process? I'll look into it, this would be my first PR so I may have questions. It does look pretty straight forward though.

ryanplusplus:
Yes, it's a manual process. We're willing to answer questions :)

mikewalker:
closing this since the PR was merged.

PR #329, opened 2018-08-09
phone-number: update test cases to match specifications

mikewalker:
@ryanplusplus hows this look as a first crack at https://github.com/exercism/c/issues/328?

### DIFF
--- a/exercises/phone-number/src/example.c
+++ b/exercises/phone-number/src/example.c
@@ -52,9 +52,8 @@ char *phone_number_clean(const char *input)
       }
    }
 
-   if (strcmp(output, INVALID_NUMBER_RESULT) != 0)
-      if (output[0] < '2' || output[EXCHANGE_OFFSET] < '2')
-         strcpy(output, INVALID_NUMBER_RESULT);
+   if (output[0] < '2' || output[EXCHANGE_OFFSET] < '2')
+      strcpy(output, INVALID_NUMBER_RESULT);
 
    return output;
 }
--- a/exercises/phone-number/src/example.c
+++ b/exercises/phone-number/src/example.c
@@ -27,19 +27,23 @@ char *phone_number_clean(const char *input)
 {
    char *output = calloc(VALID_NUMBER_LENGTH + 2, sizeof(char));
    size_t j = 0;
-
+   
    for (size_t i = 0; i < strlen(input); i++) {
-      if (isdigit(input[i])) {
+    
+      if (isdigit(input[i]) || (i == 0 && input[0] == '+')) {
+      
          if (j > VALID_NUMBER_LENGTH + 1) {
             break;
          }
-         output[j++] = input[i];
+         if (isdigit(input[i]))
+            output[j++] = input[i];
+         
       } else if (strchr(VALID_NON_DIGIT_CHARS, input[i]) == NULL) {
          strcpy(output, INVALID_NUMBER_RESULT);
          return output;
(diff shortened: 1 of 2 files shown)


mikewalker:
@ryanplusplus  As the build error instructs I have run indent.sh to attempt to fix the style issue. This does not help, verify-indent is still detecting style issues. This may be an OS specific issue but the indent.sh script does not appear to have any effect.

I have reinstalled indent, and also trying running it manually on single files as documented in the contributing document. Since its not working manually on a single file it may be a problem with the version of indent that fedora installs.

The OS that I am running on is Fedora 28, and the version of indent is GNU indent 2.2.11.

Without being able to fix the style error I'm a bit stuck at the moment. Do you have any ideas on this one?

mikewalker:
Figured it out, it was verfy-indent that was throwing me off. After seeing how it did the verify I realized I have to check in after running indent.sh, and before running verify-indent again. :)

exercises/phone-number/src/example.c
```diff
@@ -49,23 +51,10 @@ char *phone_number_clean(const char *input)
...
-   char *output = phone_number_clean(input);
-   output[3] = '\0';
-   return output;
-}
-
-char *phone_number_format(const char *input)
-{
-   char *cleaned_input = phone_number_clean(input);
-   char *output = calloc(FORMATTED_LENGTH + 1, sizeof(char));
+   if (strcmp(output, INVALID_NUMBER_RESULT) != 0)
```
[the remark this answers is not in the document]
mikewalker: Just a micro optimization so that I don't check for the if the number is less than '2' when I already know it is. It may actually be less efficient looking at it now, since its calling strcmp and iterating over the string.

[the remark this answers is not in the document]
ryanplusplus: Right, that's what I was thinking. Let's go ahead and omit it and I'll give this my thumbs up :+1:

ryanplusplus:
Looks good to me :+1: 

@wolf99 what do you think?

exercises/phone-number/src/example.c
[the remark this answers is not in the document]
mikewalker: Done...

exercises/phone-number/src/example.c
```diff
@@ -29,11 +27,15 @@ char *phone_number_clean(const char *input)
...
+      if (isdigit(input[i]) || (i == 0 && input[0] == '+')) {
+
          if (j > VALID_NUMBER_LENGTH + 1) {
             break;
          }
-         output[j++] = input[i];
+         if (isdigit(input[i]))
+            output[j++] = input[i];
+
       } else if (strchr(VALID_NON_DIGIT_CHARS, input[i]) == NULL) {
```
wolf99: I know this isn't changed already by this PR, but, for safety's sake, should this condition add  `&& !isspace(input[i])` instead of including `" "` in the `VALID_NON_DIGIT_CHARS` string?

wolf99:
Other than my comment above this LGTM

exercises/phone-number/src/example.c
wolf99: Upon reflection, maybe leave this alone as it is not related to the PR

mikewalker: So leave it as is then?

wolf99: Yep, apologies for the confusion 😁

PR #330, opened 2018-08-16
remove dead code from header

mikewalker:
Missed removing the function prototypes from header with last PR (https://github.com/exercism/c/pull/329) that we merged. :(


ryanplusplus:
Good catch!

Issue #320, opened 2018-07-29
Acronym: longest acronym not long enough


sesamemucho:
I just encountered a solution (https://exercism.io/mentor/solutions/e1e840249aa34b3a8f1e0a99c8b21f1f) in which the student allocated a buffer the size of a char pointer: `char *acr = malloc(sizeof(char *));` instead of basing it on the length of the input phrase. As it happens, none of the acronyms in the tests are longer than 4 characters, so neither `make test` nor `make memcheck` catches this error.
I would recommend making a phrase that abbreviates to something, say, 10 characters long.

I can submit a PR -- just want to go through the correct process.

wolf99:
Thanks for raising this @sesamemucho, this should probably be raised as a PR on the problem specification in the related repo so that the requirement changes drop down to the C track here and other language tracks also. 

Maybe raise a PR there first: https://github.com/exercism/problem-specifications/blob/master/exercises/acronym/canonical-data.json

(Though we dont mind you raising a PR here too once the problem spec PR is accepted! 😉 )

wolf99:
Hi @sesamemucho, were you able to raise an issue or PR on the mentioned repo? If so could you reply here with a link to the one you raised? This lets us keep on track of how things are progressing. Thanks!

sesamemucho:
Sorry! Super busy at work this week. I just submitted a PR, all on github.
First time for that way of doing a PR -- hope it works.

Thanks
Bob

wolf99:
Thanks Bob! PR is https://github.com/exercism/problem-specifications/pull/1287, is now merged and closed

PR #333, opened 2018-08-23
acronym: Added test for long abbreviation

sesamemucho:
See #320

The issue involves not allocating enough space for the result, so `make test` may succeed, and the problem would only show up in `make memcheck`.


wolf99:
Hmm, the indent check (which checks for style) has failed here. At a guess I would say that line 96 is now _too_ long in its current format, d'oh! 

In C, you can split string literals like so:

    char *my_string = "Line 1 "
                      "Line 2";

Could you apply this fix to the relevant line in your change?
Thanks Bob!

ryanplusplus:
Looks great to me. @wolf99?

patricksjackson:
LGTM 👍

PR #336, opened 2018-09-02
Extend C Isogram Test Suite

FroshVII:
- Added test for NULL input
- Added test for duplicated letters on the ends of character arrays
- ~~Added test for terminating as soon as a duplicate is found~~

### DIFF
--- a/exercises/isogram/test/test_isogram.c
+++ b/exercises/isogram/test/test_isogram.c
@@ -18,7 +18,7 @@ void test_empty_string(void)
 void test_null(void)
 {
    TEST_IGNORE();
-   TEST_ASSERT_TRUE(is_isogram(NULL));
+   TEST_ASSERT_FALSE(is_isogram(NULL));
 }
 
 void test_lower_case_only(void)
--- a/exercises/isogram/test/test_isogram.c
+++ b/exercises/isogram/test/test_isogram.c
@@ -18,6 +18,7 @@ void test_empty_string(void)
 
 void test_null(void)
 {
+   TEST_IGNORE();
    TEST_ASSERT_TRUE(is_isogram(NULL));
 }
 
@@ -47,6 +48,7 @@ void test_longest_known_isogram(void)
 
 void test_duplicated_on_ends(void)
 {
+   TEST_IGNORE();
    TEST_ASSERT_FALSE(is_isogram("oreo"));
 }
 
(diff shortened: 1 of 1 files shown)


wolf99:
Thanks for the work @FroshVII, bu are these tests part of the problems specifications? (https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json)

Also, IIRC testing for NULL cases was actually deemed to be unrequired for exercism problem sets (can't think where I read this here so this may indeed be recalled incorrectly, I'll link it if I think of it).

FroshVII:
@wolf99 Well the test for duplicated characters was inspired by @wiredancer's comment at the top of those specs, but the other two are admittedly just additions I personally want confirmed in my code.

wolf99:
Seems the discussion on invalid input does not have a conclusion  (though this is not the discussion I am trying to recall) https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json

wolf99:
FYI, clock time from the C library on any given system can have many influencing factors in comparing between any two runs, such that for single occurrences of very short times (such as this implementation) can often give incorrect results. When comparing measurements of execution time programmers usually record the time it takes to run several tens, if not hundreds, of executions.

exercises/isogram/test/test_isogram.c
```diff
@@ -15,6 +15,12 @@ void test_empty_string(void)
...
 
+void test_null(void)
+{
+   TEST_IGNORE();
+   TEST_ASSERT_FALSE(is_isogram(NULL));
+}
+
 void test_lower_case_only(void)
 {
    TEST_IGNORE();               // delete this line to run test
```
ryanplusplus: Can you move `// delete this line to run test` to the new test? We put this on the second test of all exercises.

ryanplusplus: Since you have your settings configured to allow me to make changes on your pull requests for this repository (thanks!) I went ahead and made this small change.

ryanplusplus:
@wolf99 Discussion regarding invalid inputs is ongoing, but for now it looks like there's precedent on this track for checking for a `NULL` string input (https://github.com/exercism/c/blob/master/exercises/pangram/test/test_pangram.c#L12) so let's go ahead with it for now and come back if the discussion resolves.

ryanplusplus:
Thanks @FroshVII for your contribution!

PR #345, opened 2018-12-10
Run travis against both clang and gcc.

arcuru:
This just sets up Travis to run the tests with both Clang and GCC using their default version on Travis-CI.

(Setting this up mainly because I've seen an error running with recent Clang, and I think this will cause a hit in CI.)

### DIFF
--- a/exercises/palindrome-products/src/example.c
+++ b/exercises/palindrome-products/src/example.c
@@ -4,7 +4,6 @@
 #include <limits.h>
 #include <stdbool.h>
 
-__attribute__ ((optimize("-O3")))
 static bool palindrome(int n)
 {
    /* 0 is palindrome */


patricksjackson:
Yep, there's a warning for the linked list exercise solution. It's not user visible though.

.travis.yml
```diff
@@ -1,5 +1,8 @@
----
-language: bash
+language: c
+
+compiler:
+- clang
+- gcc
```
ryanplusplus: Wow, thanks Travis!

patricksjackson:
I rebased with the new change. This is still failing in linked-list since `shift()` does the exact same thing.

(I was _very_ confused for a bit until I realized the line numbers were different)

patricksjackson:
I fixed up the rest of the issues, though they are both a little interesting...

The linked-list fixes aren't _really_ the right fixes, since that value is kinda/sorta generic a better way is probably to define an error value for the type and use that instead of just `0`. Those functions don't have a defined way of returning an error otherwise.

The palindrome-products example is interesting, it showed up in #289 to improve the performance. If performance is an issue though, the much better way is to improve the algorithm itself. There's a simple fix available of reordering the palindrome check and the bounds check (since checking if a number is a palindrome is far more expensive than just doing an integer compare) which improves the debug execution time from ~1.2s to ~.2s. More difficult in C though, you could also use a priority queue (or a hacked together equivalent) and speed it up to almost instant.

ryanplusplus:
This looks good to me.

The linked-list fix doesn't make the solution any worse, so I don't think that we need to wait on a full solution to get this change in.

I also don't think that the optimization for palindrome-products is a big deal. I agree that the better way forward to decrease run time would be to optimize the algorithm, but I really don't think we have a problem as it is. We don't need to deliver the fastest possible solution, just something that's clear and correct. Let the students push for the optimal solution if that's what they're interested in.

PR #346, opened 2018-12-10
Updating shebang for better portability.

patricksjackson:
`#!/usr/bin/env bash` is supposed to be more portable than directly linking to /bin/bash.

I hit this while running on NixOS.


ryanplusplus:
:+1:

PR #347, opened 2018-12-15
moved the free function

elyashiv:
I think it's bad the solution doesn't require freeing the tree for two reasons:
1. It's not hard to implement, and does teach something
2. I general it's a good habit to provide a free function, as I might allocate with a different function.


patricksjackson:
I think it looks good. I'd like to let somebody else Ack before merging though, since this does change the exercise.

ryanplusplus:
Looks good to me. Thanks for the contribution!

PR #354, opened 2019-02-28
Fix filename in nucleotide-count tests.

siebenschlaefer:
The unit tests for nucleotide-count contained the filename for a
different exercise (rna-transcription).


ryanplusplus:
Great catch, thanks!

Issue #353, opened 2019-02-27
Improve the core track by making "Gigasecond" optional and instead making "Word Count" Core.


QLaille:
It has been quite some time, that I can now say that every  gigasecond exercise I mentor is reduced to this: 
![image](https://user-images.githubusercontent.com/24566679/53481922-62d61b80-3a7e-11e9-80d8-ea84582bdc42.png)

I find it a bit cheap to have an exercise that doesn't let you learn much "Core", and would love instead to permute it with a "Side" exercise where implementations and solutions can differs, thus allowing more exchanges between tutor and students !

ryanplusplus:
Sounds like a great idea. Would you mind making a pull request with these changes?

PR #355, opened 2019-03-03
moved the gigasecond towards the side exercises to move word-count to core tracks

QLaille:
With giga-second being unhelpful and considered a clutter by myself, I suggested to swap it with word-count being harder but leaving room for improvement as the student might be publishing more iterations than with giga second. See issue #353 for clarifying situation.

### DIFF
--- a/config.json
+++ b/config.json
@@ -42,7 +42,7 @@
     {
       "slug": "word-count",
       "uuid": "1e4f4fcf-c32d-459f-8920-15ac56008666",
-      "core": false,
+      "core": true,
       "unlocked_by": "isogram",
       "difficulty": 3,
       "topics": [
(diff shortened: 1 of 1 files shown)


ryanplusplus:
Thanks @QLaille, but a few more things need to be updated. See these notes from the Travis failure:

> -> The exercise 'word-count' is marked as core and unlocked by another exercise. A core exercise should not be unlocked by another.
> -> The exercise 'space-age' is being unlocked by a non-core exercise. Non-core exercises can only be unlocked by core exercises.
> -> The exercise 'meetup' is being unlocked by a non-core exercise. Non-core exercises can only be unlocked by core exercises.

ryanplusplus:
Thanks, looks great!

Issue #360, opened 2019-03-17
Correctly handling extra compilation flags


QLaille:
As of now, I had a solution not compiling, because he was using the math lib.
I prompted him to add the `-lm` flag to the makefile, so that it would correctly use the lib.

What bugged me out is that it still wasn't compiling, and I noticed that moving from :

```
	@$(CC) $(CFLAGS) $SOLUTION_SOURCE_FILES -o tests.out
```

to

```
	@$(CC) $SOLUTION_SOURCE_FILES -o tests.out  $(CFLAGS)
```

Would cause the linker to correctly handle extra libraries added by the students.
Maybe we should consider modifying all makefiles of the C track (and potentially CPP) to correctly handle this ?

patricksjackson:
Well that's surprising.

You're absolutely right though, libraries need to be specified after the source files in this manner. I'd prefer it if we added a `LIBS` variable separate from CFLAGS to be a bit more clear about it.

Personally, I think it might be a good idea to just link against the math library all the time if they've found a reason to use it in these problems. It's by far the most common and would be very confusing to figure out.

QLaille:
I see where you are going, for this case it was the grains exercise, in which you have to write a function named square, perhaps we should modify subjects to forbid the usage of said libraries, but I'm afraid it would be against exercism itself...

patricksjackson:
There's an argument to be made either way, to use libs or not. The math library is so basic I'd think it would be ok though. Other libraries probably wouldn't be a great idea.

And if there's an exercise where it's being used that it really shouldn't (like, honestly, the grains exercise, since powers of two is the whole exercise) we could disallow it in the docs and remove it from the makefile.

The other maintainers might have a different opinion here though.

(Also I'm the maintainer on Cpp, we use CMake instead of Makefiles and so I'm not entirely sure if it needs to be explicitly linked. I'll try to remember to check when I get the chance).

QLaille:
That's another issue I'll open in some time, for now I'd like to focus solely on moving the inclusion of flags the correct way.

ryanplusplus:
> That's another issue I'll open in some time, for now I'd like to focus solely on moving the inclusion of flags the correct way.

Agreed. I think we should proceed with @patricksjackson's suggestion to add a new `LIBS` variable and place it properly in the compiler invocation.

QLaille:
Want me to make a PR asap @ryanplusplus  ?

ryanplusplus:
Yes please 😊

PR #361, opened 2019-03-17
Handling student's added libraries in makefile

QLaille:
Following #360 , I've modified every makefile of every track to rely on a LIBS variable referencing extra library flags for an exercise, commented how to add a library and added to the compilation line said variable. (I also still suck a bit with git and it shows previous commits).

### DIFF
--- a/exercises/secret-handshake/makefile
+++ b/exercises/secret-handshake/makefile
@@ -1,3 +1,7 @@
+### If you wish to use extra libraries (math.h for instance),
+### add their flags here (-lm in our case) in the "LIBS" variable.
+
+###
 CFLAGS  = -std=c99
 CFLAGS += -g
 CFLAGS += -Wall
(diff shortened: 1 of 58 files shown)


exercises/secret-handshake/makefile
```diff
@@ -1,3 +1,7 @@
+### If you wish to use extra libraries (math.h for instance),
+### add their flags here (-lm in our case) in the "LIBS" variable.
```
ryanplusplus: My only nitpick is that we include `(-lm in our case)` even when we don't have `-lm` (or any `LIBS` variable at all).

QLaille:
Silly me doing silly things with the clipboard, hang on.

ryanplusplus:
Looks good to me.

@patricksjackson any thoughts before merge?

patricksjackson:
I think it's fine, I'll go ahead and merge. I also was unaware that some of our Makefiles were customized for the exercise. I'll try to remember that in future.

Thanks @QLaille !

Closes #360

PR #368, opened 2019-03-30
Update test_word_count.c

ShadowWolf387:
All the change does is prevent the test from crashing when it tries to write the 21st word for the excessive words test. 
Bug was found under Win7  with Dev-C++ and Mingw 8.1.0 but was working fine with Msys2.


ryanplusplus:
Thanks!

PR #369, opened 2019-03-31
Moving to out of tree builds.

patricksjackson:
Modifies the `bin/run-tests` script to use out of tree builds, and to allow running a subset of tests to improve development.

I verified POSIX compat using https://github.com/koalaman/shellcheck, which we can add to Travis easily if needed (there are warnings in both `verify-indent` and `fetch-configlet` that should probably be fixed if we enable it).

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -137,6 +137,12 @@ Firstly make sure you have the necessary tools installed (such as `indent`, [`gi
 ~/git/c$ ./bin/run-tests
 ```
 
+If you'd like to run only some of the tests to check your work, you can specify them as arguments to the run-tests script.
+
+```bash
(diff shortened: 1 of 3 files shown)


docs/CONTRIBUTING.md
```diff
@@ -137,6 +137,12 @@ Firstly make sure you have the necessary tools installed (such as `indent`, [`gi
 ~/git/c$ ./bin/run-tests
 ```
 
+If you'd like to run only some of the tests to check your work, you can specify them as arguments to the run-tests script.
+
+```bash
+~/git/c$ ./bin/run-tests acronym all-your-base allergies
```
ryanplusplus: :+1:

ryanplusplus:
This is great work. Thanks!

PR #371, opened 2019-04-01
Removing pointer compare in binary-search.

### LINKED ISSUE #356: Flag for mismatch between declaration and implementation
Hi. I just looked at a solution for binary_search, and the student has submitted a header with the following line:
```c
int binary_search(int val, int *arr, size_t length);
```
The function returns int instead of a pointer, but the whole thing compiles without any warnings :disappointed:

Is there a way to get warning for that (casting int to int*, mismatch in return type)?

patricksjackson:
Closes #356


ryanplusplus:
:+1:

PR #372, opened 2019-04-02
Switching memcheck to use Asan.

### LINKED ISSUE #370: Improve memcheck
As @ShadowWolf387 reported in #367, clearly there is some issue with the memory checking, so I did some digging.

It turns out that Valgrind can't detect that kind of overflow (Global OOB) because it is only instrumenting heap allocations, and in all the places I've seen is being superseded by Address Sanitizer.

I propose we move from using Valgrind to use Address Sanitizer, which has been built into the major OSS compilers since ~2014 (Clang 3.1, GCC 4.8, XCode 7.0). Visual Studio has it's own tools and you can't use Valgrind on Windows anyways.

Since it requires editing each of the makefiles, I think at the same time we should just standardize on 1 makefile that is used in every exercise (mostly because I'm lazy). The new makefile template would be the below, which will work along with a small tweak to `bin/run-tests` to `mv` instead of `cp` the example files. The out of tree builds I setup in #369 would allow that to happen cleanly.

Any thoughts/objections?

```
### If you wish to use extra libraries (math.h for instance),
### add their flags here (-lm in our case) in the "LIBS" variable.

LIBS = -lm

###
CFLAGS  = -std=c99
CFLAGS += -g
CFLAGS += -Wall
CFLAGS += -Wextra
CFLAGS += -pedantic
CFLAGS += -Werror

ASANFLAGS  = -fsanitize=address
ASANFLAGS += -fno-common
ASANFLAGS += -fno-omit-frame-pointer

test: tests.out
	@./tests.out

memcheck: test/*.c src/*.c src/*.h
	@echo Compiling $@
	@$(CC) $(ASANFLAGS) $(CFLAGS) src/*.c test/vendor/unity.c test/*.c -o memcheck.out $(LIBS)
	@./memcheck.out
	@echo "Memory check passed"

clean:
	rm -rf *.o *.out *.out.dSYM

tests.out: test/*.c src/*.c src/*.h
	@echo Compiling $@
	@$(CC) $(CFLAGS) src/*.c test/vendor/unity.c test/*.c -o tests.out $(LIBS)
```

patricksjackson:
This swaps out Valgrind for AddressSanitizer and standardizes on an identical makefile for all tests, as I proposed in #370.

Closes #370 

(Temporarily setting this as WIP to demonstrate that CI still catches memory leaks)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ compiler:
 
 before_install:
 - sudo apt-get -qq update
-- sudo apt-get install -y indent valgrind
+- sudo apt-get install -y indent
 
 script:
 - bin/fetch-configlet
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -26,12 +26,12 @@ execute_test () {
         # Copy the examples with the correct name for the exercise
         if [ -e "src/example.c" ]
         then
-          cp src/example.c src/"${EXERCISE_NAME}".c
+          mv src/example.c src/"${EXERCISE_NAME}".c
         fi
 
         if [ -e "src/example.h" ]
         then
-          cp src/example.h src/"${EXERCISE_NAME}".h
+          mv src/example.h src/"${EXERCISE_NAME}".h
         fi
 
         # Make it!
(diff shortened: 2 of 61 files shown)


ryanplusplus:
Thanks for taking the time to do this :+1:

PR #373, opened 2019-04-02
Adding ShellCheck linter to Travis.

patricksjackson:
This adds a [ShellCheck](https://github.com/koalaman/shellcheck) static analysis run to lint the bash scripts under bin/. Travis has it installed by default, so it can be invoked pretty easily.


ryanplusplus:
👍

PR #376, opened 2019-04-04
palindrome-products: Made the check_factors test function order-insensitive

### LINKED ISSUE #375: palindrome-products: Should the check_factors test function be order-insensitive?
I am having a problem with the `test_largest_palindrome_from_single_digit_factors` test case, where the expected factors are initialized as `{{3, 3, NULL},  {1, 9, NULL} }` and are checked by the `check_factors` function in this specific order, while my solution returns factors in the `{ {1, 9, NULL}, {3, 3, NULL} }` order.

Perhaps the `check_factors` could be updated to accept factors in any order?

ZapAnton:
Closes #375

### DIFF
--- a/exercises/palindrome-products/test/test_palindrome_products.c
+++ b/exercises/palindrome-products/test/test_palindrome_products.c
@@ -1,6 +1,9 @@
 #include "vendor/unity.h"
 #include "../src/palindrome_products.h"
 
+#include <stdbool.h>
+#include <stddef.h>
+
 void setUp(void)
 {
 }
@@ -9,24 +12,45 @@ void tearDown(void)
 {
 }
 
-void check_factors(factor_t * actual, int depth, factor_t expected[])
+bool factor_t_are_equal(const factor_t * const f1, const factor_t * const f2)
+{
+   return ((f1->factor_a == f2->factor_a) && (f1->factor_b == f2->factor_b)) ||
+       ((f1->factor_a == f2->factor_b) && (f1->factor_a == f2->factor_b));
+}
+
+bool
+contains_factor(factor_t * factors, const factor_t * const factor,
+                const size_t depth)
+{
+   size_t current_depth = 0;
+   factor_t *current_factor = factors;
+   while ((current_factor != NULL) && (current_depth != depth)) {
+      if (factor_t_are_equal(current_factor, factor)) {
+         return true;
+      }
+      current_factor = current_factor->next;
+      current_depth += 1;
+   }
+   return false;
+}
+
+void check_factors(factor_t * actual, size_t depth, factor_t expected[])
 {
    if (depth == 0) {
       TEST_ASSERT_EQUAL_PTR(NULL, actual);
       return;
    }
-   int i;
-   int count_ok = 0;
-   for (i = 0; i < depth; i++) {
-      if (actual == NULL)
+   bool found_all = true;
+   int found_count = 0;
+   for (size_t i = 0; i < depth; ++i) {
+      if (!contains_factor(actual, &expected[i], depth)) {
+         found_all = false;
          break;
-      if ((actual->factor_a == expected[i].factor_a) &&
-          (actual->factor_b == expected[i].factor_b))
-         count_ok++;
-      actual = actual->next;
+      }
+      found_count += 1;
    }
-   TEST_ASSERT_EQUAL_PTR(NULL, actual);
-   TEST_ASSERT_EQUAL_INT(depth, count_ok);
+   TEST_ASSERT_TRUE(found_all);
+   TEST_ASSERT_EQUAL_INT(depth, found_count);
 }
 
 void test_smallest_palindrome_from_single_digit_factors(void)
--- a/exercises/palindrome-products/test/test_palindrome_products.c
+++ b/exercises/palindrome-products/test/test_palindrome_products.c
@@ -12,6 +12,18 @@ void tearDown(void)
 {
 }
 
+/* factor_t_are_equal compares the values of the factors
+ * that are stored in the passed f1, f2 factor_t structs
+ * regardless of their order in the said structs. Returns
+ * true if the values are equal, false otherwise.
+ *
+ *
+ * Examples:
+ *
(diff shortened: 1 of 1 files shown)


exercises/palindrome-products/test/test_palindrome_products.c
```diff
@@ -9,24 +12,45 @@ void tearDown(void)
...
+{
+   for (size_t i = 0; i < depth; ++i) {
+      if (factor_t_are_equal(&factors[i], factor)) {
+         return true;
+      }
+   }
+   return false;
+}
+
+void check_factors(factor_t * actual, size_t depth, factor_t expected[])
```
patricksjackson: Does this work if we pass in this?

expected == {{1, 9}, {3, 3}}
actual == {{1, 9}, {1, 9}}

ZapAnton: This test case passes with the current version on the `check_factors`. Will try to fix this.

ZapAnton: @patricksjackson I think I managed to fix the `check_factors` behavior, can you give it another look?

patricksjackson: Yep, should work now with any order. It won't give very good error messages, but I'm not quite sure how to improve that.

(I admittedly am a bit rusty on my C, so I didn't expect it to be this complicated. In C++ it'd be a lot easier to just throw the LL into an array, sort everything, and compare them. That approach might be a bit difficult for C though.)

exercises/palindrome-products/test/test_palindrome_products.c
```diff
@@ -9,24 +12,45 @@ void tearDown(void)
...
-      if ((actual->factor_a == expected[i].factor_a) &&
-          (actual->factor_b == expected[i].factor_b))
-         count_ok++;
-      actual = actual->next;
+      }
+      found_count += 1;
    }
-   TEST_ASSERT_EQUAL_PTR(NULL, actual);
-   TEST_ASSERT_EQUAL_INT(depth, count_ok);
+   TEST_ASSERT_TRUE(found_all);
```
patricksjackson: You don't need `found_all`, it's redundant with the depth check.

exercises/palindrome-products/test/test_palindrome_products.c
```diff
@@ -9,24 +12,45 @@ void tearDown(void)
 {
 }
 
-void check_factors(factor_t * actual, int depth, factor_t expected[])
+bool factor_t_are_equal(const factor_t * const f1, const factor_t * const f2)
+{
+   return ((f1->factor_a == f2->factor_a) && (f1->factor_b == f2->factor_b)) ||
```
patricksjackson: You can decide here, but it seems a lot clearer to me to use min/max to do this, but that's just me.

```
return (min(f1->factor_a, f1->factor_b) == min(f2->factor_a, f2->factor_b)) && 
       (max(f1->factor_a, f1->factor_b) == max(f2->factor_a, f2->factor_b));
```

exercises/palindrome-products/test/test_palindrome_products.c
```diff
@@ -9,24 +12,45 @@ void tearDown(void)
 {
 }
 
-void check_factors(factor_t * actual, int depth, factor_t expected[])
+bool factor_t_are_equal(const factor_t * const f1, const factor_t * const f2)
```
patricksjackson: Can you add comments for this group of functions explaining what they do? Especially since there's 3 of them, maybe add a block comment above with the explanation, and a comment below noting the end of the helper functions.

exercises/palindrome-products/test/test_palindrome_products.c
```diff
@@ -9,24 +12,45 @@ void tearDown(void)
...
+{
+   return ((f1->factor_a == f2->factor_a) && (f1->factor_b == f2->factor_b)) ||
+       ((f1->factor_a == f2->factor_b) && (f1->factor_a == f2->factor_b));
+}
+
+bool
+contains_factor(factor_t * factors, const factor_t * const factor,
```
patricksjackson: All these functions should be `static`.

exercises/palindrome-products/test/test_palindrome_products.c
```diff
@@ -9,24 +12,45 @@ void tearDown(void)
...
-          (actual->factor_b == expected[i].factor_b))
-         count_ok++;
-      actual = actual->next;
+      }
+      found_count += 1;
    }
-   TEST_ASSERT_EQUAL_PTR(NULL, actual);
-   TEST_ASSERT_EQUAL_INT(depth, count_ok);
+   TEST_ASSERT_TRUE(found_all);
+   TEST_ASSERT_EQUAL_INT(depth, found_count);
```
patricksjackson: Can you add a comment close to here to explain what is wrong if this fails? It's not very obvious.

Ideally I'd want to print out the two things but I'm not sure if we can with this test suite.

exercises/palindrome-products/test/test_palindrome_products.c
patricksjackson: Ugh, nevermind. I didn't realize min/max aren't in the C stdlib, so you'd have to define them too.

ZapAnton:
@patricksjackson, I have addressed all the moment that you have pointed out. Is there anything else?

exercises/palindrome-products/test/test_palindrome_products.c
```diff
@@ -9,24 +12,71 @@ void tearDown(void)
...
-          (actual->factor_b == expected[i].factor_b))
-         count_ok++;
-      actual = actual->next;
+      }
+      found_count += 1;
    }
-   TEST_ASSERT_EQUAL_PTR(NULL, actual);
-   TEST_ASSERT_EQUAL_INT(depth, count_ok);
+   TEST_ASSERT_EQUAL_INT_MESSAGE(depth, found_count,
+                                 "Not every expected factor found in the actual result.");
```
patricksjackson: I forgot you could add the message to be printed with the error, that's a good touch :)

patricksjackson:
LGTM, thanks!

PR #382, opened 2019-04-27
Force 64bit support in tests

patricksjackson:
We haven't been supporting this on non-64bit platforms I think, but these are the changes to make it happen.

I uncovered this while playing around with a rasberry pi. It turns out for some platforms you need to explicitly opt-in to the test suite using 64bit types. This probably breaks something for very small embedded devices, but adds real support for 32bit devices that have stdint. (C99+ I believe).

That change also exposed some type issues in a few test suites that are fixed here. In two places it's replacing a `long` with an `int64_t`, as the current examples depend on `long` being a 64bit type which is true on x86_64.

In one case it's replacing a print statement for a `size_t`, which was a compiler error on the Pi but not anywhere else, which I don't quite understand.


wolf99:
> In one case it's replacing a print statement for a `size_t`, which was a compiler error on the Pi but not anywhere else, which I don't quite understand.

`zu` is correct for `size_t` for implementations that are C99+ conformant. I imagine though that implementations that are backwards compatible to pre-C99, or default to ISO-C, define `size_t` differently and so don't mark `ld` as an error (maybe a warning or in `--pedantic` mode though?)

See https://stackoverflow.com/questions/2524611/how-can-one-print-a-size-t-variable-portably-using-the-printf-family

As Pi is a relatively modern platform, I would imagine it uses a little stricter C99+ implementation?

ryanplusplus:
LGTM :+1:

patricksjackson:
Thanks @wolf99, I didn't know that :)

PR #384, opened 2019-05-18
isogram: Add test case for duplicates following non-letter character

rootbeersoup:
Without this test, you can pass all tests while ignoring everything after the first non-letter character.

### DIFF
--- a/exercises/isogram/test/test_isogram.c
+++ b/exercises/isogram/test/test_isogram.c
@@ -45,16 +45,16 @@ void test_longest_known_isogram(void)
    TEST_ASSERT_TRUE(is_isogram("subdermatoglyphic"));
 }
 
-void test_same_first_and_last_characters(void)
-{
-   TEST_IGNORE();
-   TEST_ASSERT_FALSE(is_isogram("oreo"));
-}
-
 void test_duplicated_letter_mixed_case(void)
 {
    TEST_IGNORE();
    TEST_ASSERT_FALSE(is_isogram("Alphabet"));
+}
+
+void test_duplicated_letter_mixed_case_lowercase_first(void)
+{
+   TEST_IGNORE();
+   TEST_ASSERT_FALSE(is_isogram("alphAbet"));
 }
 
 void test_non_letter_char(void)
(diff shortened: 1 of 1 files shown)


patricksjackson:
Usually we like to track the tests used in the canonical test suites for exercism overall. For Isogram those tests are here: https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json

Often our version of the tests get out of date, since those are updated occasionally. It turns out that that case is tested by the current version of the test suite - https://github.com/exercism/problem-specifications/blob/3b14d193bc636a7c497d09ec17a3813a5c3d1cee/exercises/isogram/canonical-data.json#L82

So this is a good change, but would you be willing to go through and fix up the full test suite too? We'd want to keep the extra test for "NULL", but everything else should match the tests listed in the canonical tests.

rootbeersoup:
Sure. Do the tests need to mirror the spec 100%? E.g. The current `test_same_first_and_last_characters(void)` tests for [`"oreo"`](https://github.com/exercism/c/blob/2541068ebd3998a59ebb9c9cc6779cde3b9d0537/exercises/isogram/test/test_isogram.c#L51) while the spec tests for [`"angola"`](https://github.com/exercism/problem-specifications/blob/3b14d193bc636a7c497d09ec17a3813a5c3d1cee/exercises/isogram/canonical-data.json#L114). Practically it makes no difference but I'm not sure. If yes, I can change any other discrepancies with the test strings.

Also, does the order of the tests matter at all?

patricksjackson:
I prefer it mirrors it as closely as is reasonable, including with the strings used. In that case it doesn't matter in practice but in some of them they do. That central repo catches a lot of corner cases that we ordinarily wouldn't think of.

Keeping it very similar also makes it easier to check that they are the same and update them when necessary, since we're doing it all by hand. The order of the tests also matters since it's usually setup as a logical progression of new features to add to the solution.

rootbeersoup:
Let me know how this looks. Rearranging things makes the diff look a little strange, but I think I have everything in the correct order.

patricksjackson:
Perfect :)

Thanks!

PR #385, opened 2019-06-02
Allergies: Use stack allocated arrays.

### LINKED ISSUE #321: Static Array for Allergies?
I just got a comment from a mentor on the allergies exercise about why I was allocating memory. It made me wonder, why we are myself. The typedef is defined in the header that is part of the exercise, and the tests are feeing memory. I cannot change this without changing the exercise, so I am asking here. Should this be changed?

The comment was:
```

Why are you allocating the parameter list in get_allergens? Why simply not declaring it in allergen_list_t as static array of size ALLERGEN_COUNT? Something like the following:

typedef struct {
   int count;
   allergen_t allergens[ALLERGEN_COUNT];
} allergen_list_t;

then you can initialize it using memset instead of calloc. 
```

patricksjackson:
Closes #321 

This removes all the heap allocations from the allergies exercise. I also simplified the exercise by just switching to using an array of bools for output instead of an array of allergen values so that it more directly demonstrates how you can think of a bitfield as an array of bools.

I explicitly did _not_ update this to the [new test suite](https://github.com/exercism/problem-specifications/blob/master/exercises/allergies/canonical-data.json) since that would require a full overhaul of the tests, and this change is intended to just address the present concerns.

### DIFF
--- a/exercises/allergies/src/allergies.h
+++ b/exercises/allergies/src/allergies.h
@@ -1,6 +1,8 @@
 #ifndef ALLERGIES_H
 #define ALLERGIES_H
 
+#include <stdbool.h>
+
 typedef enum {
    ALLERGEN_EGGS = 0,
    ALLERGEN_PEANUTS,
@@ -10,12 +12,12 @@ typedef enum {
    ALLERGEN_CHOCOLATE,
    ALLERGEN_POLLEN,
    ALLERGEN_CATS,
-   ALLERGEN_COUNT
+   ALLERGEN_COUNT,
 } allergen_t;
 
 typedef struct {
    int count;
-   allergen_t *allergens;
+   bool allergens[ALLERGEN_COUNT];
 } allergen_list_t;
 
 #endif
--- a/exercises/allergies/test/test_allergies.c
+++ b/exercises/allergies/test/test_allergies.c
@@ -10,30 +10,9 @@ void tearDown(void)
 {
 }
 
-//helper functions
-void list_count_is(int count, allergen_list_t * list)
-{
-   TEST_ASSERT_EQUAL(count, list->count);
-}
-
-void list_contains(allergen_t allergen, allergen_list_t * list)
-{
-   bool allergen_found = false;
-
-   for (int i = 0; i < list->count; i++) {
-      if (list->allergens[i] == allergen) {
-         allergen_found = true;
-         break;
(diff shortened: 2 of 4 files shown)


exercises/allergies/src/allergies.h
```diff
@@ -10,12 +12,12 @@ typedef enum {
    ALLERGEN_CHOCOLATE,
    ALLERGEN_POLLEN,
    ALLERGEN_CATS,
-   ALLERGEN_COUNT
+   ALLERGEN_COUNT,
 } allergen_t;
 
 typedef struct {
    int count;
```
ryanplusplus: Does it make sense to include `count` now? The size of `allergens` is fixed and the allergens in the list are not sorted by their presence, so it's difficult to see how `count` is adding any value.

exercises/allergies/test/test_allergies.c
```diff
@@ -10,30 +10,9 @@ void tearDown(void)
...
-
-   for (int i = 0; i < list->count; i++) {
-      if (list->allergens[i] == allergen) {
-         allergen_found = true;
-         break;
-      }
-   }
-
-   TEST_ASSERT_TRUE(allergen_found);
-}
```
ryanplusplus: :+1: 

While I'm normally all for test helpers, I think keeping these tests as simple as possible makes this more accessible.

exercises/allergies/src/allergies.h
patricksjackson: Oh, I thought I explained this in my PR comment but apparently not.

I left `count` because I noticed that the test suite as it stands has the issue that it never does a negative test while returning the list. So just returning an array of all `true` would work for every test.

This suite is mostly equivalent to the old version, which was my original goal with the changes, but still has that rather glaring issue in it. To be honest though, I mostly stopped at this because I was lazy and didn't feel like finishing the full overhaul of the tests.

It might be worth it to go ahead and update to the new test suite while improving this. I would remove this `count` and add back a helper function that takes a list of allergens and validates that only those are set in the array, and then update all the tests to match. The error messages will be a _lot_ worse for that though.

ryanplusplus: Got it. Let's go ahead and merge this then we'll address the remaining issues when we overhaul the test suite to get it back in sync.

Issue #351, opened 2019-01-28
Consider adding CFLAGS += -Wmissing-declarations to makefile


Alexhans:
One correction I've seen is to make functions `static` when they're only going to be called from within the file.  

Since `-Wmissing-declarations` in `GCC` checks this I wondered if it should be added to the default `makefile`.

That would, of course, would require some changes in the `test/<exercise>.c` files with something like

    %s/^void test/void test

but it would probably help automate that problem and learners could also learn from that makefile for their own projects.

Something should also be done so that setUp and tearDown don't result in complaints from the compiler.

Alexhans:
Just to clarify, the -Wmissing declarations would force students to make the functions they don't declare as `static`, which seems to be desirable.

Without the proposed changes proposed in the first comment the warning turned error is shown like so:
> test/test_grains.c:4:6: error: no previous declaration for ‘setUp’ [-Werror=missing-declarations]
>  void setUp(void)

Defining a different `CFLAGS` for the files under `src` seems to be a reasonable idea.

ryanplusplus:
Hey @Alexhans, sorry I wasn't clear. I meant to ask how the `%s/^void test/void test` would prevent the test functions from being falsely flagged as having no previous declaration. It looks to me like it will replace all places where `void test` appears at the beginning of a line with the exact same text, but I'm not a `sed` guru so I may be misunderstanding the pattern or your intent.

Alexhans:
Oh, that was a typo.

Meant to replace `void test` occurring at the beginning of a line with `static void test`.

    %s/^void test/static void test

elyashiv:
It seems like the warning is a little cryptic. Is the point to save on repeating comments about static functions? Following the error message, I can see lots of students adding a declaration instead of making the function static (especially because the first result in google when searching for this warning tells me to do that).

patricksjackson:
Yes, I believe the point is to enforce good practices (setting non-global functions as static) with the compiler instead of through comments. I agree though that the error is not great...I'd be somewhat against it for that reason.

The suggested changes to the test files might be worth doing regardless so that the test functions themselves are declared static as better examples.

The `setUp()` and `tearDown()` methods have declarations in the latest version of the test suite, so updating to the latest Unity version would fix those issues if it's decided to add the flag.

Alexhans:
Indeed.  The point would be to force the students to confront the issue instead of possibly finding out about it later.  There's always a possibility that the mentor forgets to mention it.

[...]

PR #386, opened 2019-06-16
Update test framework

wolf99:
Updates the Unity test framework for all exercises to use the most recent release (v2.4.3).

    git/c$ cp -r -v ~/Unity-2.4.3/src/* ./exercises/*/test/vendor/

Additionally makes all test functions `static`

    git/c$ sed -i -e 's/^void test/static void test/' ./exercises/*/test/test_*.c

This highlighted errors in test files for two exercises, these are also corrected:
- an unused test function in `all-your-base`, this is now used
- an unused test helper function in `clock`, this is now removed

This is in preparation for #351

### DIFF
--- a/docs/C_STYLE_GUIDE.md
+++ b/docs/C_STYLE_GUIDE.md
@@ -45,16 +45,16 @@ Upper-case snake case for macro names is appreciated, but macros resembling func
 
 The standard fixed-width types (in `stdint.h`) are preferred over basic integer types.
 
-The type `bool` (from `stdbool.h`) is preferred over `_Bool`. 
+The type `bool` (from `stdbool.h`) is preferred over `_Bool`.
 
 Likewise, `size_t` and `ptrdiff_t` (of `stddef.h`) are preferred for representing memory-related quantities.
 
-Types defined by a program using `typedef` should be named with a `_t` suffix. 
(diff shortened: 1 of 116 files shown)


wolf99:
I noticed while doing this, that some of the exercises use different versions of the framework. While not a major issue to date, it did cause a little confusion as the framework source does not mention it and the framework only sees releases every few years (so it is not very obvious which one is in use)

As such it may be a good idea document somewhere the version of the test framework in use. Not sure the best way to do this... maybe a `versions.md` in the documentation directory?

ryanplusplus:
@wolf99 what does the newer version of Unity do for us? And will we still have `setUp` and `tearDown` be flagged as having a missing `static`?

As far as a version file that sounds good. I kind of doubt that anyone will ever really look at it, but it's low cost and could help future maintenance efforts.

wolf99:
@ryanplusplus In older versions of Unity the `setUp()` and `tearDown()` functions were not defined and declared as normal global C functions, meaning that using `-Wmissing-declarations` would flag them as not being marked as `static`, but marking them as `static` would break their usage. 

In the newer version of Unity the way these functions are defined as normal global C functions, thus `-Wmissing-declarations` doesn't flag them and so we don't need to (and shouldn't) mark them as `static` 😃 .

PR #387, opened 2019-06-17
Add versions.md

wolf99:
Adding verions doc to assist in case of future maintenance efforts
Per https://github.com/exercism/c/pull/386#issuecomment-502489832


patricksjackson:
Do you know if putting a new markdown file in "docs/" will be added to the website somehow? I don't know how the documentation is built.

I'm ok with this but it would probably be good to also add this version info to CONTRIBUTING.md and TESTS.md.

wolf99:
Have added reference and link to the VERSIONS.md in the bullet point that describes the test folder, in the CONTRIBUTING.md doc.

The TESTS.md seems (currently) to be more concerned with running tests, from a users perspective, than any track developer concerns such as 3rd party package versions - so I have not added here for now.

I'm too am not sure how the documentation is built. The docs we have seem to be what is used on the site, but not all of them. I have no idea how the selection of what to show there or not is made. I don't see it mentioned in the Exercism documentation.

.... Should we try it and see what happens?

docs/C_STYLE_GUIDE.md
```diff
@@ -108,7 +108,7 @@ What these options do is indicated by the [GNU `indent` manual](https://www.gnu.
 ## Test file layout
 
 Each exercise's test file follows a particular layout.
-The layout is intended to both help contibutors to identify parts of the file, and to help learners to understand how the tests work.
+The layout is intended to both help contributors to identify parts of the file, and to help learners to understand how the tests work.
```
ryanplusplus: :+1:

ryanplusplus:
Let's try and see what happens. Worst case scenario we fix it quickly.

PR #388, opened 2019-06-19
Add -Wmissing-declarations

wolf99:
Per #351 

    sed -i -e '/^CFLAGS += -Werror/a CFLAGS += -Wmissing-declarations' ./*/makefile

Plus correction of the errors that arose due to the additional flag

PS 
The `cp` command that I used in #386 `git/c$ cp -r -v ~/Unity-2.4.3/src/* ./exercises/*/test/vendor/` only seems to have worked for the first n exercises even though the output of the command seemed to indicate it had worked for all exercises... any clues why that might have happened?

### DIFF
--- a/exercises/acronym/makefile
+++ b/exercises/acronym/makefile
@@ -10,6 +10,7 @@ CFLAGS += -Wall
 CFLAGS += -Wextra
 CFLAGS += -pedantic
 CFLAGS += -Werror
+CFLAGS += -Wmissing-declarations
 CFLAGS += -DUNITY_SUPPORT_64
 
 ASANFLAGS  = -fsanitize=address
--- a/exercises/acronym/src/example.c
+++ b/exercises/acronym/src/example.c
@@ -1,3 +1,4 @@
+#include "acronym.h"
 #include <ctype.h>
 #include <stdlib.h>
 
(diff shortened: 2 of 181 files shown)


wolf99:
Figured out my postscript question by retrying the command and examining the output more closely
Example line:
    ./exercises/all-your-base/test/vendor/unity.h' -> './exercises/wordy/test/vendor/vendor/unity.h'

Turns out the command is completely wrong as, for each and every exercise it writes the new Unity files `wordy/vendor/vendor`... 😳 🙈

patricksjackson:
This all looks good to me. I won't merge in case you want to wait for @ryanplusplus

**The test copying:**
Ah, right, I didn't read #386 too closely. I think the issue is that the `*` in `./exercises/*/test/vendor` doesn't work like you'd hope it would in this context. I would've expected you'd get errors or a lot of extra files though.

In the future, you can probably do the copy using `xargs` or `find` with the `-exec` flag. Usually I find it easier to just write a short script since I never remember the syntax (I use fish shell, so the syntax would be a bit different for bash).

``` fish
cd exercises
for x in *:
  cp -r ~/Unity/src/* $x/test/vendor/
end
```

ryanplusplus:
:+1:

Issue #374, opened 2019-04-02
A little change for Clock in C track


fjarboleda:
Hello:
A few time ago I finnish the C Clock exercisce and I wonder ...
Why the exercisce does no define a struct for hour and minute and a primitive for print it?
If clock dosn't create the structure, why do no call it clock_reset( ) or clock_init( )?
Thank you for your time!

ryanplusplus:
Hi @fjarboleda, that's a good approach and I'd welcome a change to the exercise that made it more object oriented. Would you be interested in creating a pull request that introduces these changes to the exercise?

fjarboleda:
Yes of course, but it my very first time in GitHub, I'll try to do my best...

wolf99:
closed by #393

PR #393, opened 2019-07-03
Update clock to v2.4.0

wolf99:
Update the solution to v2.4.0 of exercise canonical data.
Also adjust arguments to struct per #374.

### DIFF
--- a/exercises/clock/src/clock.h
+++ b/exercises/clock/src/clock.h
@@ -1,7 +1,7 @@
 #ifndef CLOCK_H
 #define CLOCK_H
 
-#include<stdbool.h>
+#include <stdbool.h>
 
 #define MAX_STR_LEN (5 + 1)     // "##:##\0"
 
--- a/exercises/clock/src/clock.h
+++ b/exercises/clock/src/clock.h
@@ -5,15 +5,13 @@
 
 #define MAX_STR_LEN (5 + 1)     // "##:##\0"
 
-typedef char clock_text_t[MAX_STR_LEN];
 typedef struct {
-   int hour;
-   int minute;
+   char text[MAX_STR_LEN];
 } clock_t;
 
-void clock_create(clock_text_t clock_text, clock_t clock);
-void clock_add(clock_text_t clock_text, int minute_offset);
-void clock_subtract(clock_text_t clock_text, int minute_offset);
+clock_t clock_create(int hour, int minute);
+clock_t clock_add(clock_t clock, int minute_add);
+clock_t clock_subtract(clock_t clock, int minute_subtract);
 bool clock_is_equal(clock_t a, clock_t b);
 
 #endif
--- a/exercises/clock/src/clock.h
+++ b/exercises/clock/src/clock.h
@@ -1,11 +1,19 @@
 #ifndef CLOCK_H
 #define CLOCK_H
 
+#include<stdbool.h>
+
 #define MAX_STR_LEN (5 + 1)     // "##:##\0"
 
-typedef char time_text_t[MAX_STR_LEN];
+typedef char clock_text_t[MAX_STR_LEN];
+typedef struct {
+   int hour;
+   int minute;
+} clock_t;
 
-void clock(time_text_t time_text, int hour, int minute);
-void clock_add(time_text_t time_text, int minute_offset);
+void clock_create(clock_text_t clock_text, clock_t clock);
+void clock_add(clock_text_t clock_text, int minute_offset);
+void clock_subtract(clock_text_t clock_text, int minute_offset);
+bool clock_is_equal(clock_t a, clock_t b);
 
 #endif
--- a/exercises/clock/src/example.c
+++ b/exercises/clock/src/example.c
@@ -23,7 +23,7 @@ void clock_create(clock_text_t clock_text, clock_t clock)
 
 void clock_add(clock_text_t clock_text, int minute_offset)
 {
-   clock_t clock = { 0 };
+   clock_t clock = { 0, 0 };
    sscanf(clock_text, "%02d:%02d", &clock.hour, &clock.minute);
    clock.minute += minute_offset;
    clock_create(clock_text, clock);
@@ -31,7 +31,7 @@ void clock_add(clock_text_t clock_text, int minute_offset)
 
 void clock_subtract(clock_text_t clock_text, int minute_offset)
 {
-   clock_t clock = { 0 };
+   clock_t clock = { 0, 0 };
    sscanf(clock_text, "%02d:%02d", &clock.hour, &clock.minute);
    clock.minute -= minute_offset;
    clock_create(clock_text, clock);
--- a/exercises/clock/src/example.c
+++ b/exercises/clock/src/example.c
@@ -1,45 +1,57 @@
 #include <stdio.h>
 #include "clock.h"
 
-static void normalize_clock(clock_t * clock)
+static const char CLOCK_FORMAT[] = "%02d:%02d";
+
+static void normalize_clock(int *hour, int *minute)
 {
-   while (clock->minute < 0) {
-      clock->minute += 60;
-      clock->hour -= 1;
+   while (*minute < 0) {
+      *minute += 60;
+      *hour -= 1;
    }
-   clock->hour += clock->minute / 60;
-   while (clock->hour < 0) {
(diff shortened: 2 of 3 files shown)


wolf99:
Hmm tests failing in Clang for missing struct initializer value, a la

    struct {
        int a;
        int b;
    } foo_t;

    foo_t foo = { 0 };   // struct member b is not explicitly initialized

This should be acceptable according to https://en.cppreference.com/w/c/language/struct_initialization

Is this a bug on Clang?

ryanplusplus:
> Hmm `-Wmissing-field-initializers` is turned on for Clang but not GCC?

I've run into this before and I find it pretty annoying. It's nice to be able to zero-initialize everything without needing to list out all the members.

exercises/clock/src/clock.h
```diff
@@ -1,11 +1,19 @@
...
+   int hour;
+   int minute;
+} clock_t;
 
-void clock(time_text_t time_text, int hour, int minute);
-void clock_add(time_text_t time_text, int minute_offset);
+void clock_create(clock_text_t clock_text, clock_t clock);
+void clock_add(clock_text_t clock_text, int minute_offset);
+void clock_subtract(clock_text_t clock_text, int minute_offset);
+bool clock_is_equal(clock_t a, clock_t b);
```
ryanplusplus: The API here still feels pretty unintuitive. Given that the `clock_text_t` type hides the fact that `clock_text` is actually an output parameter, I actually had to do a double-take to understand `clock_create`. Additionally, using `clock_text_t` instead of `clock_t` for the other methods feels a bit wonky.

How about something like this?

```c
clock_t clock_create(int hour, int min);
clock_text_t clock_to_string(clock_t clock); // I'm pretty sure this will just work
clock_t clock_add(clock_t clock, int minute_offset);
clock_t clock_subtract(clock_t clock, int minute_offset);
bool clock_is_equal(clock_t a, clock_t b);
```

I could go either way on using `clock_t` as a value or passing by reference so imagine that part however you'd prefer. This fully eliminates the use of the rendered state of a clock (a `clock_text_t`) as an awkward stand-in for a clock (ie: a `clock_t`).

Thoughts?

exercises/clock/test/test_clock.c
```diff
@@ -16,406 +15,524 @@ void tearDown(void)
...
+   TEST_ASSERT_EQUAL_STRING(expected_clock_text, actual_clock_text);
 }
 
 // the tests
 static void test_on_the_hour(void)
 {
-   int hour = 8;
-   int minute = 0;
+   clock_t clock = { 8, 0 };
    char *expected_clock_text = "08:00";
```
ryanplusplus: This can be written as

```c
clock_text_t expected_clock_text = "08:00";
```

Which will help us eliminate some additional type weirdness.

exercises/clock/test/test_clock.c
```diff
@@ -16,406 +15,524 @@ void tearDown(void)
...
+static void test_compare_clocks_with_negative_hour_that_wraps(void)
+{
+   TEST_IGNORE();
+   clock_t clock1 = { 17, 3 };
+   clock_t clock2 = { -31, 3 };
+
+   TEST_ASSERT_TRUE(clock_is_equal(clock1, clock2));
+}
+
+static void
```
ryanplusplus: I kind of hate our formatter

exercises/clock/src/clock.h
```diff
@@ -1,11 +1,19 @@
 #ifndef CLOCK_H
 #define CLOCK_H
 
+#include<stdbool.h>
```
ryanplusplus:
```suggestion
#include <stdbool.h>
```

exercises/clock/src/clock.h
```diff
@@ -1,11 +1,19 @@
...
+
 #define MAX_STR_LEN (5 + 1)     // "##:##\0"
```
ryanplusplus: Neat trick:

```c
enum
{
  MaxStringLength = strlen("##:##") + 1
};
```

exercises/clock/src/clock.h
wolf99: With you on the output being more intuitive as the return value rather than an argument - but it does mean that the student will need to create the string from scratch. On the other hand it reduces the string handling to the one `clock_to_string()` method. I'll code it up and see what it looks like 😃

exercises/clock/test/test_clock.c
wolf99: The 3 space indent catches me out every time 😖 
I know it's the Linux style but it is idiosyncratic in some parts.
Scope to adjust to more generally accepted formatting? (e.g. 4 space indent, 120 char line, etc)

exercises/clock/src/clock.h
wolf99: Fancy ✨  .
But is it clear enough for beginners what is actually happening?

exercises/clock/test/test_clock.c
wolf99: Oh, for my intital `clock_t clock = { .hour = 12, .minute = 5 };` it did `clock_t clock = {.hour = 12,.minute = 5 };`  🙅‍♂

exercises/clock/src/clock.h
wolf99: I recall the reason that the strings are the output -> it is per the exercise specification.
If we stick to but retain the output value by return we end up with:

    clock_test_t clock_create(int hour, int minute);
    clock_text_t clock_add(clock_text_t clock, int minute offset);
    clock_text_t clock_subtract(clock_text_t clock, int minute offset);
    bool clock_is_equal(clock_text_t a, clock_text_t b);

The `clock_t` type disappears from the interface again.

exercises/clock/test/test_clock.c
ryanplusplus: I'm mostly just complaining to complain. It's hard to find a formatter/configuration that doesn't have some sort of odd wart.

exercises/clock/src/clock.h
ryanplusplus: Probably not, but I wanted to share anyway :slightly_smiling_face:

exercises/clock/src/clock.h
ryanplusplus: One option is to simply deviate a bit from the specifications. I think that we still meet the intent of the cases even if we deviate from the interface a bit. For example, see the sample solution for this problem in Lua: https://github.com/exercism/lua/blob/master/exercises/clock/example.lua.

However, if we want to match the upstream problem specification more closely, I like your last suggestion. It's awkward to have a string representation be the primary interchange type, but at least the tension between `clock_text_t` and `clock_t` has been resolved and it's that there _is_ a single dominant representation.

I'm happy with either approach.

wolf99: Neglected to think the string version completely through, it would require the student to manage memory allocation and freeing. As to currently an array is used as the input/output parameter. Changing to returning the string requires changing to pointers which introduces memory handling 😒

ryanplusplus: You should can pass/return an array by value if you wrap it in a struct :+1:

wolf99: Of course 🙈

wolf99:
Closes #374

exercises/clock/src/example.c
```diff
@@ -15,17 +17,41 @@ static void normalize_clock(int *hour, int *minute)
...
-   int minute;
-   sscanf(time_text, "%02d:%02d", &hour, &minute);
-   minute += minute_offset;
-   clock(time_text, hour, minute);
+   int ah = 0, am = 0, bh = 0, bm = 0;
+   to_int(a, &ah, &am);
+   to_int(b, &bh, &bm);
+   normalize_clock(&ah, &am);
+   normalize_clock(&bh, &bm);
+   return (ah == bh) && (am == bm);
```
ryanplusplus: Can't you use strcmp here?

ryanplusplus:
Looks good with one simplification :+1:

exercises/clock/src/example.c
wolf99: Need to convert to `int` for the normalisation. E.g. `strcmp("24:00", "00:00")` does not return true, but for a clock it should. The canonical tests specifically test for these scenarios.

PR #394, opened 2019-07-07
run-tests: clone only necessary exercises

wolf99:
Currently the run-tests bash script clones all exercises to a build, even when only running specific exercises. (e.g. `$ ./bin/run-tests acronym beer-song`)

This PR adjusts the script to only copy the exercises selected for testing.


ryanplusplus:
:+1:

PR #395, opened 2019-07-14
Update the *sieve* exercise.

bgraf:
Rationale:
- Signature used relied on some array of fixed size
- Types used are rather arbitrary
- Test cases exhibited bad style

### DIFF
--- a/exercises/sieve/src/example.c
+++ b/exercises/sieve/src/example.c
@@ -1,50 +1,42 @@
-#include <string.h>
-#include <stdlib.h>
-#include <math.h>
 #include "sieve.h"
 
-unsigned int sieve(const unsigned int limit, primes_array_t primes)
+#include <math.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+uint32_t sieve(uint32_t limit, uint32_t * primes, size_t max_primes)
 {
-   unsigned int number_of_primes = 0;
-
-   // clear the results
-   memset(primes, 0, sizeof(*primes));
-
-   if (limit > 1) {
-      //allocate 1 more than limit for convenience so the number and the index are same
-      unsigned char *number_is_prime = malloc(limit + 1);
-      memset(number_is_prime, 1, limit + 1);
-
-      unsigned int max_factor = sqrt(limit) + 1;
-
-      // mark 0 and 1 as not prime
-      for (unsigned int i = 0; i < 2; i++) {
-         number_is_prime[i] = 0;
-      }
-
-      // mark the remaining numbers in the array according to the algo.
-      for (unsigned int index = 2; index < max_factor;) {
-         // mark all of multiples that can't be prime
-         for (unsigned int non_prime_index = (2 * index);
-              non_prime_index < (limit + 1); non_prime_index += index) {
-            number_is_prime[non_prime_index] = 0;
-         }
-
-         // adjust the index
-         do {
-            index++;
-            if (number_is_prime[index]) {
-               break;
-            }
-         } while (index <= max_factor);
-      }
-      // collect and count the primes found
-      for (unsigned int i = 1; i < limit + 1; i++) {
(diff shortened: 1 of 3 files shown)


exercises/sieve/src/sieve.h
```diff
@@ -1,10 +1,13 @@
...
-#define MAX_LIMIT_TESTED (1000)
-
-typedef unsigned int primes_array_t[MAX_LIMIT_TESTED];
-
-unsigned int sieve(const unsigned int limit, primes_array_t primes);
+#include <stdint.h>
+#include <stddef.h>
+
+/// Calculate at most `max_primes` prime numbers in the interval [2,limit] 
+/// using the Sieve of Eratosthenes and store the prime numbers in `primes`
```
wolf99: Should not mention implementation requirements here, this should be covered by the readme (or additional hints to the readme)

exercises/sieve/src/example.c
```diff
@@ -1,50 +1,41 @@
...
-      unsigned char *number_is_prime = malloc(limit + 1);
-      memset(number_is_prime, 1, limit + 1);
+uint32_t sieve(uint32_t limit, uint32_t * primes, size_t max_primes)
+{
+   if (limit <= 1) {
+      // no primes that small
+      return 0;
+   }
+   // allocate one boolean for each number in [0,limit]
+   bool *data = calloc(limit + 1, sizeof(*data));
```
wolf99: Can you use a better name than the rather generic `data`?

wolf99:
Thanks for the contribution @bgraf 

General question (@ryanplusplus @patricksjackson et al): I realise we have specified in the style guide that the stdlib types are preferred over regular types, but are we expecting that students would be sufficiently au fait with the stdlib types? I'm not really clear on what level of C knowledge we're expecting any more 😕

exercises/sieve/src/sieve.h
bgraf: The exercise is called *sieve* and the README explicitly states that students should use the algorithm. But the majority of published solutions I've seen do not use the Sieve of E. at all. Given that the function is also named `sieve`, I wouldn't consider the requirement of using the Sieve of E. an implementation detail.

exercises/sieve/src/example.c
bgraf: Changed e7c5ba9a11afd4640fc1f573d047aa56c426e8f1

exercises/sieve/src/sieve.h
wolf99: Yes it is covered by the readme. The readme is populated automatically from the exercism/problem-specifications repo. We should not duplicate the information in the exercise where it is not in synch with that repo. If students are not reading the readme, that is a separate issue (that should also be dealt with but is not specific to this exercise)

ryanplusplus:
> General question (@ryanplusplus @patricksjackson et al): I realise we have specified in the style guide that the stdlib types are preferred over regular types, but are we expecting that students would be sufficiently au fait with the stdlib types? I'm not sure I'm clear on what level of C knowledge we're expecting any more.

It's not super clear to me either, but I think that's largely a product of the fact that the C community is large, diverse, and fragmented. We're likely to get a little bit of everything. I personally prefer using fixed-size types (I work on microcontrollers that cost in the low single digits USD) and I also think that `uint32_t` is both relatively clear and easily googleable. I don't have any heartburn about the `stdint` types being used here.

exercises/sieve/src/sieve.h
ryanplusplus: @wolf99 I see your point, but we're unlikely to contradict the README in this case. I suspect there will still be plenty of non-sieve solutions, but this seems like reasonable documentation for the function itself.

ryanplusplus:
This looks good to me. @wolf99 go ahead and merge if you are happy as well.

PR #398, opened 2019-07-27
Fix word length check in word-count

### LINKED ISSUE #397: word-count: ignore punctuation test broken due to change to test_assert_equal_string_len
The change to using Unity TEST_ASSERT_EQUAL_STRING_LEN has the unfortunate side effect of not comparing the characters past the length of the expected result and as such will match any string that has the same prefix as the expected solution. 

I assume this can be mitigated by adding one to the comparison length in the call to check_solution to confirm that the actual result does not differ at the null terminator.

ryanplusplus:
Fixes #397 

Thanks @lukebiggerstaff for finding the issue


wolf99:
LGTM 👍

Issue #390, opened 2019-06-29
Better trade-offs for mentors 


bgraf:

ryanplusplus:
I maintain and mentor Lua. I only maintain but don't mentor C for this exact reason.

patricksjackson:
I see two separate things here:

First, I suggest mentioning this on the Slack channel, I think more active mentors hang out there and may be able to give advice/commiserate/have solutions? I admit I don't actually mentor these days, though I don't know about the other maintainers.

Second, You bring up a point that I've been...reluctant to bring up for a while. I think that some of the exercises are just not appropriate for learning on the C track, and if I could go back I would've tried to make an effort to just not include them. The same exercises don't work for learning Javascript and C, and we shouldn't try to pretend that they're appropriate.

Mainly I'm concerned about some of the string manipulation exercises. Those are easy in any language that has basic string functions, but in C they become a nightmare. And you generally don't _learn_ anything after the first one, they're just tedious. IMO we should have one of the complicated string manipulation exercises and just delete the rest.

Even things like palindrome-products become kind of absurd with C. The core math is the easy part, since you have to figure out from the test file alone [this header](https://github.com/exercism/c/blob/master/exercises/palindrome-products/src/example.h), which uses an error message field and two linked lists for the return value. The core problem is a small part of the example solution, and the rest is contrived just so we can shoehorn the return values through the C interface.

I don't know how to fix it other than to be more willing to say 'no' to a problem in general, or to be willing to deviate from the problem-specifications test suites more. That said I never actually write code in C unless I absolutely have to, so I've been deferring to others here who may have more experience with C directly. I work in C++ mainly.

wolf99:
First I'll say, I don't mentor as I am unable to commit the time.

Second I agree that complex string exercises in C are tedious. 

From one side we can say that if a student wishes to go on and write any non-trivial program that involves string manipulation in C, it will be mostly complexity and tedium - a student should learn and understand this to inform future choices.

On the other hand, this is not what Exercism is about; Exercism is "supposed" to be fun, quick and allow a student to easily progress past typical tutorial toy problems to a next-level understanding of the language.

(I am a little conflicted as often such complexity and tedium is exactly what all non-trivial programs in C are like, so it is exactly what students should learn! 😄 )

[...]

PR #401, opened 2019-08-18
armstrong-numbers: add exercise

wolf99:


exercises/armstrong-numbers/src/example.c
```diff
@@ -0,0 +1,30 @@
...
+   }
+   return digit_count;
+}
+
+bool isArmstrongNumber(uint32_t candidate)
+{
+   bool result = false;
+   if (candidate < 10) {
+      result = true;
+   } else if (candidate >= 100) {
```
ryanplusplus: Are these cases necessary? Seems like it complicates the solution without any real benefit.

ryanplusplus:
Looks good, just one comment. This is at least a little bit of progress toward addressing #390.

exercises/armstrong-numbers/src/example.c
wolf99: All 1 digit numbers are Armstrong numbers but no 2 digit ones are. The cases shortcut the while loop which are unnecessary for these cases.
Technically, yes the solution would work without them... but the logic is evident from the test cases and exercise description.

PR #402, opened 2019-08-19
darts: add exercise

wolf99:
In aid of #390 
Not sure of good ordering in `config.json` - maybe best to decide this with a view of the whole on #390?


wolf99:
Need to merge #401 first 🙄

PR #403, opened 2019-08-20
darts: add exercise

wolf99:
Towards #390 
Unsure of best place for these exercises in `config.json` - maybe that can be decided on #390?

### DIFF
--- a/config.json
+++ b/config.json
@@ -30,6 +30,19 @@
       ]
     },
     {
+      "slug": "darts",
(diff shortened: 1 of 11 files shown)


wolf99:
Don't know why verify-indent is failing, ran the following on all files before and after push with no change in the files

```bash
$ indent -linux -i3 -nut src/example.h
$ diff src/example.h src/example.h~
$
```

ryanplusplus:
Looks great to me. I'm happy with it being unlocked by Armstrong Numbers. If we change our minds later it's easily modified.

PR #407, opened 2019-09-08
resistor-color-trio: add exercise

wolf99:
For #390

### DIFF
--- a/exercises/resistor-color-trio/src/example.h
+++ b/exercises/resistor-color-trio/src/example.h
@@ -1,0 +1,36 @@
+#ifndef RESISTOR_COLOR_TRIO_H
+#define RESISTOR_COLOR_TRIO_H
+
+#include <stdint.h>
+
+typedef enum {
+   BLACK = 0,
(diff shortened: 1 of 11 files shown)


exercises/resistor-color-trio/src/example.h
```diff
@@ -0,0 +1,36 @@
...
+   GREEN = 5,
+   BLUE = 6,
+   VIOLET = 7,
+   GREY = 8,
+   WHITE = 9
+} resistor_band_t;
+
+typedef enum {
+   OHMS = 0,
+   KILOOHMS = 1,
```
ryanplusplus: Amazingly I don't think I've ever seen this fully written out and it's really weird to see both "o"s back to back O_o

ryanplusplus:
LGTM :+1:

exercises/resistor-color-trio/src/example.h
wolf99: I actually had to double check it it looked so weird!

PR #408, opened 2019-09-08
square-root: add exercise

wolf99:
Adds a new exercise to calculate the square root of a natural radicand.

Because the tests use only natural numbers up to 255, the exercise can be implemented via a simplified algorithm that uses bit shifting.

Will add readme after exercise addition PR is merged in problem specifications.

### DIFF
--- a/exercises/square-root/src/example.c
+++ b/exercises/square-root/src/example.c
@@ -12,8 +12,7 @@ uint16_t squareRoot(uint16_t radicand)
       if (radicand >= result + bit) {
          radicand -= result + bit;
          result = (result >> 1) + bit;
-      } else
-      {
+      } else {
          result >>= 1;
       }
       bit >>= 2;
--- a/exercises/square-root/src/example.c
+++ b/exercises/square-root/src/example.c
@@ -1,0 +1,22 @@
+#include "square_root.h"
+
+uint16_t squareRoot(uint16_t radicand)
+{
+   uint16_t result = 0;
+   uint16_t bit = 1 << 14;
+
+   while (bit > radicand)
+      bit >>= 2;
+
+   while (bit != 0) {
+      if (radicand >= result + bit) {
+         radicand -= result + bit;
+         result = (result >> 1) + bit;
+      } else
+      {
+         result >>= 1;
+      }
+      bit >>= 2;
+   }
+   return result;
+}
--- a/config.json
+++ b/config.json
@@ -257,6 +257,18 @@
       ]
     },
     {
+      "slug": "square-root",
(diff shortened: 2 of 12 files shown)


exercises/square-root/src/example.c
```diff
@@ -0,0 +1,21 @@
...
+      if (radicand >= result + bit) {
+         radicand -= result + bit;
+         result = (result >> 1) + bit;
+      } else {
+         result >>= 1;
+      }
+      bit >>= 2;
+   }
+   return result;
+}
```
ryanplusplus: Is there a writeup on this algorithm available somewhere? I'm curious and would like to know more.

wolf99:
Seems the `problem-specifications` repo is not accepting new exercises at the moment, but is recommending that tracks go ahead and implement any new exercise they like within the track and then add to the specifications when it starts accepting again.

So the question is do we want to add this exercise if so doing diverges from the existing specifications?

I don't mind overly if the exercise is put in WIP until whenever it can be added to the specifications (this is what has been done on the specification PR).

ryanplusplus:
@wolf99 I haven't been following the discussion upstream and didn't realize they were no longer accepting new exercises. Is there any indication of how long that might last?

wolf99:
@ryanplusplus Nope. Have asked on the related issue on the repo (https://github.com/exercism/problem-specifications/issues/1560#issuecomment-531685923)

However, I note a recent exercism blog post by Jeremey Walker mentions a program for structuring project anatomy is intended to cover a period of 12 months (https://exercism.io/blog/redesigning-tracks-in-partnership-with-chicago-university-and-sloan-foundation).

ryanplusplus:
@wolf99 okay, sounds like we'll be stuck in a holding pattern for a long time if we don't move forward on the C track. Will this actually show up properly in the web UI if we don't have the upstream stuff taken care of? That's the only lingering concern I have.

wolf99:
@ryanplusplus I think it should yes. For final confirmation I think jut try it and see really 🤷‍♂  

I will have to add a `hints.md` and custom `readme.md` file until that PR can be merged though. I will push that shortly.

SleeplessByte:
Hi there! You may add track specific exercises WITHOUT having the problem spec one merged. There are no technical limitations here.

Make sure there is a uuid and that the slug is unique. Request an icon. The usual.

wolf99:
I've added the exercise README and the icon is live https://github.com/exercism/website-icons/issues/34.

I think this is ready for final review/merge

ryanplusplus:
Sorry @wolf99 for not getting to this sooner

Issue #413, opened 2019-10-06
Memory Allocation on Saddle Points Exercise


gabriel376:
I just solved Saddle Points exercise on C track and could not resolve it without a memory leak, using valgrind to check. I have checked all solutions published and all have the same memory leak issue.

The structs of this solution are:
```
typedef struct {                                                                
   size_t row, column;                                                          
} saddle_point_t;                                                               
                                                                                
typedef struct {                                                                
   size_t count;                                                                
   saddle_point_t points[];                                                     
} saddle_points_t; 
```

In file [test_saddle_points.c](https://github.com/exercism/c/blob/master/exercises/saddle-points/test/test_saddle_points.c), `free()` is called just for the main struct `saddle_points_t`, but not for the field `points`.

The [example.c](https://github.com/exercism/c/blob/master/exercises/saddle-points/src/example.c) uses `realloc()` and calculates the size of the main struct + points field, so the `free()` works fine and there is no memory leak.
```
realloc(saddle_points,                                                                                                                                                    
            sizeof(saddle_points_t) +       
            sizeof(saddle_point_t) *        
            ++count);
```

However, I think this approach is not very common, and we can see that all students have submit solutions using `calloc()` for the main struct and another `calloc()` for points field. It's part of mentoring program to check if the solution handles memory allocation correctly and we provide a valgrind option on `make`.

I was expecting the test to free fields point too:
```
free(actual->points)
```

Another minor issue is that students should implement function `saddlePoints`, but the naming convention we have been using for C track is snake_case, so it should be `saddle_points`.

wolf99:
Hi @gabriel376 

Good catch!

I added this exercise. On the PR it was mentioned that we are happy we have a solution that uses `realloc()` 😄 . 

However, if students are not finding a path to implementing it this way and are instead using ways that are "incorrect", then short of adding valgrind testing to solutions I don't see a way to resolve this of the top of my head.

wolf99:
Wondering how compilation is working using `calloc()` on a flexible array member of a struct 🤔

ryanplusplus:
@wolf99 I think that we should probably move in the direction of always having the `free` be done inside of the code under test by exposing a `whatever_free`/`whatever_release`/`whatever_destroy` method. This puts fewer restrictions on the implementation.

wolf99:
Good call @ryanplusplus

gabriel376:
Thank you all

PR #414, opened 2019-10-06
add free() to points and rename function

gabriel376:
fix https://github.com/exercism/c/issues/413

### DIFF
--- a/exercises/saddle-points/src/example.c
+++ b/exercises/saddle-points/src/example.c
@@ -57,8 +57,8 @@ static bool is_least_in_column(int16_t row, int16_t column, size_t rows,
    return true;
 }
 
-saddle_points_t *saddlePoints(size_t rows, size_t columns,
-                              uint8_t matrix[][columns])
+saddle_points_t *saddle_points(size_t rows, size_t columns,
+                               uint8_t matrix[][columns])
 {
    saddle_points_t *saddle_points = noSaddlePoints();
 
@@ -70,4 +70,9 @@ saddle_points_t *saddlePoints(size_t rows, size_t columns,
                saddle_points = add(saddle_points, row + 1, column + 1);
 
    return saddle_points;
+}
+
+void free_saddle_points(saddle_points_t * saddle_points)
+{
+   free(saddle_points);
 }
--- a/exercises/saddle-points/src/example.h
+++ b/exercises/saddle-points/src/example.h
@@ -13,7 +13,9 @@ typedef struct {
    saddle_point_t points[];
 } saddle_points_t;
 
-saddle_points_t *saddlePoints(size_t rows, size_t columns,
-                              uint8_t matrix[][columns]);
+saddle_points_t *saddle_points(size_t rows, size_t columns,
+                               uint8_t matrix[][columns]);
+
+void free_saddle_points(saddle_points_t * saddle_points);
 
 #endif
--- a/exercises/saddle-points/test/test_saddle_points.c
+++ b/exercises/saddle-points/test/test_saddle_points.c
@@ -44,19 +44,19 @@ static void test_single_saddle_point(void)
 
    size_t expected_count = 1;
    saddle_point_t expected_points[] = { {2, 1} };
-   saddle_points_t *actual = saddlePoints(3, 3, matrix);
+   saddle_points_t *actual = saddle_points(3, 3, matrix);
    check_count(expected_count, actual->count);
    check_points(expected_count, expected_points, actual->points);
-   free(actual);
+   free_saddle_points(actual);
 }
 
 static void test_empty_matrix_has_no_saddle_points(void)
 {
    TEST_IGNORE();               // delete this line to run test
    size_t expected_count = 0;
-   saddle_points_t *actual = saddlePoints(0, 0, NULL);
+   saddle_points_t *actual = saddle_points(0, 0, NULL);
    check_count(expected_count, actual->count);
-   free(actual);
+   free_saddle_points(actual);
 }
 
 static void test_lack_of_saddle_points_when_there_are_none(void)
@@ -69,9 +69,9 @@ static void test_lack_of_saddle_points_when_there_are_none(void)
    };
 
    size_t expected_count = 0;
-   saddle_points_t *actual = saddlePoints(3, 3, matrix);
+   saddle_points_t *actual = saddle_points(3, 3, matrix);
    check_count(expected_count, actual->count);
-   free(actual);
+   free_saddle_points(actual);
 }
 
 static void test_multiple_saddle_points_in_a_column(void)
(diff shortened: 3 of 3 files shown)


exercises/saddle-points/src/example.h
```diff
@@ -5,15 +5,16 @@
...
 
 typedef struct {
    size_t count;
-   saddle_point_t points[];
+   saddle_point_t *points;
 } saddle_points_t;
 
-saddle_points_t *saddlePoints(size_t rows, size_t columns,
-                              uint8_t matrix[][columns]);
+saddle_points_t *saddle_points(size_t nrows,
```
wolf99: What's `n`? For a non-English native this may actually be _more_confusing

exercises/saddle-points/src/example.h
```diff
@@ -5,15 +5,16 @@
...
+                               size_t ncols, uint8_t matrix[nrows][ncols]);
```
wolf99: Please use full words (i.e. `columns`), this is for clarity. We aren't limited on space 😄

exercises/saddle-points/test/test_saddle_points.c
```diff
@@ -44,18 +44,20 @@ static void test_single_saddle_point(void)
 
    size_t expected_count = 1;
    saddle_point_t expected_points[] = { {2, 1} };
-   saddle_points_t *actual = saddlePoints(3, 3, matrix);
+   saddle_points_t *actual = saddle_points(3, 3, matrix);
```
wolf99: Good catch on the function name 👍

exercises/saddle-points/src/example.c
```diff
@@ -1,73 +1,45 @@
...
-   uint8_t candidate = matrix[row][column];
-   for (size_t r = 0; r < rows; ++r)
-      if (matrix[r][column] < candidate)
-         return false;
    return true;
 }
 
-saddle_points_t *saddlePoints(size_t rows, size_t columns,
-                              uint8_t matrix[][columns])
+saddle_points_t *saddle_points(size_t nrows,
```
wolf99: Whats `n` prefix? See same comment below.

exercises/saddle-points/src/example.c
```diff
@@ -1,73 +1,45 @@
...
+                               size_t ncols, uint8_t matrix[nrows][ncols])
```
wolf99: `cols` -> `columns` parameters. Also see same comment below

exercises/saddle-points/src/example.c
```diff
@@ -1,73 +1,45 @@
 #include "saddle_points.h"
+
 #include <stdlib.h>
 #include <stdbool.h>
 
-static saddle_points_t *noSaddlePoints(void)
+static bool is_saddle_point(size_t nrows,
+                            size_t ncols,
+                            size_t row,
+                            size_t col, uint8_t matrix[nrows][ncols])
```
wolf99: Same here with `n` and `cols` :)

exercises/saddle-points/src/example.h
```diff
@@ -5,15 +5,16 @@
...
 typedef struct {
-   size_t row, column;
+   size_t row;
+   size_t column;
 } saddle_point_t;
 
 typedef struct {
    size_t count;
-   saddle_point_t points[];
+   saddle_point_t *points;
```
wolf99: The aim was to specifically introduce flexible array members rather than just one more pointer exercise.

exercises/saddle-points/src/example.c
```diff
@@ -1,73 +1,45 @@
...
-   saddle_points_t *saddle_points = noSaddlePoints();
 
-   if (rows > 0 && columns > 0 && matrix != NULL)
-      for (size_t row = 0; row < rows; ++row)
-         for (size_t column = 0; column < columns; ++column)
-            if (is_greatest_in_row(row, column, columns, matrix)
-                && is_least_in_column(row, column, rows, columns, matrix))
-               saddle_points = add(saddle_points, row + 1, column + 1);
+   saddle_points_t *saddle = calloc(1, sizeof(saddle_points_t));
+   saddle->points = calloc(nrows * ncols, sizeof(saddle_point_t));
```
wolf99: doesn't this allocate too many items? `rows * columns` must be significantly more than the number of actual saddle points in most cases.

exercises/saddle-points/src/example.h
wolf99: Same goes for other occurrences of the same in the change

exercises/saddle-points/src/example.h
wolf99: Same goes for other occurrences of the same in the change

exercises/saddle-points/src/example.c
wolf99: I don't think we have a way to enforce this as a requirement in the test's though.
Interesting question: should the example here show best practice, or simply adhere to the tests?
More pertinent question: is there a way to do this that removes the extra allocated memory yet is still accessible to students?

exercises/saddle-points/src/example.h
gabriel376: I am not a C expert, but it seems that if we use flexible array, we can't use memory allocation on it. Instead, we should do a single allocation with size of struct + size of array, just as the code it is right now. If we choose to keep flexible array declaration, which seems to be easier and simpler, we can discard this PR.
I don't see how the tests can handle all cases. If a solution declares points as pointer and we don't free it, there will be a memory leak, but if the user declares points as a flexible array and we try to free it, we will get an invalid pointer error.
Maybe we can provide a boilerplate in header file.

wolf99: Maybe I need to look at the student solutions to date to fully understand the issue. I might get to this tomorrow evening (I will have to do submissions for all the other exercises up to that point to see them, so it won't happen this evening as it's late in my timezone).

exercises/saddle-points/test/test_saddle_points.c
wolf99: Had to go change the same for all the other exercises I've submitted recently! 🤦‍♂

exercises/saddle-points/src/example.c
ryanplusplus: IMO we should just move the `free` internal to this module and expose it via a public function. Then we can have a correct solution that doesn't impose any extra requirements on submitted solutions. It would also be a relatively minor change to the existing solution.

exercises/saddle-points/src/example.h
ryanplusplus: @wolf99 I think there's a way to declare yourself a maintainer/mentor so that you can view all exercises and all submitted solutions for a track. I have done it on C, but don't recall how.

exercises/saddle-points/src/example.c
gabriel376: do you mean to provide a function in header file, so students implement their own free function ?
```
void free_saddle_points(saddle_points_t saddle_points);
```

ryanplusplus: @gabriel376 yes. That way the implementation owns allocating and freeing.

gabriel376: great!
I will update my PR to implement that
thank you!

wolf99:
Good work @gabriel376 thanks for bringing this up and following it 👍

gabriel376:
Thank you all

ryanplusplus:
Thanks for the contribution!

PR #417, opened 2019-10-12
rational-numbers: add exercise

wolf99:
Towards #390

### DIFF
--- a/exercises/rational-numbers/src/example.c
+++ b/exercises/rational-numbers/src/example.c
@@ -45,8 +45,9 @@ rational_t add(rational_t r1, rational_t r2)
    int16_t denominator = 1;
    if (numerator != 0)
       denominator = r1.denominator * r2.denominator;
-   return (rational_t) {
-   numerator, denominator};
+   return reduce((rational_t) {
+                 numerator, denominator}
+   );
 }
 
 rational_t subtract(rational_t r1, rational_t r2)
@@ -57,8 +58,9 @@ rational_t subtract(rational_t r1, rational_t r2)
    int16_t denominator = 1;
    if (numerator != 0)
       denominator = r1.denominator * r2.denominator;
-   return (rational_t) {
-   numerator, denominator};
+   return reduce((rational_t) {
+                 numerator, denominator}
+   );
 }
 
 rational_t multiply(rational_t r1, rational_t r2)
@@ -80,8 +82,9 @@ rational_t divide(rational_t r1, rational_t r2)
       denominator = -denominator;
       numerator = -numerator;
    }
-   return (rational_t) {
-   numerator, denominator};
+   return reduce((rational_t) {
+                 numerator, denominator}
+   );
 }
 
 rational_t absolute(rational_t r)
@@ -89,8 +92,9 @@ rational_t absolute(rational_t r)
    // |a/b| = |a|/|b|
    int16_t numerator = abs(r.numerator);
    int16_t denominator = abs(r.denominator);
-   return (rational_t) {
-   numerator, denominator};
+   return reduce((rational_t) {
+                 numerator, denominator}
+   );
 }
 
 rational_t exp_rational(rational_t r, uint16_t n)
@@ -105,14 +109,17 @@ rational_t exp_rational(rational_t r, uint16_t n)
    // r^n = (a^|n|)/(b^|n|)
    int16_t numerator = pow(r.numerator, n);
    int16_t denominator = pow(r.denominator, n);
-   return (rational_t) {
-   numerator, denominator};
+   return reduce((rational_t) {
+                 numerator, denominator}
+   );
 }
 
 float exp_real(uint16_t n, rational_t r)
 {
    if (r.numerator == 0)        // shortcut power of 0
       return 1.0f;
+
+   r = reduce(r);
 
    // x^(a/b) = root(x^a, b), where root(p, q) is the qth root of p
    float p = pow(n, r.numerator);
--- a/exercises/rational-numbers/src/example.c
+++ b/exercises/rational-numbers/src/example.c
@@ -1,7 +1,7 @@
 #include "rational_numbers.h"
 #include <math.h>
 
-static int16_t absi(int16_t n)
+static int16_t absolute_int(int16_t n)
 {
    const int16_t mask = n >> 15;
    return (n + mask) ^ mask;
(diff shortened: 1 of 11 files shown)


wolf99:
Looks like clang counts `abs()` as a builtin.

*Secretly happy that I have an excuse to update the problem specification to use full words* 😄 

Will open PR for change on the specification, then continue here when that get's merged

exercises/rational-numbers/test/test_rational_numbers.c
```diff
@@ -0,0 +1,439 @@
...
+   TEST_IGNORE();
+   rational_t r = { 1, 2 };
+   rational_t expected = { 1, 2 };
+   rational_t actual = abs(r);
+   check_rational_numbers_equal(expected, actual);
+}
+
+static void
+    test_absolute_value_of_positive_rational_with_negative_numerator_and_denominator
+    (void) {
```
wolf99: Well this sucks

wolf99:
Recommendation from 
- https://github.com/exercism/problem-specifications/issues/1603
- https://github.com/exercism/problem-specifications/pull/1604

is that this is not a bug and the track should just diverge from the specification here. 
Have pushed changes to suit this, so removing from WIP

exercises/rational-numbers/src/example.c
```diff
@@ -0,0 +1,125 @@
...
+   const int16_t mask = n >> 15;
+   return (n + mask) ^ mask;
+}
+
+static float nth_root(float x, float n)
+{
+   return pow(x, 1. / n);
+}
+
+static int16_t greatestCommonDivisor(int n, int m)
```
patricksjackson: This function name should be snake_case :)

exercises/rational-numbers/test/test_rational_numbers.c
```diff
@@ -0,0 +1,439 @@
...
+}
+
+void tearDown(void)
+{
+}
+
+static void check_rational_numbers_equal(rational_t expected, rational_t actual)
+{
+   TEST_ASSERT_EQUAL_INT16(expected.numerator, actual.numerator);
+   TEST_ASSERT_EQUAL_INT16(expected.denominator, actual.denominator);
```
patricksjackson: Is there a way to have an error message to make it easier to know which of the numerator/denominator is wrong? I think there's a version of the assert with a string as a third argument.

exercises/rational-numbers/test/test_rational_numbers.c
patricksjackson: The formatting in the example solution is weird too, when constructing the return values it doesn't look like it needs to be split over two lines.

exercises/rational-numbers/src/example.c
```diff
@@ -0,0 +1,125 @@
+#include "rational_numbers.h"
+#include <math.h>
+
+static int16_t absolute_int(int16_t n)
+{
+   const int16_t mask = n >> 15;
+   return (n + mask) ^ mask;
+}
```
patricksjackson: Nit: Shouldn't this instead be the `abs()` library function?

exercises/rational-numbers/src/example.c
```diff
@@ -0,0 +1,125 @@
...
+
+rational_t add(rational_t r1, rational_t r2)
+{
+   // r1 + r2 = a1/b1 + a2/b2 = (a1 * b2 + a2 * b1) / (b1 * b2)
+   int16_t numerator =
+       r1.numerator * r2.denominator + r2.numerator * r1.denominator;
+   int16_t denominator = 1;
+   if (numerator != 0)
+      denominator = r1.denominator * r2.denominator;
+   return (rational_t) {
```
patricksjackson: For correctness, I'd assume that `add()`, `subtract()`, `multiply()` and `divide()` should all be fed through `reduce()` before returning, or is that not right?

I'd assume that 2/6 + 2/6 = 2/3, not 4/6.

exercises/rational-numbers/test/test_rational_numbers.c
wolf99: It shouldn't split the return according to line length.
Maybe it's doing that because it thinks `return (something) { blah }` looks like a function?

exercises/rational-numbers/test/test_rational_numbers.c
wolf99: Yes - I thought about adding it earlier but forgot! 🤦‍♂

exercises/rational-numbers/src/example.c
wolf99: 😆 😆  I can't believe I did this!

exercises/rational-numbers/src/example.c
wolf99: Hmm. It did occur to me but tests pass as is so I left it. 
IIRC the example implementation is not shown to users; so I guess there's somewhat of a philosophical question here - what is the purpose of this implementation?

Is it to be a canonical/reference implementation? 
Or just something that ensures the tests are constructed correctly?

SleeplessByte: > Or just something that ensures the tests are constructed correctly?

This! This file is just to make CI pass and to prove that an implementation is _possible_. It's not meant to be the _perfect_ solution :)

ryanplusplus: > > Or just something that ensures the tests are constructed correctly?
> 
> This! This file is just to make CI pass and to prove that an implementation is _possible_. It's not meant to be the _perfect_ solution :)

Counterpoint: I've seen many students refer to the repository to look for a "good" example solution so I try to strive for an example that can serve as a good reference solution. It may not be optimal in runtime or concision, but it should serve to instruct.

ryanplusplus: In this case I think it's an issue with the exercise definition itself. It would be nice if it required rationals to be in reduced form and the tests verified this.

SleeplessByte: > Counterpoint: I've seen many students refer to the repository to look for a "good" example solution so I try to strive for an example that can serve as a good reference solution. It may not be optimal in runtime or concision, but it should serve to instruct.

I think it's always fine to make a good example solution (we try it too), but I don't think it's a good counterpoint. It should be stated, in mentoring, explicitly, that the examples are just that: example solutions. Not good solutions. Students shouldn't be looking at them at all. I actively encourage mentors to discourage students from looking at them, and adding that they are not canonical / idiomatic examples makes that _easier_. The better they are, the harder it is "not to cheat" (for some people).

> In this case I think it's an issue with the exercise definition itself. It would be nice if it required rationals to be in reduced form and the tests verified this.

Absolutely agreed. I would suggest adding track-specific tests for this!

SleeplessByte: Ai. Good enough is always better than those with obvious problems 👍

ryanplusplus:
LGTM :+1:

PR #418, opened 2019-10-15
Correct incorrect arguments to UnityBegin()

wolf99:
I noticed that a few other exercises also had incorrect arguments passed to the `UnityBegin()` call.
This PR corrects them.
To paraphrase the Pokémon theme: hope I've caught 'em all!

### DIFF
--- a/exercises/anagram/test/test_anagram.c
+++ b/exercises/anagram/test/test_anagram.c
@@ -243,7 +243,7 @@ static void test_misleading_unicode_anagrams(void)
 
 int main(void)
 {
-   UnityBegin("anagram.c");
+   UnityBegin("test/test_anagram.c");
 
    RUN_TEST(test_no_matches);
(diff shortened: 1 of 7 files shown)


patricksjackson:
I validated that this is the full list. Can we just add a test to the CI to check?

``` bash
for f in exercises/*/test/test_*; do if ! grep -q UnityBegin\(\"test\/$(basename $f)\"\) $f; then echo "$f needs correct UnityBegin line"; fi; done
```

ryanplusplus:
> I validated that this is the full list. Can we just add a test to the CI to check?
> 
> ```shell
> for f in exercises/*/test/test_*; do if ! grep -q UnityBegin\(\"test\/$(basename $f)\"\) $f; then echo "$f needs correct UnityBegin line"; fi; done
> ```

Good idea. I'm going to merge and we can take care of that in a follow-up PR.

PR #419, opened 2019-10-16
void tearDown

babugyan:


ryanplusplus:
What is the intent of this change?

wolf99:
This seems a very spurious change at best.
(As such I would recommend that the PR be closed without merge if no response is forthcoming to Ryan's question within the next few days)

wolf99:
Checking the requester's profile, the other only other actions are all PRS all from the same period, all making similar spurious commits. No idea why but I can't think of any good reasons.
Given this and the lack of response to earlier questions, I'm closing this.

ryanplusplus:
Hacktoberfest

PR #420, opened 2019-10-17
Add test for argument to UnityBegin()

wolf99:
Based on the bash script written by @patricksjackson in https://github.com/exercism/c/pull/418#issuecomment-542425960


wolf99:
The checks are failing - correctly.
The new check introduced by this PR is detecting that the `crypto-square` exercise does not have the correct argument to `UnityBegin()`.

I can add a commit to this PR to fix that, or it can be done in a follow up PR?

ryanplusplus:
I'm certainly not a bash expert, but this looks good to me. Conveniently we can have a lot of confidence that this works because it already found a problem :)

patricksjackson:
Oh nice, the ERROR_COUNT is a good touch :+1:

PR #422, opened 2019-10-22
armstrong-numbers: added function declaration

computermouth:
This PR is simultaneously an issue, and a potential fix for that issue.

I've been going through these exercises with a couple friends, serving as their offline mentor while they go through these. Right off the bat we came to recognize that some of the exercises are shipped with the function declaration in the header, and others aren't.

It's particularly frustrating for them, since they're so new to the language, and they're not accustomed to the way GCC can throw ten or more errors, most not particularly informative, ( eg. "what's an empty translation unit?" ).

I thought I'd try and contribute and ask for some more strict standards on the C track. For example:

1. Function declarations in the header should either be there for all exercises or for none.
2. A central decision on types. If function declarations are provided, they should either all use stdbool for true/false functions, or none of them should. Similarly for functions with number parameters or return types should either all use stdint types, or none of them should.
3. It'd be lovely if there were a singular style with provided code. Snake case, camel case, etc.

I've noticed that most of the other languages don't provide function declarations, but other languages also don't have the waterfall of compiler errors, and other languages don't really reproduce the same experience of C's headers.

I'd be happy to do some cleanup work either way (removing or adding function decl's, etc.). I just think it'd be a huge improvement to add some consistency.

### DIFF
--- a/exercises/armstrong-numbers/src/armstrong_numbers.h
+++ b/exercises/armstrong-numbers/src/armstrong_numbers.h
@@ -1,4 +1,6 @@
 #ifndef ARMSTRONG_NUMBERS
 #define ARMSTRONG_NUMBERS
 
+int is_armstrong_number(int candidate);
+
 #endif


ryanplusplus:
> Function declarations in the header should either be there for all exercises or for none.

I'm on board with this.

> A central decision on types. If function declarations are provided, they should either all use stdbool for true/false functions, or none of them should. Similarly for functions with number parameters or return types should either all use stdint types, or none of them should.

I'm on board with this as well.

> It'd be lovely if there were a singular style with provided code. Snake case, camel case, etc.

Our guideline for style is the [Linux kernel style](https://01.org/linuxgraphics/gfx-docs/drm/process/coding-style.html). There are a few places where we've intentionally deviated (typedefs for structs, for example), but we're mostly compliant. If there are places where we have a mix of snake case and camel case this is a mistake and not intentional.

@wolf99 @patricksjackson what are your thoughts on 1 and 2?

wolf99:
Function declarations in the header should either be there for all exercises or for none.

I'm not so sure about this. Part of learning C in this manner is deducing the signatures required by the tests. Doing this forces students to think about what and _why_ things are the way they are.

Another part of learning how to C is deciphering debugger output. Understanding this output doesn't get easier in "real life". If students don't learn how to understand it now then they are missing out on an important part of working in C.

> but other languages

I think this illustrates my point a little - C isn't other languages. It doesn't have the slick tools that more modern languages have, it doesn't still have a lot of change in the language. Presenting a learning experience that doesn't highlight the difficulties (and advantages) that this presents would not be representative of what it is like to program in C. 

Recall that in most cases in "real life" a C developer will not even have the tests to guide them, they will essentially be starting with a blank directory. Here they are already provided the directory structure, the required source files (empty) the directory structure, and the makefile.

> A central decision on types. If function declarations are provided, they should either all use stdbool for true/false functions, or none of them should. Similarly for functions with number parameters or return types should either all use stdint types, or none of them should.

I'm OK with this. _Mostly_. Again this is a real life vs exercism-life question. The reality is that in the wild, C programmers will find all kinds of types used, intrinsic, std and custom. Do we use all the same here to make life easier, or do we present students with the variety that C makes possible - are we teaching only a subset of the language if we do not do the latter?

ryanplusplus:
> Function declarations in the header should either be there for all exercises or for none.
> > I'm not so sure about this. Part of learning C in this manner is deducing the signatures required by the tests. Doing this forces students to think about what and why things are the way they are.

I have to disagree with you here. While I think that a C developer _should_ be able to define the type based on its usage, this isn't something that I'd expect a learner or an entry level developer to be able to do. This requires a much greater familiarity with the language and types than is required to, for example, determine whether a given year is a leap year.

Overall I think the crux of the problem is that the setup for some of the problems is harder than many of the actual problems. As much as I'd love for all C developers to be able to do these problems without being given a skeleton, the learning curve has to be right.

computermouth:
Honestly, I understand @wolf99's stance, and largely agree with it after thinking about it. But I do think that maybe some of the order of the exercises should be re-evaluated, and maybe even much much simpler tasks ought to be introduced.

> Overall I think the crux of the problem is that the setup for some of the problems is harder than many of the actual problems. As much as I'd love for all C developers to be able to do these problems without being given a skeleton, the learning curve has to be right.

I can attest to this as well. Bringing in two people who I've handed a couple "C for beginners" books, done a handful of exercises with them before exercism, and just getting started was difficult. But when we got to Resistor Colors, and the initial `make` produces 41 errors, and beginners are intended to figure our they need a newly typed enum and two functions... I start to wonder if the C track is particularly geared towards beginners.

If I think back to when I had started with C, I can easily see myself having a very difficult time with this. One way to bridge the gap, and serve both beginners and intermediate -> experts would be to slowly ramp down on what's provided.

In the case of Resistor Colors, it'd take off quite a few of the errors, and give beginners a point of reference if that enum type existed, but was perhaps missing a few colors. Whereas if Resistor Colors were one of the later exercises, maybe leave out any provided code.

I see that in another thread you're discussing removing some of the string manipulation-heavy exercises for this same overall reason. C is a medieval blood and thunder language, and so I don't think it's the worst idea to put learners through the ringer.

I do think that putting learners through the ringer on the 4th exercise is a little extreme. Just as algorithmic exercises should be ordered on their difficulty, I can't help but think that the same consideration should be put into the difficulty of getting to passing the first test?

wolf99:
I would agree that there should be a ramp up to the harder stuff!
I think we had before a point up to which the headers would be given (IIRC it was up to difficulty level 4 - but I might be imagining that).

We have the discussion ongoing here that you reference @computermouth about the type exercises and the order. There's a catch 44 situation in that we need the exercises added before we can order them 😬 . There is also a larger ongoing effort in exercism to thin about how curricula in general are structured - that is expected to have some sort of result or output in a few months last I heard.

I can certainly say that I often am very unsure about where in the ordering a new exercise implementation should go. It doesn't help that its a long time since I have struggled with the basics of C. For example your mention of enums in the resistor exercise; the logic required is simple enough, it did not occur to me that learners would struggle understanding enums! 🤦‍♂ 

I'd be happy to see PRs on a per-exercise basis that adds similar to your example @computermouth :

>In the case of Resistor Colors, it'd take off quite a few of the errors, and give beginners a point of reference if that enum type existed, but was perhaps missing a few colors.

On the other side, I feel there is still benefit in leaving later exercises without this help, students need to learn it at some point.

I scrolled up again and, as usual, @ryanplusplus  has said it in fewer sentences 😄  (emphasis mine):

>Overall I think the crux of the problem is that the setup for some of the problems is harder than many of the actual problems. As much as I'd love for all C developers to be able to do these problems without being given a skeleton, **the learning curve has to be right**.

wolf99:
IF you'd like to propose a new ordering to the existing exercises, please do! (on a new issue maybe)

computermouth:
> There's a catch 44 situation in that we need the exercises added before we can order them grimacing . There is also a larger ongoing effort in exercism to think about how curricula in general are structured - that is expected to have some sort of result or output in a few months last I heard.

> I'd be happy to see PRs on a per-exercise basis that adds similar to your example @computermouth 

> IF you'd like to propose a new ordering to the existing exercises, please do! (on a new issue maybe)

Sounds great! I'll try and speed through more/most/all of them and take some notes along the way.

In the meantime, sounds like we can merge and close this one? :)

PR #423, opened 2019-10-28
list-ops: add exercise

wolf99:
Towards #390 

Proposing this as a non-core exercise, unlocked by linked-list.


exercises/list-ops/README.md
```diff
@@ -0,0 +1,54 @@
...
+In functional languages list operations like `length`, `map`, and
+`reduce` are very common. Implement a series of basic list operations,
+without using existing functions.
+
+The precise number and names of the operations to be implemented will be 
+track dependent to avoid conflicts with existing names, but the general
+operations you will implement include:
+
+* `append` (*given two lists, add all items in the second list to the end of the first list*);
+* `concatenate` (*given a series of lists, combine all items in all lists into one flattened list*);
```
wolf99: This is not actually checked in this implementation. This is because as C does not allow dynamic typing, the list implementation can only be for one type at a time. Concatenation implies that the list implementation is used for data type, and list type in parallel.

exercises/list-ops/src/example.c
```diff
@@ -0,0 +1,239 @@
...
+   if (!is_list_empty(list)) {
+      list_item_t *next_item = peek_first(list);
+      while (next_item) {
+         ++length;
+         next_item = peek_next(next_item);
+      }
+   }
+
+   return length;
+}
```
wolf99: This could also be done by tracking the length in a wrapping "list" type, but as the struct definition is currently exposed in the header, doing this would place requirement on the implementation

exercises/list-ops/test/test_list_ops.c
```diff
@@ -0,0 +1,391 @@
+#include "vendor/unity.h"
+#include "../src/list_ops.h"
+
+struct list_item **list = NULL;
```
ryanplusplus: As far as I can tell this is unused. Is this left over from a WIP?

exercises/list-ops/test/test_list_ops.c
```diff
@@ -0,0 +1,391 @@
...
+static void check_lists_match(list_item_t ** expected, list_item_t ** actual)
+{
+   // expects inputs to be valid lists
+   while (1) {
+      bool expected_empty = is_list_empty(expected);
+      bool actual_empty = is_list_empty(actual);
+      TEST_ASSERT_EQUAL_MESSAGE(expected_empty, actual_empty,
+                                "List lengths differ");
+      if (expected_empty)
+         break;
```
ryanplusplus: Instead of tucking this inside of a `while`, why not just `return` here?

ryanplusplus:
It seems like the representation here is making things more complicated than they need to be and distracts a bit from the focus of the exercise. Did you consider having a list's internal representation be an array and a size?

wolf99:
@ryanplusplus How would you see this working for the append functionality?

wolf99:
Closing in favour of #423 which uses a much simpler array backed implementation

PR #424, opened 2019-10-30
list-ops: add exercise

wolf99:
This implementation uses a different, simpler, backing structure than in #423.
Still does not implement or test concatenation - for the same reasons.
Have provided the definitions in the header as I think this is the first exercise that uses function pointers.

If this approach is preferred then #423  can be closed.

### DIFF
--- a/exercises/list-ops/test/test_list_ops.c
+++ b/exercises/list-ops/test/test_list_ops.c
@@ -1,0 +1,325 @@
+#include "vendor/unity.h"
+#include "../src/list_ops.h"
+
+void setUp(void)
+{
+}
+
+void tearDown(void)
+{
+}
+
+static void check_lists_match(list_t * expected, list_t * actual)
+{
+   // check inputs to be valid lists
+   TEST_ASSERT_NOT_NULL(expected);
+   TEST_ASSERT_NOT_NULL(actual);
+
+   // check lengths match
+   TEST_ASSERT_EQUAL_MESSAGE(expected->length, actual->length,
+                             "List lengths differ");
+
+   // check values match in non-zero length lists
+   if (expected->length)
+      TEST_ASSERT_EQUAL_MEMORY_ARRAY(expected, actual, sizeof(list_value_t),
+                                     expected->length);
+}
+
+static bool filter_modulo(list_value_t value)
+{
+   return (value % 2 == 1);
+}
+
+static list_value_t map_increment(list_value_t value)
+{
+   return value + 1;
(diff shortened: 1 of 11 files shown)


exercises/list-ops/test/test_list_ops.c
```diff
@@ -0,0 +1,325 @@
...
+   delete_list(list1);
+   delete_list(list2);
+   delete_list(expected);
+   delete_list(actual);
+}
+
+static void test_append_non_empty_lists(void)
+{
+   TEST_IGNORE();
+   list_t *list1 = new_list(2, (list_value_t[]){ 1, 2 });
```
ryanplusplus: We probably don't want to blow people's minds, but it would be cool if you could specify a new list using variadic arguments

ryanplusplus:
This looks great to me and seems like it would be a bit easier for a learner to approach. How do you feel about this vs. #423?

exercises/list-ops/test/test_list_ops.c
wolf99: Coded this up out of curiosity because I realised I have somehow never used variadic arguments in C.
While it results in a nicer interface, it feels ... janky.  The type (un)safety 😱

In the main I think not a good idea though as this exercise already introduces one other new thing: function pointers  🙌 . 
Neither function pointers nor variadic functions are covered by any other exercise yet. While the exercise can work without variadic functions, I think it cannot work without function pointers.

wolf99:
I much prefer this over #423  😄 
Thanks for the suggestion on that one for this!

exercises/list-ops/test/test_list_ops.c
```diff
@@ -0,0 +1,325 @@
...
+}
+
+static void test_reverse_non_empty_list(void)
+{
+   TEST_IGNORE();
+   list_t *list = new_list(4, (list_value_t[]){ 1, 3, 5, 7 });
+   list_t *expected = new_list(4, (list_value_t[]){ 7, 5, 3, 1 });
+
+   reverse_list(list);
+   check_lists_match(expected, list);
```
rootbeersoup: The type signature for `reverse_list` suggests the function should return a new `list_t` but this test expects the same list to be reversed and doesn't actually check the return value. This test should probably be:

```c
list_t *actual = reverse_list(list);
check_lists_match(expected, actual);
```

rootbeersoup: I'll just submit a PR

PR #425, opened 2019-11-16
list-ops: Compare array contents

patricksjackson:
The current solution only compares the length of the arrays, and never compares the contents.

The first two arguments to `TEST_ASSERT_EQUAL_MEMORY_ARRAY` are cast to pointers, and internally Unity doesn't access the memory locations if the pointers are equal, so we don't get the crashes you'd expect from accessing invalid addresses.

Also added a message here, because Unity's error messages are not exactly useful.

Pointed out on slack - https://exercism-team.slack.com/archives/CATCU7LSJ/p1573889084013200

### DIFF
--- a/exercises/list-ops/test/test_list_ops.c
+++ b/exercises/list-ops/test/test_list_ops.c
@@ -1,12 +1,41 @@
 #include "vendor/unity.h"
 #include "../src/list_ops.h"
 
+const int MAX_STRING_LEN = 100;
+
 void setUp(void)
 {
 }
 
 void tearDown(void)
 {
+}
+
+static char *print_list(list_t * list)
+{
+   char *array = (char *)malloc(MAX_STRING_LEN * sizeof(char));
+   char *ptr = array;
+   for (size_t i = 0; i < list->length; i++) {
+      int printed = snprintf(ptr, MAX_STRING_LEN - (ptr - array), " %d ",
+                             (list->values[i]));
+      ptr += printed;
+      if (ptr - array > MAX_STRING_LEN) {
+         break;
+      }
+   }
+   return array;
+}
+
+static char *create_error_message(list_t * expected, list_t * actual)
+{
+   char *message = (char *)malloc(MAX_STRING_LEN * sizeof(char));
+   char *expected_string = print_list(expected);
+   char *actual_string = print_list(actual);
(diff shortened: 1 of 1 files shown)


exercises/list-ops/test/test_list_ops.c
```diff
@@ -9,6 +11,33 @@ void tearDown(void)
 {
 }
 
+static char *print_list(list_t * list)
+{
+   char *array = (char *)malloc(MAX_STRING_LEN * sizeof(char));
```
wolf99: It's best practice to not cast the value returned from `malloc()`

exercises/list-ops/test/test_list_ops.c
```diff
@@ -9,6 +11,33 @@ void tearDown(void)
...
+      if (ptr - array > MAX_STRING_LEN) {
+         break;
+      }
+   }
+   return array;
+}
+
+static char *create_error_message(list_t * expected, list_t * actual)
+{
+   char *message = (char *)malloc(MAX_STRING_LEN * sizeof(char));
```
wolf99: Same goes for the cast here.

exercises/list-ops/test/test_list_ops.c
```diff
@@ -9,6 +11,33 @@ void tearDown(void)
 {
 }
 
+static char *print_list(list_t * list)
+{
+   char *array = (char *)malloc(MAX_STRING_LEN * sizeof(char));
+   char *ptr = array;
+   for (size_t i = 0; i < list->length; i++) {
+      int printed = snprintf(ptr, MAX_STRING_LEN - (ptr - array), " %d ",
```
wolf99: Nice 👍

wolf99:
> internally Unity doesn't access the memory locations if the pointers are equal, so we don't get the crashes you'd expect from accessing invalid addresses.
😳

exercises/list-ops/test/test_list_ops.c
patricksjackson: 👍 

Done

patricksjackson:
To access the Slack there's a link in this which I think will let you join https://github.com/orgs/exercism/teams/track-maintainers/discussions/1

If not you can either sign up to mentor or just ping iHiD or kytrinyx. It's for mentors and maintainers and I could never tell if students could get in.

ryanplusplus:
This is awesome. The error message is a nice touch :+1:

PR #426, opened 2019-11-26
list-ops: Check return value of reverse_list()

rootbeersoup:
The type signature for `reverse_list()` suggests the function should return a new `list_t` but the tests don't actually test the return value; only the same list that is passed to `reverse_list()`.

### DIFF
--- a/exercises/list-ops/test/test_list_ops.c
+++ b/exercises/list-ops/test/test_list_ops.c
@@ -311,11 +311,12 @@ static void test_reverse_empty_list(void)
    list_t *list = new_list(0, NULL);
    list_t *expected = new_list(0, NULL);
 
-   reverse_list(list);
-   check_lists_match(expected, list);
+   list_t *actual = reverse_list(list);
+   check_lists_match(expected, actual);
 
    delete_list(list);
    delete_list(expected);
+   delete_list(actual);
 }
 
 static void test_reverse_non_empty_list(void)
@@ -324,11 +325,12 @@ static void test_reverse_non_empty_list(void)
    list_t *list = new_list(4, (list_value_t[]){ 1, 3, 5, 7 });
    list_t *expected = new_list(4, (list_value_t[]){ 7, 5, 3, 1 });
 
-   reverse_list(list);
-   check_lists_match(expected, list);
+   list_t *actual = reverse_list(list);
+   check_lists_match(expected, actual);
 
    delete_list(list);
    delete_list(expected);
+   delete_list(actual);
 }
 
 int main(void)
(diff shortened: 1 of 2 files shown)


exercises/list-ops/test/test_list_ops.c
```diff
@@ -311,11 +311,12 @@ static void test_reverse_empty_list(void)
...
    list_t *expected = new_list(0, NULL);
 
-   reverse_list(list);
-   check_lists_match(expected, list);
+   list_t *actual = reverse_list(list);
+   check_lists_match(expected, actual);
 
    delete_list(list);
    delete_list(expected);
+   delete_list(actual);
```
rootbeersoup: Not sure what's more appropriate, but removing either `delete_list(list)` or `delete_list(actual)` causes the memcheck to pass with the current given `example.c`. This applies to the non-empty test as well.

rootbeersoup:
Latest commit changes the example to more closely match the style of the other example functions while returning a list of type `list_t`. It also allows the test functions to free all 3 of the allocated lists (`list`, `expected`, `actual`) with no memcheck failures.

rootbeersoup:
No idea what's up with the `indent.sh` script. Each time I run it, it alternates between two styles:

```diff
-if (!list || !filter)
+if(!list || !filter)
```

```diff
-if(!list || !filter)
+if (!list || !filter)
```

ryanplusplus:
@rootbeersoup what OS are you using? On MacOS I'm seeing that `indent.sh` is stable.

exercises/list-ops/test/test_list_ops.c
ryanplusplus: Can you help me understand what you mean here? It looks like all three of these deletions are necessary.

ryanplusplus:
> @rootbeersoup what OS are you using? On MacOS I'm seeing that `indent.sh` is stable.

I tried on Linux as well and I'm seeing what you're seeing. Are you by chance on Ubuntu 19.10 (`indent` 2.2.12)?

ryanplusplus:
This looks good outside of the indent issues.

@wolf99 if you're happy with this we can resolve the indent issues post-merge.

rootbeersoup:
Yeah, macOS. GNU indent 2.2.12

exercises/list-ops/test/test_list_ops.c
rootbeersoup: The memcheck was originally failing with all three deletions with the original `example.c` but passed if you deleted one of them. I changed `remove_list()` in the example file to return a new `list_t` instead of mutating the one passed to the function. Everything passes fine now.

wolf99:
Nice. 
Well noticed on the double free. This was happening because `reverse()` did not return a new list but I had followed the same deletion pattern as all the rest of the tests, which assumed that the UUT _did_ return a new list. Hence deleting `list` was trying to delete the same list as deleting `actual`. And also why this change resolves that issue.

LGTM

PR #434, opened 2020-01-19
WIP: grade-school: add exercise

wolf99:
Really not sure where in the track to add this 🤷‍♂ particularly now v3 is started.
(This is why lint is failing in the CI, I haven't added the exercise to the config.json)
Suggestions welcome.

For #390

### DIFF
--- a/exercises/grade-school/src/example.c
+++ b/exercises/grade-school/src/example.c
@@ -2,75 +2,70 @@
 #include <stdlib.h>
 #include <string.h>
 
+roster_t roster = { 0 };
+
 static int compare_student_names(const void *s1, const void *s2)
 {
-   return strncmp(*(const char **)s1, *(const char **)s2, MAX_NAME_LENGTH);
+   const student_t *student1 = (const student_t *)s1;
+   const student_t *student2 = (const student_t *)s2;
+   return strncmp(student1->name, student2->name, MAX_NAME_LENGTH);
 }
 
 static int compare_student_grades_and_names(const void *s1, const void *s2)
 {
-   const student_t *student1 = *(const student_t **)s1;
-   const student_t *student2 = *(const student_t **)s2;
+   const student_t *student1 = (const student_t *)s1;
+   const student_t *student2 = (const student_t *)s2;
 
    if (student1->grade > student2->grade)
       return 1;
    else if (student1->grade < student2->grade)
       return -1;
    else
-      return compare_student_names(&student1->name, &student2->name);
+      return compare_student_names(student1, student2);
 }
 
-char **roster(size_t count, student_t * students)
+bool add_student(char *name, uint8_t grade)
 {
-   if (!count || !students)
-      return NULL;
+   bool added = false;
 
-   student_t **listing = malloc(sizeof(student_t *) * count);
-
-   if (listing) {
-      for (size_t i = 0; i < count; ++i) {
-         listing[i] = &students[i];
-      }
-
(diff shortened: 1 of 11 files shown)


wolf99:
Also please check I haven't done something silly in the tests, this seems to be my Achilles heel in other recent exercises even though I've double checked! 😄

exercises/grade-school/src/example.h
```diff
@@ -0,0 +1,20 @@
...
+typedef struct {
+   uint8_t grade;
+   char *name;
+} student_t;
+
+char **roster(size_t count, student_t * students);
+
+char **grade(size_t count, student_t * students, uint8_t desired_grade);
+
+void delete_listing(char **listing);
```
ryanplusplus: This interface is pretty confusing to me. Even looking at the tests I'm having a hard time understanding what `roster` and `grade` do. When I've done this exercise in other tracks there's generally a pretty clear flow:

- Create school
- Add students
- Query school
- Cleanup (if necessary)

Do these functions map to this flow?

wolf99: Ah OK, I interpreted the exercise specification JSON pretty literally. The description does convey the flow you mention a bit more.

That said, this was an interesting one to think about what the best way to solve it was in relation to the interface implied by the specification JSON. But on the other hand, it was interesting because it was complex 🤣 
(I realised as well that there is no specification independent way of know the length of the returned arrays 🙈  )

I'll have a go at reworking it with the flow you suggest and see how it looks thanks 👍

exercises/grade-school/src/example.c
```diff
@@ -0,0 +1,71 @@
+#include "grade_school.h"
+#include <stdlib.h>
+#include <string.h>
+
+roster_t roster = { 0 };
```
ryanplusplus:
```suggestion
static roster_t roster = { 0 };
```

exercises/grade-school/test/test_grade_school.c
```diff
@@ -0,0 +1,187 @@
...
+static void test_adding_student_adds_to_roster(void)
+{
+   TEST_IGNORE();               // delete this line to run test
+   roster_t input = {
+      .count = 1,
+      .students = {
+                   (student_t) {
+                                .grade = 2,
+                                .name = "Aimee"}}
+   };
```
ryanplusplus: Wow :)

ryanplusplus:
@wolf99 is this still a work-in-progress?

wolf99:
@ryanplusplus the code is GTG if you're happy with it. 
But the location within the curriculum is uncertain (haven't added exercise to config.json yet).

Perhaps @workingjubilee may have a suggestion given the lead on the recent curriculum re-jig?

ryanplusplus:
This looks good to me once it has a place in `config.json` :+1:

workingjubilee:
I have no particular idea, honestly! I would probably place it nearer the middle-ish.

wolf99:
Thanks both for input. 

Have added in the middle-ish 😉. Although there are strings involved I have proposed it to be core as (IIRC) I think it is the only exercise that uses `qsort()` .

workingjubilee:
"uses `qsort()`" seems worthy of core, yeah.
this falls mostly on the side of "uses strings" in the "is about strings" vs. "uses strings" spectrum, which should keep the worst parts of string fussing to a minimum, but I'll whine about it if it's a problem. :wink:

PR #437, opened 2020-03-19
Demote redundant string exercises from core

workingjubilee:
Removes most string exercises from C track's "core" to reduce redundancy
and to minimize track pains. See issue #390 on GitHub for more detail.
"isogram" and "phone-number" preserved as core string exercises with
simple interfaces.

Demoted for specific complaints:
* palindrome-products
* word-count

While I initially proposed "atbash cipher" and "scrabble score" as possible candidates for the "advanced" string exercise, I realized "phone number" is a more "exemplary" exercise as it asks people to do a genuinely challenging task, and one they might realistically face (as it is "formatting input correctly or erroring"), but in a straightforward way that hopefully is easy to mentor. Also, because I mentioned "isogram" and "word-count", but word-count has specific complaints voiced against it, I favored "isogram" as the "easier" one to keep in core, insofar as that applies.

### DIFF
--- a/config.json
+++ b/config.json
@@ -333,7 +333,7 @@
       "slug": "roman-numerals",
       "uuid": "6cde608f-9819-46cd-884b-dbda39a4fa16",
       "core": false,
-      "unlocked_by": null,
+      "unlocked_by": "grains",
       "difficulty": 3,
       "topics": [
         "arrays",
--- a/config.json
+++ b/config.json
@@ -108,8 +108,8 @@
     {
       "slug": "word-count",
       "uuid": "1e4f4fcf-c32d-459f-8920-15ac56008666",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -332,7 +332,7 @@
     {
       "slug": "roman-numerals",
       "uuid": "6cde608f-9819-46cd-884b-dbda39a4fa16",
-      "core": true,
+      "core": false,
       "unlocked_by": null,
```
ryanplusplus: This needs an unlocking exercise

ryanplusplus:
Thanks for doing this!

ryanplusplus:
Looks good to me

wolf99:
LGTM -  merging :)

PR #438, opened 2020-03-19
Fix fetch-configlet script

workingjubilee:
Shamelessly dupes @coriolinus' fix: https://github.com/exercism/rust/pull/929

"Looks like Github has recently started returning non-uppercased
HTTP headers, at least some of the time. This broke the script,
which looked for a case-sensitive 'Location' header to find the
newest version. We can see this problem in spurious build failures like this."

"This fixes the script such that it no longer cares whether the initial
L is capital or not, and if it breaks again in the future, it will
give a more informative error."

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 LATEST=https://github.com/exercism/configlet/releases/latest
 
 OS=$(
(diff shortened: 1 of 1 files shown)


workingjubilee:
Aha, got a partial fix, now that `fetch-configlet` runs sometimes at all.
Will also review https://github.com/exercism/bash/pull/424 to see what they did.

ryanplusplus:
:+1:

Issue #439, opened 2020-03-24
Make Test Doesn't Run Tests


JimLynchCodes:
Hi, 

When I download the hello world challenge and run `make test` I just get this output:

```
make: `test' is up to date.
```
![Screen Shot 2020-03-24 at 2 58 52 PM](https://user-images.githubusercontent.com/5354163/77466167-3fe07d00-6de0-11ea-8b09-19a56a1911f1.png)

Anyone know what I am doing wrong here? 🤔

patricksjackson:
It should be running the tests even if you haven't made any changes, so that output is never expected. It's likely there's just something not quite right with your setup.

What platform are you trying to run this on?
Also the output of these commands run from the directory you're using would be helpful to rule some things out:

```
make test
make --version
cc --version
ls -R
```

And the contents of your makefile, `cat makefile` (you may have an old version if you started exercism a while ago, as the site will pin the version when you start the exercise)

JimLynchCodes:
Thanks @patricksjackson !

I'm running it on macOS v10.15.3

`make --version`
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0

`cc --version`
Apple clang version 11.0.0 (clang-1100.0.33.17)
Target: x86_64-apple-darwin19.3.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

`ls -R`
README.md      src            tests.out
makefile       test           tests.out.dSYM

./src:
hello_world.c hello_world.h

./test:
test_hello_world.c vendor

./test/vendor:
unity.c           unity.h           unity_internals.h

./tests.out.dSYM:
Contents

./tests.out.dSYM/Contents:
Info.plist Resources

./tests.out.dSYM/Contents/Resources:
DWARF

./tests.out.dSYM/Contents/Resources/DWARF:
tests.out

`cat makefile`
### If you wish to use extra libraries (math.h for instance),
### add their flags here (-lm in our case) in the "LIBS" variable.

LIBS = -lm

###
CFLAGS  = -std=c99
CFLAGS += -g
CFLAGS += -Wall
CFLAGS += -Wextra
CFLAGS += -pedantic
CFLAGS += -Werror
CFLAGS += -Wmissing-declarations
[...]

PR #440, opened 2020-03-26
Add make clean note

wolf99:
Per the suggestion in #439

PR #441, opened 2020-03-27
Make targets phony where appropriate

ryanplusplus:
This fixes the root cause of https://github.com/exercism/c/issues/439.

In `make` if there's a target shouldn't correspond to a file and that you want to run unconditionally, you need to mark it as `.PHONY`. This tells `make` to ignore any files/folders that happen to have the same name as the target (otherwise `make` will use the timestamp of that file to determine whether to run the target).

Where this burned us is that we have a target named `test` and we have a directory `test/`. The target `test` depends upon `tests.out` and so `make test` will only run `./tests.out` if `tests.out` is newer than `test/`. You can see this yourself by doing a `touch test` to trick `make` into thinking that `test` does not need to be run because its inputs are older than `test/`. With this change `make` will ignore `test/` and treat the target `test` as just a fancy name, not a file/folder.

### DIFF
--- a/exercises/acronym/makefile
+++ b/exercises/acronym/makefile
@@ -17,15 +17,18 @@ ASANFLAGS  = -fsanitize=address
 ASANFLAGS += -fno-common
 ASANFLAGS += -fno-omit-frame-pointer
 
+.PHONY: test
 test: tests.out
 	@./tests.out
 
+.PHONY: memcheck
 memcheck: test/*.c src/*.c src/*.h
 	@echo Compiling $@
 	@$(CC) $(ASANFLAGS) $(CFLAGS) src/*.c test/vendor/unity.c test/*.c -o memcheck.out $(LIBS)
 	@./memcheck.out
 	@echo "Memory check passed"
 
+.PHONY: clean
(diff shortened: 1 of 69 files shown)


wolf99:
I had considered PHONY instead of documenting in #440. Main concern was, if PHONY causes a build _every_ time, could it hide other problems and then make them harder to debug?

Documentation instead sidesteps the concern by offering a solution but not applying the same solution to every build problem.

Maybe my concern here is unwarranted though?

ryanplusplus:
I think you're misunderstanding how `.PHONY` works. `.PHONY` only applies to the target, not its dependencies. For `make test` specifically this means that it will always run the tests, not that it will always _build_ the tests. This is the intent of having a non-file target like `test` in the first place, so `.PHONY` here is just going to guarantee that the makefile works the way that you'd think it would work (and the way that it would work in the case where you you're not unfortunate enough to have a file/folder that has the same name as your "symbolic" target).

wolf99:
Change will impact the c-test-runner if makefiles there should remain aligned? @bergjohan

PR #442, opened 2020-04-10
Add stubs for grade-school

wolf99:
There has been [discussion on the mentors slack channel](https://exercism-team.slack.com/archives/CATCU7LSJ/p1586008798002800) about the requirements for grade-school exercise not being clear. To make this more obvious to students I here add the source file scope declaration of the roster.

While doing so I noticed that, if a student has not done any side exercises, then this is the first exercise with structures. To help out, and so that the source builds with the definition added per above, I here add the `struct` definitions to the header file. This has the side benefit that the files will build without too many `struct` related compile errors when initially downloaded.

What do you think, is this needed, or giving too much info up front?


ryanplusplus:
Given that this is the first contact with structs this looks good to me :+1:

PR #443, opened 2020-04-12
hello-world: Add explanatory comments

### LINKED ISSUE #392: Hello World test fails with apparently-correct solution
I'm sorry, this hardly seems like the right place to gripe about being unable to get my Hello World unit test to pass ... but I'm not seeing anyplace else in exercism to do so (didn't there used to be a discussion area?).

I added the line "printf("Hello, World!");" to the function "const char *hello(void)" in "hello_world.c" ... but "make test" still fails with the message "control reaches end of non-void function" error message.  I also added an #include <stdio.h> to the header, since it doesn't seem to be declared anywhere else, and the unit test complained about it.

I am an experienced programmer.  I'm not even completely new to C (although I am very rusty).  As I understand it, my "printf()" line should be a correct solution.  I only halfway understand the rest of the exercise's content (the various test and vendor/unity stuff, etc) ... but honestly, I shouldn't have to.  I can see in the GitHub history that y'all have already simplified the C "Hello, World" exercise at least once or twice ... I hate to think what it was like originally.

My complete solution looks like this ...

`
#include "hello_world.h"
#include <stdio.h>

const char *hello(void)
{
  printf("Hello, World!");
}
`

What am I missing here?  Thanks.

wolf99:
Closes #392 

Have added inline comments inspired by the examples from other tracks linked on the referenced issue.

Have also changed the example function definition to now have a valid return statement, so that the first build does not fail (required adding `#include <stddef.h>` for `NULL`).

Also moved the `#include <stddef.h>` out of the test file as it is not used there

### DIFF
--- a/exercises/hello-world/src/hello_world.h
+++ b/exercises/hello-world/src/hello_world.h
@@ -1,6 +1,14 @@
+// This is called an include guard, which ensures that the header is only
+// included once. You could alternatively use '#pragma once'. See
+// https://en.wikipedia.org/wiki/Include_guard .
 #ifndef HELLO_WORLD_H
 #define HELLO_WORLD_H
 
+// Declare the 'hello()' function, which takes no arguments and returns a
+// 'const char *', i.e. a pointer to a character (in this case the first
+// character in a string). The function itself is defined in the hello_world.c
+// source file. Ths function is called by the test case(s) in the test source
+// file test/test_hello_world.c .
 const char *hello(void);
 
 #endif
--- a/exercises/hello-world/src/hello_world.h
+++ b/exercises/hello-world/src/hello_world.h
@@ -1,6 +1,6 @@
 // This is called an include guard, which ensures that the header is only
 // included once. You could alternatively use '#pragma once'. See
-// https://en.wikipedia.org/wiki/Include_guard .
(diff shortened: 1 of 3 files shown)


exercises/hello-world/src/hello_world.h
```diff
@@ -1,6 +1,14 @@
+// This is called an include guard, which ensures that the header is only
+// included once. You could alternatively use '#pragma once'. See
+// https://en.wikipedia.org/wiki/Include_guard .
```
ryanplusplus, line 3:
```suggestion
// https://en.wikipedia.org/wiki/Include_guard.
```

exercises/hello-world/src/hello_world.h
```diff
@@ -1,6 +1,14 @@
...
 #ifndef HELLO_WORLD_H
 #define HELLO_WORLD_H
 
+// Declare the 'hello()' function, which takes no arguments and returns a
+// 'const char *', i.e. a pointer to a character (in this case the first
+// character in a string). The function itself is defined in the hello_world.c
+// source file. Ths function is called by the test case(s) in the test source
+// file test/test_hello_world.c .
```
ryanplusplus, line 11:
```suggestion
// file test/test_hello_world.c.
```

PR #448, opened 2020-05-01
Added a static assert

elyashiv:
This will fix what caused some pain [here](https://exercism.io/mentor/solutions/d7b4724fa9a74b7f996001a3a51cf626)

### DIFF
--- a/exercises/resistor-color/test/test_resistor_color.c
+++ b/exercises/resistor-color/test/test_resistor_color.c
@@ -4,7 +4,7 @@
 
 #define ARRAY_LENGTH(A) (sizeof(A) / sizeof(A[0]))
 _Static_assert(sizeof(resistor_band_t) == sizeof(int),
-              "We expected the size to be the size of int");
+               "We expected the size to be the size of int");
 
 void setUp(void)
 {
--- a/exercises/resistor-color/test/test_resistor_color.c
+++ b/exercises/resistor-color/test/test_resistor_color.c
@@ -1,7 +1,10 @@
+#include <assert.h>
 #include "vendor/unity.h"
 #include "../src/resistor_color.h"
 
 #define ARRAY_LENGTH(A) (sizeof(A) / sizeof(A[0]))
+static_assert(sizeof(resistor_band_t) == sizeof(int),
+              "We expected the size to be the size of int");
 
 void setUp(void)
 {
--- a/exercises/resistor-color/test/test_resistor_color.c
+++ b/exercises/resistor-color/test/test_resistor_color.c
@@ -3,7 +3,7 @@
 #include "../src/resistor_color.h"
 
 #define ARRAY_LENGTH(A) (sizeof(A) / sizeof(A[0]))
-static_assert(sizeof(resistor_band_t) == sizeof(int),
+_Static_assert(sizeof(resistor_band_t) == sizeof(int),
               "We expected the size to be the size of int");
 
 void setUp(void)
--- a/exercises/resistor-color/test/test_resistor_color.c
+++ b/exercises/resistor-color/test/test_resistor_color.c
@@ -3,8 +3,6 @@
 #include "../src/resistor_color.h"
 
 #define ARRAY_LENGTH(A) (sizeof(A) / sizeof(A[0]))
-_Static_assert(sizeof(resistor_band_t) == sizeof(int),
-               "We expected the size to be the size of int");
 
 void setUp(void)
 {
--- a/exercises/resistor-color/test/test_resistor_color.c
+++ b/exercises/resistor-color/test/test_resistor_color.c
@@ -1,4 +1,3 @@
-#include <assert.h>
 #include "vendor/unity.h"
 #include "../src/resistor_color.h"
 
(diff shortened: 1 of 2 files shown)


exercises/resistor-color/test/test_resistor_color.c
```diff
@@ -1,7 +1,10 @@
+#include <assert.h>
 #include "vendor/unity.h"
 #include "../src/resistor_color.h"
 
 #define ARRAY_LENGTH(A) (sizeof(A) / sizeof(A[0]))
+static_assert(sizeof(resistor_band_t) == sizeof(int),
+              "We expected the size to be the size of int");
```
ryanplusplus, line 7: `static_assert` doesn't exist in C which is why you're failing CI. How about changing this to a check in one of the tests?

wolf99:
Thanks for raising this @elyashiv 

There are probably other exercises similar tests could be added @ryanplusplus , what do you think?

ryanplusplus:
> Thanks for raising this @elyashiv
> 
> There are probably other exercises similar tests could be added @ryanplusplus , what do you think?

After thinking more about this, I wonder if the right answer is to simply provide the `typedef` when we have some assumptions based on the specifics of the type implementation.

exercises/resistor-color/test/test_resistor_color.c
elyashiv: In [newer version of libc](https://code.woboq.org/userspace/glibc/assert/assert.h.html#138) this works (I tested it, I realy did!)

elyashiv:
> After thinking more about this, I wonder if the right answer is to simply provide the typedef when we have some assumptions based on the specifics of the type implementation.

In this exercise we define a enum, so a typedef isn't quite a good option...
The problem is that the testing framework discarded some type checking (by casting to void * and back). We can fix this by wrinting a test function:
```c
static void assert_array(int *expected, int *given, size_t len)
{
  TEST_ASSERT_EQUAL_INT_ARRAY(expected, given, len);
}
```

ryanplusplus:
> > After thinking more about this, I wonder if the right answer is to simply provide the typedef when we have some assumptions based on the specifics of the type implementation.
> 
> In this exercise we define a enum, so a typedef isn't quite a good option...
> The problem is that the testing framework discarded some type checking (by casting to void * and back). We can fix this by wrinting a test function:
> 
> ```c
> static void assert_array(int *expected, int *given, size_t len)
> {
>   TEST_ASSERT_EQUAL_INT_ARRAY(expected, given, len);
> }
> ```

Why wouldn't a `typedef`ed `enum` work here? If we need to know for sure that the size is `int` then `typedef`ing an `enum` would guarantee that.

exercises/resistor-color/test/test_resistor_color.c
patricksjackson: Unfortunately that doesn't help, since we need to support old versions.

elyashiv:
@ryanplusplus did you mean something like this?
```
--- a/exercises/resistor-color/src/resistor_color.h
+++ b/exercises/resistor-color/src/resistor_color.h
@@ -1,4 +1,9 @@
 #ifndef RESISTOR_COLOR_H
 #define RESISTOR_COLOR_H
 
+typedef enum
+{
+
+} resistor_band_t;
+
 #endif

```

ryanplusplus:
@elyashiv yes, that should have size `int`. You can also do something like

```
enum {
   resistor_band_blah,
   ...
};
typedef int resistor_band_t;
```

if you want to be more explicit or use a non-`int` type.

exercises/resistor-color/test/test_resistor_color.c
```diff
@@ -1,7 +1,10 @@
+#include <assert.h>
 #include "vendor/unity.h"
 #include "../src/resistor_color.h"
 
 #define ARRAY_LENGTH(A) (sizeof(A) / sizeof(A[0]))
+_Static_assert(sizeof(resistor_band_t) == sizeof(int),
```
ryanplusplus, line 6: Neat, apparently this is C11 and is actually standard. I've been stuck in C99 land and had no idea.

bergjohan: The macro `static_assert` should also be available in C11, see [https://en.cppreference.com/w/c/error/static_assert](https://en.cppreference.com/w/c/error/static_assert). However, the `makefile` for the C track specifies `-std=c99`. Also, I'm not sure how MSVC handles this, they are pretty far behind on the C11 support I have heard.

elyashiv:
I changed to @ryanplusplus suggestion - added a typedef.

exercises/resistor-color/test/test_resistor_color.c
```diff
@@ -1,3 +1,4 @@
+#include <assert.h>
```
ryanplusplus, line 1:
```suggestion
```

ryanplusplus:
:+1:

exercises/resistor-color/test/test_resistor_color.c
wolf99: Very far behind, C99 support is not complete even. MS have stated there is no intention of catching up either.

PR #449, opened 2020-05-07
circular buffer test case fix

lbblack:
In the util method, read_values_from_buffer, read value is set to zero and fails the assertion tests.

### DIFF
--- a/exercises/circular-buffer/test/test_circular_buffer.c
+++ b/exercises/circular-buffer/test/test_circular_buffer.c
@@ -36,7 +36,7 @@ static void read_values_from_buffer(size_t length, buffer_value_t values[],
                                     circular_buffer_t * buffer)
 {
    for (size_t i = 0; i < length; ++i) {
-      buffer_value_t read_value = 0;
+      buffer_value_t read_value = values[i];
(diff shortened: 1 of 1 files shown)


exercises/circular-buffer/test/test_circular_buffer.c
```diff
@@ -36,7 +36,7 @@ static void read_values_from_buffer(size_t length, buffer_value_t values[],
                                     circular_buffer_t * buffer)
 {
    for (size_t i = 0; i < length; ++i) {
-      buffer_value_t read_value = 0;
+      buffer_value_t read_value = values[i];
```
ryanplusplus, line 39: Since `read_value` should be assigned during the `read` all on the next line, why is this change necessary? I'm having a hard time understanding why we need an initial value at all and why `value[i]` is better than any other value.

wolf99: Agree with @ryanplusplus, this change would cause even an empty `read()` function to pass the tests. I'm not sure what is trying to be achieved by this change.

patricksjackson:
Closing, since this looks to just be a misunderstanding.

PR #450, opened 2020-05-08
Use better typing

elyashiv:


ryanplusplus:
This looks good, but CI is failing because the new signature doesn't match the example. Can you update `example.c` too?

ryanplusplus:
Thanks!

PR #452, opened 2020-05-12
Check unity version

### LINKED ISSUE #447: [CI] Unity version check
Unity now has the version included in `#define` directives in `unity,h`, all current exercises are already updated to this new version

Can we add a check to the CI to confirm that the correct version of Unity is being used for any exercises that are submitted?

Bonus approach 💪 : The expected version is covered in our documentation [here](https://github.com/exercism/c/blob/master/docs/VERSIONS.md). That file could be changed format to specifically cover only the Unity version and then read by the proposed CI check to prevent needing to hard code it and update it in the CI check itself.

I can add this myself, but it will be a while before I have time to get to it. If someone wants to pick it up before me, be my guest! 😄

elyashiv:
closes #447


bin/verify-unity-version
```diff
@@ -0,0 +1,39 @@
...
+set -e
+
+# Move to the root directory of the repo so you can run this script from anywhere
+SCRIPT_DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
+cd "${SCRIPT_DIR}"/..
+
+set +e
+
+
+
```
ryanplusplus, line 16:
```suggestion
```

ryanplusplus:
Looks great, thanks!

wolf99:
Wow, thanks for picking this up @elyashiv 🚀

PR #453, opened 2020-05-22
removed mentions of VS

### LINKED ISSUE #451: Working with visual studio
Currently there is no support for visual studio: we don't provide any way to compile with visual studio, and we don't check it in the CI.

But, it's a good IDE that people use, and it is mentioned in the getting started docs.

I had students that compiled in VS and didn't have warning on high level nor as errors, and this is a bad habit in c. Also, I have a hard time checking for platform specific problems if they use VS, because I need to setup a project.

I suggest one of two:
1. Moving to use a different build system (cmake, scons ...) instead of make, that supports VS
2. Adding vcproj files that students using VS can use.

I'm happy to help with this, but I will want to hear what other people think first.

elyashiv:
This closes #451

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -4,7 +4,7 @@ on your system:
 * A modern C compiler
 * The Make build tool
 
-Apple's Xcode IDEs combine the compiler and
+Apple's Xcode IDE combines the compiler and
 the native build system into a single IDE (integrated development
 environment). Linux environments typically expose the compiler and native
 build system as separate command-line tools.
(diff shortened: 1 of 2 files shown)


docs/INSTALLATION.md
```diff
@@ -4,7 +4,7 @@ on your system:
 * A modern C compiler
 * The Make build tool
 
-Microsoft's Visual Studio IDE and Apple's Xcode IDEs combine the compiler and
+Apple's Xcode IDEs combine the compiler and
```
ryanplusplus, line 7:
```suggestion
Apple's Xcode IDE combines the compiler and
```

ryanplusplus:
Just one typo then I'll be happy. Thanks!

PR #454, opened 2020-05-27
word_count: Clear actual_solution on every test.

Bulagro:
Added another memset for actual_solution on each test.
Not clearing it results on the tests failing when a test has less unique words than the previous one (unless the user clears the struct).


ryanplusplus:
Thanks!

PR #457, opened 2020-07-06
Use stdbool for armstrongnumbers

elyashiv:
I hate returning int instead of bool.


ryanplusplus:
I also hate it, thanks!

PR #458, opened 2020-07-11
Allow anagram tests to be ignored.

xihh87:
Usually TEST_IGNORE is at the top of the test
however the free in the teardown segfaults and thus
does not allow to run ignored tests.

By moving the ignore after each initialization test can run.

As the tests are not up I give them space
so that students can identify them easily.

### DIFF
--- a/exercises/anagram/test/test_anagram.c
+++ b/exercises/anagram/test/test_anagram.c
@@ -59,7 +59,6 @@ static void test_no_matches(void)
 
 static void test_detect_simple_anagram(void)
 {
-   TEST_IGNORE();               // delete this line to run test
    char inputs[][MAX_STR_LEN] = {
       "tan",
       "stand",
@@ -71,6 +70,8 @@ static void test_detect_simple_anagram(void)
    candidates = build_candidates(*inputs, sizeof(inputs) / MAX_STR_LEN);
    enum anagram_status expected[] = { IS_ANAGRAM, NOT_ANAGRAM, NOT_ANAGRAM };
 
+   TEST_IGNORE();               // delete this line to run test
+
    anagrams_for(word, &candidates);
    assert_correct_anagrams(&candidates, expected);
 
@@ -78,7 +79,6 @@ static void test_detect_simple_anagram(void)
 
 static void test_does_not_confuse_different_duplicates(void)
 {
-   TEST_IGNORE();
    char inputs[][MAX_STR_LEN] = {
       "eagle"
    };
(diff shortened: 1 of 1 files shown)


ryanplusplus:
The downside to this is that if there's a problem generating the solution for any of the ignored tests the code still runs (and possibly crashes).

How about just making `teardown()` safe?

patricksjackson:
It's probably better to fix this by adding `candidates.candidates = null" just after the free statement. That way we keep the same expected format of the other tests.

Even better would be to get rid of the global `candidates` variable and just do the free in each test, as that's a lot more readable.

xihh87:
It seems there are some opposing interests here:

-   Maintaining consistency throughout the tests.
-   Providing tests that readily work when you have an empty skeleton.
-   Keeping the testcase maintainable.

> The downside to this is that if there's a problem generating the solution for any of the ignored tests the code still runs (and possibly crashes).

@ryanplusplus This is exactly the problem I want to solve:

Right now, you can't simply run all tests **when you have an empty skeleton** for functions on `src/anagram.c` and `src/anagram.h` because **the tests already crash** trying to free a NULL pointer.

This change fixes that problem but does not maintain the testcase consistent with the rest.

I think it is important that the tests all run and can be readily ignored when your skeleton is ready because the objective it's that students can focus on working on the solution rather than fixing the tests. What do you think about that?

> How about just making teardown() safe?

«Just making teardown() safe» could be done by:

-    Deleting all the code there and adding a free statement on each test.

    I'd rather not do that because this adds a maintenance burden on the structure of each test.

-    Checking that the pointer is not null.

    This seems like it would be a simple fix.

    However I'd like that the simple skeleton can be tested.

> It's probably better to fix this by adding candidates.candidates = null just after the free statement. That way we keep the same expected format of the other tests.

This would not fix the problem because the problem is trying to free a NULL pointer.
If the free is removed, there would be memory leaks.

> Even better would be to get rid of the global candidates variable and just do the free in each test, as that's a lot more readable.

I think that would add a maintainability burden because each test has to do it'.

ryanplusplus:
> Right now, you can't simply run all tests when you have an empty skeleton for functions on src/anagram.c and src/anagram.h because the tests already crash trying to free a NULL pointer.

The scenario I'm talking about is different. If you have some user code that works for some tests but crashes for others due to the solution being incomplete, the proposed ignore solution will not work.

I understand the desire to be able to compile an empty skeleton and not have it crash and agree that it's a bug, but you can solve this problem without introducing the problem that I identified. Just make sure that `teardown` is safe by making sure that the pointer is made `NULL` after `free`.

>> It's probably better to fix this by adding candidates.candidates = null just after the free statement. That way we keep the same expected format of the other tests.

> This would not fix the problem because the problem is trying to free a NULL pointer.
If the free is removed, there would be memory leaks.

This _does_ fix the problem because calling `free` on `NULL` is guaranteed to be safe. @patricksjackson is not proposing to remove the `free`, so there will not be a memory leak.

ryanplusplus:
@xihh87 this allows the solution to be compiled and tested without error:
https://github.com/exercism/c/pull/462

PR #460, opened 2020-07-11
anagram: Add test for non anagrams.

xihh87:
A solution which sums each char on the word and compares the sum
will pass all tests because no words tested have a char combination
that sum the same as another word.

*   Provide an adversarial example for that situation.

*   Test is ignored after creating the candidates
    so it will not cause segfault when ignored.

### DIFF
--- a/exercises/anagram/test/test_anagram.c
+++ b/exercises/anagram/test/test_anagram.c
@@ -76,6 +76,22 @@ static void test_detect_simple_anagram(void)
 
 }
 
+static void test_detect_adversarial_sum_anagram(void)
+{
+   TEST_IGNORE();
+
+   char inputs[][MAX_STR_LEN] = {
+      "abcde"
+   };
+
+   char word[] = { "aacdf" };
+   candidates = build_candidates(*inputs, sizeof(inputs) / MAX_STR_LEN);
+   enum anagram_status expected[] = { NOT_ANAGRAM };
+
+   anagrams_for(word, &candidates);
+   assert_correct_anagrams(&candidates, expected);
+}
+
 static void test_does_not_confuse_different_duplicates(void)
 {
    TEST_IGNORE();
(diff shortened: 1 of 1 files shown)


exercises/anagram/test/test_anagram.c
```diff
@@ -76,6 +76,22 @@ static void test_detect_simple_anagram(void)
...
+{
+   char inputs[][MAX_STR_LEN] = {
+      "abcde"
+   };
+
+   char word[] = { "aacdf" };
+   candidates = build_candidates(*inputs, sizeof(inputs) / MAX_STR_LEN);
+   enum anagram_status expected[] = { NOT_ANAGRAM };
+
+   TEST_IGNORE();
```
ryanplusplus, line 89: Per my comment in #458, can you put this into the "normal" spot? Looks good otherwise.

xihh87: Done!

ryanplusplus:
Thanks!

wolf99:
This might indicate an issue with the problem spec.
Or perhaps the problem spec has been updated to cover it since this exercise was last aligned with the spec.
If you're interested to check it out @xihh87 , the spec is at: https://github.com/exercism/problem-specifications/blob/master/exercises/anagram/canonical-data.json

ryanplusplus:
@wolf99 I suspect this is kind of track specific. It may only make sense in languages where characters are integers are so closely related. Additionally, aren't all of the problem specifications frozen until v3?

wolf99:
Last I heard (some time ago) the problem specs is accepting bug fixes, which this may be considered.
I think it is higher probability though that the spec has been updated since the exercise was implemented and this case may be covered by now. Possibly the following case may cover it when translated into the C domain - depending on implementors choice: https://github.com/exercism/problem-specifications/blob/3f1d2a26a481f3f126a7f489c04478663a2695af/exercises/anagram/canonical-data.json#L67

PR #461, opened 2020-07-11
rational-nubers: Test a real rational zero.

xihh87:
All tests where zero is expected have denominator 1.

The test provides an invalid value and maybe should fail
because a rational number with 0 denominator is undefined.


ryanplusplus:
Good catch, looks like this was a miss when this exercise was implemented:
https://github.com/exercism/problem-specifications/blob/master/exercises/rational-numbers/canonical-data.json#L326-L332

wolf99:
Well caught!

PR #462, opened 2020-07-11
Ensure that empty anagram solution does not double-free

ryanplusplus:
See https://github.com/exercism/c/pull/458


ryanplusplus:
@wolf99 the header uses `size_t` which comes from `stddef.h` so it's needed in the header.

PR #465, opened 2020-07-23
list-ops: Make list test independent of user.

xihh87:
-  A defective new_list function can pass most tests
   provided it always returns the same output.

-  This fix tests for the expected contents
   instead of relying on user provided function.


ryanplusplus:
Thanks :+1:

PR #469, opened 2020-07-31
Some fixes to list ops

elyashiv:
This addresses some of the issues @siebenschlaefer raised in https://exercism.io/mentor/solutions/0ff83e4bbbfc432d83e04a3410264027

### DIFF
--- a/exercises/list-ops/src/example.h
+++ b/exercises/list-ops/src/example.h
@@ -1,7 +1,7 @@
-#ifndef LINKED_LIST_H
-#define LINKED_LIST_H
+#ifndef LIST_OPS_H
+#define LIST_OPS_H
 
-#include <stdlib.h>
+#include <stddef.h>
 #include <stdbool.h>
 
 typedef int list_value_t;
@@ -14,11 +14,11 @@ typedef struct {
 // constructs a new list
 list_t *new_list(size_t length, list_value_t values[]);
 
-// append entries to a list and return the new list
-list_t *append_list(list_t * list1, list_t * list2);
+// concat two lists into one new list
+list_t *concat_list(list_t * list1, list_t * list2);
 
 // filter list returning only values that satisfy the filter function
-list_t *filter_list(list_t * list, bool(*filter) (list_value_t value));
+list_t *filter_list(list_t * list, bool (*filter)(list_value_t value));
 
 // returns the length of the list
 size_t length_list(list_t * list);
@@ -41,7 +41,7 @@ list_value_t foldr_list(list_t * list, list_value_t initial,
 list_t *reverse_list(list_t * list);
 
 // destroy the entire list
-// list will be a dangling pointer after calling this method on it
+// list will be a dangling pointer after calling this function on it
 void delete_list(list_t * list);
 
 #endif
--- a/exercises/list-ops/src/example.h
+++ b/exercises/list-ops/src/example.h
@@ -15,30 +15,30 @@ typedef struct {
 list_t *new_list(size_t length, list_value_t values[]);
 
 // concat two lists into one new list
-list_t *concat_list(list_t * list1, list_t * list2);
+list_t *concat_list(const list_t * list1, const list_t * list2);
 
 // filter list returning only values that satisfy the filter function
-list_t *filter_list(list_t * list, bool (*filter)(list_value_t value));
+list_t *filter_list(const list_t * list, bool (*filter)(list_value_t value));
 
 // returns the length of the list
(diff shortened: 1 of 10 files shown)


exercises/list-ops/src/example.h
```diff
@@ -14,34 +14,34 @@ typedef struct {
...
                         list_value_t(*foldr) (list_value_t value,
                                               list_value_t initial));
 
 // reverse the elements of the list
-list_t *reverse_list(list_t * list);
+list_t *reverse_list(const list_t * list);
 
 // destroy the entire list
-// list will be a dangling pointer after calling this method on it
+// list will be a dangling pointer after calling this function on it
```
ryanplusplus, line 44: I don't agree with this one. There's a difference between functions and methods, it's not just a matter of all functions in C being called "function". Methods are a specific type of function that take the object it acts upon as the first (or hidden) argument.

exercises/list-ops/src/list_ops.h
```diff
@@ -14,34 +14,34 @@ typedef struct {
...
                         list_value_t(*foldr) (list_value_t value,
                                               list_value_t initial));
 
 // reverse the elements of the list
-list_t *reverse_list(list_t * list);
+list_t *reverse_list(const list_t * list);
 
 // destroy the entire list
-// list will be a dangling pointer after calling this method on it
+// list will be a dangling pointer after calling this function on it
```
ryanplusplus, line 44:
```suggestion
// list will be a dangling pointer after calling this method on it
```

exercises/list-ops/src/example.h
ryanplusplus:
```suggestion
// list will be a dangling pointer after calling this method on it
```

exercises/list-ops/src/example.h
```diff
@@ -14,34 +14,34 @@ typedef struct {
...
-list_t *map_list(list_t * list, list_value_t(*map) (list_value_t value));
+list_t *map_list(const list_t * list, list_value_t(*map) (list_value_t value));
 
 // folds (reduces) the given list from the left with a function
-list_value_t foldl_list(list_t * list, list_value_t initial,
+list_value_t foldl_list(const list_t * list, list_value_t initial,
                         list_value_t(*foldl) (list_value_t value,
                                               list_value_t initial));
 
 // folds (reduces) the given list from the right with a function
```
ryanplusplus, line 35:
```suggestion
// folds (reduces) the given list from the right with a method
```

exercises/list-ops/src/example.h
```diff
@@ -14,34 +14,34 @@ typedef struct {
...
 // returns the length of the list
-size_t length_list(list_t * list);
+size_t length_list(const list_t * list);
 
 // return a list of elements whose values equal the list value transformed by
 // the mapping function
-list_t *map_list(list_t * list, list_value_t(*map) (list_value_t value));
+list_t *map_list(const list_t * list, list_value_t(*map) (list_value_t value));
 
 // folds (reduces) the given list from the left with a function
```
ryanplusplus, line 30:
```suggestion
// folds (reduces) the given list from the left with a method
```

ryanplusplus:
@patricksjackson did you find a solution for the change in behavior on newer versions of `indent`? I seem to remember you having done some investigation into the matter.

exercises/list-ops/src/example.h
siebenschlaefer: I complained about "method" in a comment on my solution of this exercise because "method" is not mentioned in the C standard (at least not in the sense of "function" or "function operating on some object"). There was a discussion about that on v3 (I think) where they agreed that they didn't want to introduce technical terms from other languages.

But that's not a hill I want to die on.

ryanplusplus: Method isn't a term from another language, though. It's a general term referring to functions defined for or on objects which this definitely is. To me saying we want to avoid it is like saying this shouldn't be done in an object oriented way since C doesn't have built in support for object orientation.

siebenschlaefer: Does that also apply to the `foldl()` and `foldr()` functions/methods? Is every function that takes an argument a method becaus it acts upon it?  
If you insist on "method" that's OK with me. I'm just trying to understand the argument.

ryanplusplus: In the case of the fold methods, I think the word "function" refers to the function passed as an argument to the method so the terminology is accurate.

ryanplusplus:
Closing as stale. Feel free to reopen when changes are made.

PR #473, opened 2020-09-03
add test for NULL pointer to match example code

fredrb:
Example solution checks if `buffer` is passed and returns `EINVAL` in case it wasn't. `EINVAL` isn't checked in any test case.

[example.c](https://github.com/exercism/c/blob/master/exercises/circular-buffer/src/example.c):
```c
static int16_t optional_overwrite(circular_buffer_t * buffer,
                                  buffer_value_t value, bool writeover)
{
   if (!buffer) {
      errno = EINVAL;
      return EXIT_FAILURE;
   }
// ...
}
```

### DIFF
--- a/exercises/circular-buffer/test/test_circular_buffer.c
+++ b/exercises/circular-buffer/test/test_circular_buffer.c
@@ -65,9 +65,18 @@ static void test_reading_empty_buffer_fails(void)
    delete_buffer(buffer);
 }
 
-static void test_can_read_item_just_written(void)
+static void test_reading_invalid_buffer_fails(void)
 {
    TEST_IGNORE();               // delete this line to run test
+   buffer_value_t read_value = 0;
+   int16_t status = read(NULL, &read_value);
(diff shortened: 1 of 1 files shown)


wolf99:
Hi @fredrb 
The tests are usually aligned with the problem specifications: https://github.com/exercism/problem-specifications/blob/master/exercises/circular-buffer/canonical-data.json

Normally I would recommend adding to the problem specification first and then propogating here;, however the probolem-specs repo is only accepting fixes (https://github.com/exercism/problem-specifications/issues/1560)

We have added new exercises based on proposed problem specs before to get around this, but this would be a fork of an existing exercise with no way of back-porting the change.... I'm not sure if that is a good idea.

I think the EINVAL check in the example is only there for completeness and good practice in checking if the pointer argument is valid.

fredrb:
Hey @wolf99 thanks for the reply. Should we just close this PR then?

PR #474, opened 2020-09-09
Improve consistency in square root naming

ryanplusplus:
Functions in other exercises use `snake_case` but square root used `camelCase`. This switches `squareRoot` to `square_root` to be consistent with other exercises.


krishna-pixel:
add the new function

PR #478, opened 2020-09-30
Update LEARNING.md

Prathu121:


kotp:
In order to be registered as invalid in terms of Hacktoberfest, the label should be `invalid`.

kotp:
At least, and that would be nice, I think, I am unaware if `hacktoberfest spam` works.  It would be nice if it does.

ryanplusplus:
> In order to be registered as invalid in terms of Hacktoberfest, the label should be `invalid`.

Their docs say any label with invalid or spam in the name will work. This may be new for this year's Hacktoberfest.

kotp:
I like the specific label myself, so that would be a nice way to do so.  I could think of situations where something could be invalid, to apply, that you would want to mark as such, but might be valuable in the learning, so not invalid for credit for Hacktoberfest.

PR #492, opened 2020-10-14
acronym: Update to most recent spec

wolf99:
- Add missing test cases
  - Update implementation to handle leading underscores for one of these new cases
- Move C-specific extra test cases (null and empty string) to top
- Remove outdated test cases
- Refactor helper function to indicate it is a check rather than a test (in line with practice in other exercises)

### DIFF
--- a/exercises/acronym/src/example.c
+++ b/exercises/acronym/src/example.c
@@ -1,13 +1,14 @@
 #include "acronym.h"
 #include <ctype.h>
 #include <stdlib.h>
+#include <stdbool.h>
 
-static int is_first_letter_of_word(int character, int preceding_character)
+static bool is_first_letter_of_word(int character, int preceding_character)
 {
    return isalpha(character)
        && ((preceding_character == ' ')
            || (preceding_character == '-')
-           || (preceding_character == '_')) ? 1 : 0;
+           || (preceding_character == '_'));
 }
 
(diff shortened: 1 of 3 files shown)


wolf99:
Travis seems to be intermittently failing with:

```
The command "bin/fetch-configlet" exited with 2.
0.00s$ bin/configlet lint .
/home/travis/.travis/functions: line 109: bin/configlet: No such file or directory
The command "bin/configlet lint ." exited with 127.
```

Any clues why?

iHiD:
(cc @ErikSchierboom - I'm seeing this travis issue in lots of places)

exercises/acronym/src/example.c
```diff
@@ -6,7 +6,8 @@ static int is_first_letter_of_word(int character, int preceding_character)
 {
    return isalpha(character)
        && ((preceding_character == ' ')
-           || (preceding_character == '-')) ? 1 : 0;
+           || (preceding_character == '-')
+           || (preceding_character == '_')) ? 1 : 0;
```
ryanplusplus, line 10: Is the `? 1 : 0` actually necessary?

ryanplusplus: I'm fairly certain the answer is no. We should probably tweak this to return a `bool` for clarity.

PR #531, opened 2020-10-16
Migrate from Travis to GitHub Actions

### LINKED ISSUE #494: Moving from Travis to GitHub Actions
Hello 🙂 

Over the last few months we've been transferring all our CI from Travis to [GitHub Actions (GHA)](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions). We've found that GHA are easier to work with, more reliable, and much much faster. 

Based on our success with GHA and increasing intermittent failures on Travis, we have now decided to try and remove Travis from Exercism's org altogether and [shift everything to GHA](https://github.com/exercism/exercism/issues/5396). This issue acts as a call to action if your track is still using Travis. 

For most CI checks this should be a transposing from Travis' syntax to [GHA syntax](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions), and hopefully quite straightforward (see [this PR for an example](https://github.com/exercism/csharp/pull/1338/files)).  However, if you do encounter any issues doing this, please ask on Slack where lots of us now have experience with GHA, or post a comment here and I'll tag relevant people. This would also make a good Hacktoberfest issue for anyone interested in making their first contribution 🙂 

If you've already switched this track to GHA, please feel free to close this issue and ignore it.

Thanks!

jrr:
closes #494 

Here's an example build from my fork: https://github.com/jrr/c/runs/1262250896?check_suite_focus=true


.github/workflows/build.yml
```diff
@@ -0,0 +1,28 @@
+name: test
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
ryanplusplus, line 7: Given that we're sensitive to the version of indent, I'd rather pin a specific version of Ubuntu.

.github/workflows/build.yml
```diff
@@ -0,0 +1,28 @@
...
+      run: |
+        ./bin/fetch-configlet
+        ./bin/configlet lint .
+        ./bin/verify-indent
+        ./bin/check-unitybegin
+        ./bin/verify-unity-version
+        shellcheck bin/fetch-configlet bin/verify-indent bin/check-unitybegin bin/run-tests indent.sh
+
+    - name: test
+      run: ./bin/run-tests
```
ryanplusplus, line 28:
```suggestion
      run: ./bin/run-tests

```

jrr: Interestingly, the GitHub UI won't let me accept the suggestion, but I just pushed a commit with the newline.

.github/workflows/build.yml
jrr: Indeed, `ubuntu-latest` still resolves to 18.04, and 20.04 doesn't work! There's a problem for the future ;)

ryanplusplus (approved):

PR #544, opened 2020-12-22
Add repo code formatting in CI

### LINKED ISSUE #472: Fix indent.sh
Different versions of `indent` produce slightly different output. One option is to update `indent.sh` to use a version of `indent` that is included in this repo (64-bit Linux and 64-bit MacOS builds are probably sufficient) so that we have tighter control over the version being used.

wolf99:
Fixes #472 

cc @elyashiv 

There's an example commit where the use of this is demonstrated, on my fork of this repo, here: https://github.com/wolf99/c/pull/2

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -1,0 +1,47 @@
+name: "Format C code"
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  format:
+    name: "Format C code"
+    runs-on: ubuntu-18.04
+    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/format')
+    steps:
+      - name: "Checkout code"
+        run: |
+          PR_DATA="/tmp/pr.json"
+          jq -r ".issue.pull_request.url" "$GITHUB_EVENT_PATH" | \
+            xargs curl --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -o "$PR_DATA" --url
+          HEAD_REF=$(jq -r ".head.ref" "$PR_DATA")
+          HEAD_REPO=$(jq -r '.head.repo.clone_url | sub("https://"; "https://x-access-token:${{ secrets.GITHUB_TOKEN }}@")' "$PR_DATA")
+
+          git clone $HEAD_REPO .
+          git checkout -b "$HEAD_REF" "origin/$HEAD_REF"
+
+      - name: install indent
+        run: |
+          sudo apt-get -qq update
+          sudo apt-get install -y indent
+
+      - name: "Format C code"
+        run: ./indent.sh
+
+      - name: "Commit formatted C code"
+        run: |
+          # Check if there is nothing to commit (i.e. no formatting changes made)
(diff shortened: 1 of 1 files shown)


.github/workflows/format-code.yml
```diff
@@ -0,0 +1,47 @@
+name: "Format C code"
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  format:
+    name: "Format C code"
+    runs-on: ubuntu-18.04
```
wolf99, line 10: Should we be able to change to `ubuntu-latest` now? 
Not sure if there is a concern for compatibility with local, or if we are able to "ignore" local once this is in.
If  we can change it, it should probably change in the `build.yml` also

.github/workflows/format-code.yml
```diff
@@ -0,0 +1,47 @@
...
+    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/format')
+    steps:
+      - name: "Checkout code"
+        run: |
+          PR_DATA="/tmp/pr.json"
```
wolf99, line 15: @patricksjackson I've lifted this from the C++ workflow you linked on #472. Do you know why the checkout action isn't used instead of all this?
I suspect that maybe checkout action does not get PR code, but just the current master, but I need to learn about the action.

wolf99: Had a look at the action (https://github.com/marketplace/actions/checkout) and the `issue_comment` event spec (https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#issue_comment).
Seems my guess above is essentially the case.
Looks like the action may be able to do the same, if I add some additional futzing about, but the difference would then be negligible I think.

ryanplusplus:
Does/can this trigger another build?

.github/workflows/format-code.yml
patricksjackson: Especially for the formatting code here, I think we should keep this pinned to a specific version so that we aren't surprised by changes to indent. It's a small thing but may avoid confusion in the future.

ubuntu-latest is still ubuntu-18.04 AFAICT

.github/workflows/format-code.yml
patricksjackson: Makes sense. If there isn't an official way to do it with the checkout action we should probably just stick with this as it already works.

patricksjackson (approved):

ryanplusplus:
> Does/can this trigger another build?

I guess as long as the formatting is "stable" this wouldn't be an issue

wolf99:
> Does/can this trigger another build?

I though on my example PR that it did, but I see on the following similar PR on problem specs repo that there is specific code nd output to deal with that. Intend to investigate further.

https://github.com/exercism/problem-specifications/pull/1754

wolf99:
Turns out, yes the commit created by CI triggers the CI, but not on PRs from forks.

wolf99:
Hrm, the above *should* be the case... but it is not. I tested on #545 and had to push a second commit to trigger the CI checks.
Will investigate further.

PR #550, opened 2020-12-29
binary-search: align with specification per #499

### LINKED ISSUE #499: binary-search: update to most recent canonical specification
Recently a `.meta/tests.toml` file was introduced to each exercise on this track.
The purpose being to show how an exercise aligns (or does not align)  to the exercise's `canonical-data.json` (in the [problem-specifications](https://github.com/exercism/problem-specifications/) repo), by comparing the UUID assigned to each test case.

See the [canonical-data-syncer](https://github.com/exercism/canonical-data-syncer) for how to check the exercises for this.

This has shown that the binary-search exercise has missing or out of date tests.

The request for this issue is to assign the issue to yourself and then (having read our [CONTIBUTING.md](https://github.com/exercism/c/blob/master/docs/CONTRIBUTING.md)) update the exercise such that:

- Existing test names are correct
- Existing tests match the requirements of the related `canonical-data.json`
- Missing tests are added
  - If needed, the example implementation is adjusted to fulfill the added test cases
- Update the related `tests.toml` file to add the UUIDs that correspond to the updated tests

See the following past PR for an example #493

wolf99:
fixes #499

### DIFF
--- a/exercises/binary-search/test/test_binary_search.c
+++ b/exercises/binary-search/test/test_binary_search.c
@@ -97,14 +97,6 @@ static void test_nothing_is_found_when_the_left_and_right_bounds_cross(void)
    TEST_ASSERT(NULL == binary_search(0, arr, length))
 }
 
-static void test_nothing_is_found_if_array_is_null(void)
-{
-   TEST_IGNORE();
-   int *arr = NULL;
-   size_t length = 0;
-   TEST_ASSERT(NULL == binary_search(1, arr, length));
-}
-
 int main(void)
 {
    UnityBegin("test/test_binary_search.c");
@@ -120,7 +112,6 @@ int main(void)
    RUN_TEST(test_a_value_larger_than_the_arrays_largest_value_is_not_found);
    RUN_TEST(test_nothing_is_found_in_an_empty_array);
    RUN_TEST(test_nothing_is_found_when_the_left_and_right_bounds_cross);
-   RUN_TEST(test_nothing_is_found_if_array_is_null);
 
    return UnityEnd();
 }
--- a/exercises/binary-search/test/test_binary_search.c
+++ b/exercises/binary-search/test/test_binary_search.c
@@ -9,14 +9,14 @@ void tearDown(void)
 {
 }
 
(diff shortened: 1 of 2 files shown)


exercises/binary-search/test/test_binary_search.c
```diff
@@ -9,89 +9,98 @@ void tearDown(void)
...
+static void test_nothing_is_found_when_the_left_and_right_bounds_cross(void)
 {
    TEST_IGNORE();
-   int arr[] = { 1 };
+   int arr[] = { 1, 2 };
+   size_t length = sizeof(arr) / sizeof(arr[0]);
+   TEST_ASSERT(NULL == binary_search(0, arr, length))
+}
+
+static void test_nothing_is_found_if_array_is_null(void)
```
wolf99, line 100: This test is additional to the spec.
I am in two minds if to keep it in or not.
On one hand it seems a logical test for idiomatic C based languages, on the other hand it diverges from the spec.

This question comes about because of the interpretation of the semantics of the spec for the (specified) test `test_nothing_is_found_in_an_empty_array()`.
As implemented here, the aforementioned test checks what it says, an empty array. That is, an array that exists, but that has nothing in it.
Whereas this test in question checks the case that is more idiomatic to C specifically, i.e. the array does not exist.

There are three options
1. It's not important to check every C idiomatic thing on every exercise, so remove this test and stick to a strict interpretation of the specification
2. Remove this test, but change `test_nothing_is_found_in_an_empty_array()` to actually check for no array, rather than an empty array (as this would be a more likely implementation for "empty array" in C), resulting in tests that follow the specification, but with a slightly looser interpretation.
3. Keep both tests as they are - ( i.e. the strict specification semantics + our extra test )

ryanplusplus: I lean toward 1). For something like Exercism (which definitely isn't production code), having a ton of checks on every exercise just gets in the way of the important parts. I'm happy with including idiomatic checks in some exercises and letting the "meat" of the exercises take center stage in the general case.

ryanplusplus (approved):

PR #556, opened 2020-12-29
hamming: align with specification per #504

### LINKED ISSUE #504: hamming: update to most recent canonical specification
Recently a `.meta/tests.toml` file was introduced to each exercise on this track.
The purpose being to show how an exercise aligns (or does not align)  to the exercise's `canonical-data.json` (in the [problem-specifications](https://github.com/exercism/problem-specifications/) repo), by comparing the UUID assigned to each test case.

See the [canonical-data-syncer](https://github.com/exercism/canonical-data-syncer) for how to check the exercises for this.

This has shown that the hamming exercise has missing or out of date tests.

The request for this issue is to assign the issue to yourself and then (having read our [CONTIBUTING.md](https://github.com/exercism/c/blob/master/docs/CONTRIBUTING.md)) update the exercise such that:

- Existing test names are correct
- Existing tests match the requirements of the related `canonical-data.json`
- Missing tests are added
  - If needed, the example implementation is adjusted to fulfill the added test cases
- Update the related `tests.toml` file to add the UUIDs that correspond to the updated tests

See the following past PR for an example #493

wolf99:
Fixes #504 

Again here the question of semantics in the spec.
Does empty mean a zero-length string (i.e. ""), or NULL?

### DIFF
--- a/exercises/hamming/test/test_hamming.c
+++ b/exercises/hamming/test/test_hamming.c
@@ -50,18 +50,6 @@ static void test_disallow_second_strand_when_longer(void)
    TEST_ASSERT_EQUAL(-1, compute("ATA", "AGTG"));
 }
 
-static void test_disallow_left_empty_strand(void)
-{
-   TEST_IGNORE();
-   TEST_ASSERT_EQUAL(-1, compute(NULL, "G"));
-}
-
-static void test_disallow_right_empty_strand(void)
-{
-   TEST_IGNORE();
-   TEST_ASSERT_EQUAL(-1, compute("G", NULL));
-}
-
 int main(void)
 {
    UnityBegin("test/test_hamming.c");
@@ -73,8 +61,6 @@ int main(void)
    RUN_TEST(test_long_different_strands);
    RUN_TEST(test_disallow_first_strand_when_longer);
    RUN_TEST(test_disallow_second_strand_when_longer);
-   RUN_TEST(test_disallow_left_empty_strand);
-   RUN_TEST(test_disallow_right_empty_strand);
 
    return UnityEnd();
 }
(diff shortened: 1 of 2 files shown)


ryanplusplus:
> Again here the question of semantics in the spec.
> Does empty mean a zero-length string (i.e. ""), or NULL?

IMO, empty means `""` instead of `NULL`. Assuming that `NULL` can be passed as the "zero"/empty/whatever value is a common problem I see. I'd rather not muddy the waters between empty and missing.

exercises/hamming/test/test_hamming.c
```diff
@@ -14,95 +14,67 @@ static void test_empty_strands(void)
...
+   TEST_ASSERT_EQUAL(-1, compute(NULL, "G"));
 }
 
-static void test_hamming_distance_in_very_long_strand(void)
+static void test_disallow_right_empty_strand(void)
 {
    TEST_IGNORE();
-   TEST_ASSERT_EQUAL(9, compute("GGACGGATTCTG", "AGGACGGATTCT"));
+   TEST_ASSERT_EQUAL(-1, compute("G", NULL));
 }
```
ryanplusplus, line 63: Well, now I'm having second thoughts about my earlier comment regarding `""` and `NULL`. I don't think that the spec makes much sense here. The name and the check don't match. An empty strand (`""`) shouldn't be an error , but a missing strand (`NULL`) should be. Further confusing things is that because the strand lengths don't match, it would be an error regardless of whether it was empty or missing so it's confusing to say that an _empty_ strand should be treated as an error.

I think these tests would make far more sense if named `test_disallow_<which>_strand_missing`.

wolf99: Agreed, the logic means that these two tests don't make sense.
I've removed them and set the related UUIDs to false in the TOML file.
Have also opened an issue on the problem specs repo about it (https://github.com/exercism/problem-specifications/issues/1761).

ryanplusplus: Thanks! 👍

ryanplusplus (approved):

PR #618, opened 2021-02-10
Update practice exercises to V3 structure

wolf99:
Updates the structure of the practice exercises to match the spec (https://github.com/exercism/docs/blob/main/anatomy/tracks/practice-exercises.md).

This is a large commit so I have not squashed it, hopefully this makes it a little easier to review as the first commit is only changes to the scripts, whereas the second commit is the bulk of the change but should be the same for every exercise.

---

If interested, I created the following `restructure.sh` bash script for the changes in the second commit, which I ran from the exercises/practice directory using xargs to provide each practice directory name `$ exa -D --color=never | xargs -I{} ./restructure.sh {}

```bash
#!/usr/bin/env bash

# abort on error
set -e

SLUG=$1

echo "Updating ${SLUG}"

# move the example implementation to .meta
mv ./${SLUG}/src/example.* ./${SLUG}/.meta/

# move the stubs to root of exercise
mv ./${SLUG}/src/* ./${SLUG}/

# move the tests to the root of the exercise
mv ./${SLUG}/test/*.c ./${SLUG}/

# make new test-framework dir
mkdir ./${SLUG}/test-framework

# move vendor files to test framework
mv ./${SLUG}/test/vendor/* ./${SLUG}/test-framework/

# remove old, now unused, dirs
rm -rf ./${SLUG}/test ./${SLUG}/src

# add a copy of the updated makefile
# THIS OVERWRITES ANY PER-EXERCISE CUSTOMIZATIONS!
cp ./makefile ./${SLUG}/

# update the include paths in the tests file
sed -i "0,/vendor/{s/vendor/test-framework/}" ./${SLUG}/test_*.c
sed -i "0,/..\/src\//{s/..\/src\///}" ./${SLUG}/test_*.c

# update the UnityBegin() line in the tests file
sed -i "s/UnityBegin(\"test\//UnityBegin(\"/" ./${SLUG}/test_*.c
```


wolf99:
Tests will fail on this due to Indent complaining about the Unity code in a directory we havent told indent not to ignore. 
The latest commit on this change fixes that, but CI will still fail until this is merged

wolf99:
Shoot - need updates to the exercise `.meta/config.json` files.... Will take that on a separate PR I think. as there's already a lot of change in this one.

ryanplusplus (approved):

PR #630, opened 2021-02-23
Linked list

wolf99:
Updates linked list to the current spec.

A question to consider when reviewing: two of the functions in the example implementation return values that are never checked. This is because while it is idiomatic in C to return these, the test specification does not have any need to specify them as expected - do we test them or leave them as-is?

1. The benefit of leaving them untested is conformance to the spec and leaving a wider solution space.
2. The benefit of testing them is communicating C idioms to the student and being slightly more specific. (however, see below)

The functions in question:

``` c
struct list_node *list_push(struct list * list, ll_data_t data);
struct list_node *list_unshift(struct list * list, ll_data_t data);
```

If we are to test them:
- Do we add this for every use of the mentioned functions or just some?
- What exactly do we test? The returned pointer can be tested for non-null and that the data value is correct... but does this realise the value from 2. above?

### DIFF
--- a/exercises/practice/linked-list/.meta/example.c
+++ b/exercises/practice/linked-list/.meta/example.c
@@ -107,6 +107,32 @@ ll_data_t list_shift(struct list * list)
    return result;
 }
 
+void list_delete(struct list *list, ll_data_t data)
+{
+   assert(list);
+
+   struct list_node *node = list->first;
+
+   while(node) {
+      if (node->data == data)
+      {
+         if (node == list->first)
+            list->first = node->next;
+         else
+            node->prev->next = node->next;
+
+         if (node == list->last)
+            list->last = node->prev;
+         else
+            node->next->prev = node->prev;
+
+         free(node);
+         return;
+      }
+      node = node->next;
+   }
+}
+
 void list_destroy(struct list *list)
 {
    if (!list)
--- a/exercises/practice/linked-list/.meta/example.c
+++ b/exercises/practice/linked-list/.meta/example.c
@@ -6,6 +6,7 @@ struct list_node {
    struct list_node *prev, *next;
    ll_data_t data;
 };
+
 struct list {
    struct list_node *first, *last;
 };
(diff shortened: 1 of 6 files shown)


wolf99:
Not sure how this can fail indent check after indent has been run in the same env? 🤔

exercises/practice/linked-list/.meta/example.c
```diff
@@ -34,7 +34,7 @@ struct list *list_create(void)
    return list;
 }
 
-size_t list_count(const struct list *list)
+size_t list_count(const struct list * list)
```
wolf99, line 37: This...

exercises/practice/linked-list/.meta/example.c
```diff
@@ -49,7 +49,7 @@ size_t list_count(const struct list *list)
    return count;
 }
 
-struct list_node *list_push(struct list * list, ll_data_t data)
```
wolf99: ... seems odd

exercises/practice/linked-list/.meta/example.h
```diff
@@ -1,30 +1,33 @@
...
 
-// checks if the list is empty
-bool list_is_empty(const struct list *list);
+// counts the items on a list
+size_t list_count(const struct list *list);
 
-// inserts item at back of list
-bool list_push(struct list *list, ll_data_t item_data);
+// inserts item at back of a list
+struct list_node *list_push(struct list *list, ll_data_t item_data);
```
ryanplusplus, line 16: Is this return value necessary? I don't think it's being used anywhere and it seems like there's very little good that a client could do with it.

ryanplusplus: It looks like the existing implementation has been returning a `struct list_node *` but was advertised as returning a `bool` (I'm not sure how the compiler didn't get upset about this...). I think the real fix is to not return anything, though.

guygastineau: @ryanplusplus I agree it seems weird to return the new node after pushing or unshifting.  I don't mind the idea of returning `bool` though.  I also don't mind the idea of some sort of enum return value that indicates the status of the operation, but I recognize that might seem like overkill (it certainly goes well beyond the spec).

guygastineau:
@wolf99 I think that a value returned is a value that should be checked in the tests.  I don't think there is any reason to worry about deviating from the spec with such additions.  It is quite common in multiple tracks to do so.

exercises/practice/linked-list/.meta/example.h
ryanplusplus: I'm fine with an addition to the spec as long as it's tested.

ryanplusplus (approved):

PR #633, opened 2021-02-26
🤖 Update tests.toml files to latest spec

ErikSchierboom:
Track maintainers found that they wanted to add comments to the tests.toml file to e.g. indicate _why_ a test was not included.
Unfortunately, running configlet sync would re-generate the entire file so any manually added comments were lost.

In this PR we're updating the format of tests.toml files to support adding comments.
We do this by creating a separate table for each test case which has `description` and `include` fields.
Tracks are then free to add additional fields, like a `comment` field, but also any other fields they feel might be useful to them.

configlet has _not_ yet been updated to support this new format, but we hope to do this soon. Sorry for the inconvenience.

For more information, see this discussion: https://github.com/exercism/configlet/issues/186

## Implementation

The PR expects the tests.toml files to be in their original format:

```toml
# <description>
"<uuid>" = <include>
```

This is transformed to:

```toml
[<uuid>]
description = "<description>"
include = <include>
```

## Example

```toml
# one factor has multiples within limit
"361e4e50-c89b-4f60-95ef-5bc5c595490a" = true
```

becomes

```toml
[361e4e50-c89b-4f60-95ef-5bc5c595490a]
description = "one factor has multiples within limit"
include = true
```

## Existing comments

As some tracks have manually added comments to tests, we try to detect them by assuming they are either:

- Added as a line comment _before_ the description (we'll ignore empty lines)
- Added as an inline comment _after_ boolean include value

For any such manually detected comments, we'll add a `comment` field.

## Tracking

https://github.com/exercism/v3-launch/issues/22

### DIFF
--- a/exercises/practice/acronym/.meta/tests.toml
+++ b/exercises/practice/acronym/.meta/tests.toml
@@ -1,28 +1,39 @@
-[canonical-tests]
+# This is an auto-generated file. Regular comments will be removed when this
+# file is regenerated. Regenerating will not touch any manually added keys,
+# so comments can be added in a "comment" key.
 
-# basic
-"1e22cceb-c5e4-4562-9afe-aef07ad1eaf4" = true
+[1e22cceb-c5e4-4562-9afe-aef07ad1eaf4]
+description = "basic"
+include = true
 
-# lowercase words
-"79ae3889-a5c0-4b01-baf0-232d31180c08" = true
+[79ae3889-a5c0-4b01-baf0-232d31180c08]
+description = "lowercase words"
+include = true
 
-# punctuation
-"ec7000a7-3931-4a17-890e-33ca2073a548" = true
+[ec7000a7-3931-4a17-890e-33ca2073a548]
+description = "punctuation"
+include = true
 
-# all caps word
-"32dd261c-0c92-469a-9c5c-b192e94a63b0" = true
+[32dd261c-0c92-469a-9c5c-b192e94a63b0]
+description = "all caps word"
+include = true
 
-# punctuation without whitespace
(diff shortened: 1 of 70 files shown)


wolf99:
Hi @ErikSchierboom is merging this dependent on the configlet update?

That is, if we merge this change right now and then run configlet will it complain that the `tests.toml` is the wrong format - or just overwrite it? Or maybe it's a chicken/egg situation and we should change this and then *not* run configlet on it until configlet is updated?

wolf99:
Running configlet against the tests.toml in this branch shows that it does not recognise the tests when specified in this new format. 

Leaving this PR unmerged until the configlet catches up.

ErikSchierboom:
> Running configlet against the tests.toml in this branch shows that it does not recognise the tests when specified in this new format. 

That is correct, sorry about that.

wolf99:
NP @ErikSchierboom . Is there a PR or work ongoing to get configlet up on this? I'll try to help out if I can.

ErikSchierboom:
@wolf99 There is no PR being worked on at the moment. The spec can be found here: https://github.com/exercism/configlet/issues/198 Help is much appreciated!

ee7:
As of the [latest configlet release](https://github.com/exercism/configlet/releases/tag/4.0.0-alpha.14), `configlet sync` now supports the new `tests.toml` format.

Please let me know if you find any problems with it.

wolf99:
Changes look good.
Have run newest configlet locally and this seems to result in the same output as earlier configlet versions with the existing tests.toml files.

wolf99 (approved):

wolf99:
This missed linked-list again @ErikSchierboom 😔

PR #634, opened 2021-02-27
Correct workflow names

### LINKED ISSUE #629: CI: split tests from checks and add clang build
It would be good to have the exercise tests built on their own job, separate job to the "checks" (`verify-indent`, `check-unitybegin`, `verify -unity-version`, `shellcheck`).

It would also be good, once the above is in place, to build the tests with Clang (also on its own job) as well as with GCC, as used to happen when Travis CI was used.
It seems the Ubuntu 18.04 GH job runner virtual environment that we use has both Clang and GCC available (https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md) and actually I'm not too sure which is being used currently - I assumed `$CC` picked up GCC automagically but am not really sure!

Thoughts?

wolf99:
Hopefully, fixes #629

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,9 +3,9 @@ name: Test Exercises
 on: [push, pull_request]
 
 jobs:
-  strategy:
-    matrix:
-      compiler: [gcc, clang]
+  # strategy:
+  #   matrix:
+  #     compiler: [gcc, clang]
 
   test:
     runs-on: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


wolf99:
Nope, still not running the `build.yaml` for some reason - any ideas?

wolf99:
Thats done it! 🕺

ryanplusplus:
Thanks! Any idea why the original change didn't fail CI?

ryanplusplus (approved):

wolf99:
Seems like if the job configs themselves are wrong they just don't run. Without running it couldn't fail.

PR #661, opened 2021-05-29
Stricter tests for grade-school

### LINKED ISSUE #660: grade-school: student name pointer assignment
Hello!

Thank you for the brilliant work you are doing!

---

After finishing this exercise and reading some of the community solutions, I saw that some of the solutions passed even though, in my opinion, they are technically not correct.

Student structure:
```
#define MAX_NAME_LENGTH 20

typedef struct {
   uint8_t grade;
   char *name;
} student_t;
```
The prototype of `add_student()`:

```
bool		add_student(const char* name ,uint8_t grade);
```
`name` from `student_t` struct is a pointer, and assigning the pointer to the `name` parameter passed to the function pass the tests.

Let's say we have a roster variable:

```
static roster_t roster;
```

If in the `add_student()` function we do:

```
roster.students[i].name = name;
```

Both pointers will point to the same address. The tests will pass because the parameter is the `expected->students[i].name`.

---

Thank you for your time!

ryanplusplus:
Fixes #660

### DIFF
--- a/exercises/practice/grade-school/.meta/example.c
+++ b/exercises/practice/grade-school/.meta/example.c
@@ -29,7 +29,7 @@ bool add_student(char *name, uint8_t grade)
    bool added = false;
 
    if (roster.count < MAX_STUDENTS) {
-      roster.students[roster.count].name = name;
+      strcpy(roster.students[roster.count].name, name);
       roster.students[roster.count].grade = grade;
       ++roster.count;
       qsort(roster.students, roster.count, sizeof(roster.students[0]),
@@ -52,8 +52,8 @@ roster_t get_grade(uint8_t grade)
    for (size_t i = 0; i < roster.count && grade_roster.count < MAX_STUDENTS;
         ++i) {
       if (roster.students[i].grade == grade) {
-         grade_roster.students[grade_roster.count].name =
-             roster.students[i].name;
+         strcpy(grade_roster.students[grade_roster.count].name,
+                roster.students[i].name);
          grade_roster.students[grade_roster.count].grade = grade;
          ++grade_roster.count;
       }
@@ -64,7 +64,7 @@ roster_t get_grade(uint8_t grade)
 void clear_roster(void)
 {
    for (size_t i = 0; i < roster.count; ++i) {
-      roster.students[i].name = NULL;
+      roster.students[i].name[0] = '\0';
       roster.students[i].grade = 0;
    }
    roster.count = 0;
--- a/exercises/practice/grade-school/.meta/example.c
+++ b/exercises/practice/grade-school/.meta/example.c
@@ -28,7 +28,7 @@ bool add_student(char *name, uint8_t grade)
 {
    bool added = false;
 
-   if (roster.count < MAX_STUDENTS) {
+   if (roster.count < MAX_STUDENTS && strlen(name) <= MAX_NAME_LENGTH) {
       strcpy(roster.students[roster.count].name, name);
       roster.students[roster.count].grade = grade;
       ++roster.count;
@@ -63,9 +63,5 @@ roster_t get_grade(uint8_t grade)
 
 void clear_roster(void)
 {
-   for (size_t i = 0; i < roster.count; ++i) {
-      roster.students[i].name[0] = '\0';
-      roster.students[i].grade = 0;
-   }
-   roster.count = 0;
+   memset(&roster, 0, sizeof(roster));
 }
--- a/exercises/practice/grade-school/.meta/example.c
+++ b/exercises/practice/grade-school/.meta/example.c
@@ -28,7 +28,7 @@ bool add_student(char *name, uint8_t grade)
 {
    bool added = false;
 
-   if (roster.count < MAX_STUDENTS && strlen(name) <= MAX_NAME_LENGTH) {
+   if (roster.count < MAX_STUDENTS && strlen(name) < MAX_NAME_LENGTH) {
       strcpy(roster.students[roster.count].name, name);
       roster.students[roster.count].grade = grade;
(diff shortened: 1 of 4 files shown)


sylenth1cs50 (approved):
Overall looks great. Thank you!

The only suggestion I have is to consider `snprintf`, even though `strcpy` is perfectly fine now because the length of the names is lower than `MAX_NAME_LENGTH`.

exercises/practice/grade-school/.meta/example.h
```diff
@@ -10,7 +10,7 @@
 
 typedef struct {
    uint8_t grade;
-   char *name;
+   char name[MAX_NAME_LENGTH + 1];
```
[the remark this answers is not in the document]
ryanplusplus: Good catch! Fixed.

exercises/practice/grade-school/.meta/example.c
```diff
@@ -29,7 +29,7 @@ bool add_student(char *name, uint8_t grade)
    bool added = false;
 
    if (roster.count < MAX_STUDENTS) {
```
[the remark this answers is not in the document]
ryanplusplus: :+1:

exercises/practice/grade-school/.meta/example.c
```diff
@@ -64,7 +64,7 @@ roster_t get_grade(uint8_t grade)
 void clear_roster(void)
 {
    for (size_t i = 0; i < roster.count; ++i) {
-      roster.students[i].name = NULL;
+      roster.students[i].name[0] = '\0';
       roster.students[i].grade = 0;
    }
    roster.count = 0;
```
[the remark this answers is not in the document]
ryanplusplus: Good call, fixed.

sylenth1cs50 (approved):
Thank you very much, @ryanplusplus!

---

Everything looks good to me.

I appreciate what you and the other maintainers are doing.

wolf99:
Think this was purposely done this way so as to make the exercise more about ordering than messing around with memory so as to be able to copy strings around the place.
Having said that this does implementation does not seem like much extra burden on the student.

exercises/practice/grade-school/test_grade_school.c
```diff
@@ -13,10 +14,11 @@ void tearDown(void)
 
 static void populate_roster(roster_t * expected)
 {
-   for (size_t i = 0; i < expected->count; ++i)
-      TEST_ASSERT_TRUE(add_student
-                       (expected->students[i].name,
-                        expected->students[i].grade));
+   for (size_t i = 0; i < expected->count; ++i) {
+      char name_buffer[MAX_NAME_LENGTH + 1];
```
wolf99, line 18: `MAX_NAME_LENGTH` is part of the implementation, technically students are free to change it or remove it completely... this would then break.

exercises/practice/grade-school/.meta/example.c
```diff
@@ -28,8 +28,8 @@ bool add_student(char *name, uint8_t grade)
 {
    bool added = false;
 
-   if (roster.count < MAX_STUDENTS) {
-      roster.students[roster.count].name = name;
+   if (roster.count < MAX_STUDENTS && strlen(name) <= MAX_NAME_LENGTH) {
```
wolf99, line 31: If we consider the string terminator to be part of the name length, then this can chane from `<=` to `<` and the `+1`'s won't be needed when declaring the `name[]`  string in the header file.

```suggestion
   if (roster.count < MAX_STUDENTS && strlen(name) < MAX_NAME_LENGTH) {
```

exercises/practice/grade-school/test_grade_school.c
ryanplusplus: If it's changed but not removed, this will change accordingly because it's using the macro from the implementation's header.

exercises/practice/grade-school/.meta/example.c
ryanplusplus: Good suggestion, I made these changes.

ryanplusplus:
> IIRC this was exercise was purposely implemented this way so as to make the exercise focus more on sorting the pointers based on the contents (e.g. `qsort()` and `strncmp()`) rather than copying strings around the place for the sake of it and the memory management that can creep in to student solutions to allow that. Part of our previous effort to have more logic and less string manipulation in the track.
> 
> Having said that this implementation does not seem like much extra burden on the student.

Your call, I can go either way.

exercises/practice/grade-school/grade_school.h
```diff
@@ -9,7 +9,7 @@
 
 typedef struct {
    uint8_t grade;
-   char *name;
+   char name[MAX_NAME_LENGTH + 1];
```
wolf99, line 12: Missed one 😉 

```suggestion
   char name[MAX_NAME_LENGTH];
```

ryanplusplus: Whoops, thanks

exercises/practice/grade-school/test_grade_school.c
wolf99: > Your call, I can go either way. If we decide to keep the existing behavior, we should remote `MAX_NAME_LENGTH` since it is inconsistent with using statically allocated strings.

Go with this new approach, if we get feedback that possible solutions are too complex then we can adjust again a needed.

wolf99:
>Your call, I can go either way. If we decide to keep the existing behavior, we should remote MAX_NAME_LENGTH since it is inconsistent with using statically allocated strings.

Go with this new approach, if we get feedback that possible solutions are too complex then we can adjust again a needed.

wolf99 (approved):

PR #675, opened 2021-07-28
Add icons for key features

ee7:
The icons for key features are now available, and `configlet lint` now requires that each key feature has a valid `icon` value.

This PR sets an initial icon for each key feature - feel free to change them however you want. You can see the list of available icons [here](https://github.com/exercism/docs/issues/189). Note that an icon can be used regardless of its name.

With the initial state of this PR, the key features look something like the below.

---

cross-platform Cross-platform
C is portable and can run on almost any platform and chipset.

general-purpose General purpose
C can be used for a wide variety of workloads from IoT to game engines.

fast Performant
C has been consistently rated as one of the highest performing languages for energy and time.

stable Stable
The C standard values stability and backwards compatibility, C programs can run for decades.

powerful Powerful
C allows both lower and higher level functionality, making it very powerful.

fast Efficient
C allows swift implementation using constructs that map efficiently to machine instructions.

### DIFF
--- a/config.json
+++ b/config.json
@@ -996,32 +996,32 @@
   "concepts": [],
   "key_features": [
     {
-      "icon": "TODO",
+      "icon": "cross-platform",
       "title": "Cross-platform",
       "content": "C is portable and can run on almost any platform and chipset."
     },
     {
-      "icon": "TODO",
+      "icon": "general-purpose",
       "title": "General purpose",
       "content": "C can be used for a wide variety of workloads from IoT to game engines."
     },
     {
-      "icon": "TODO",
+      "icon": "fast",
       "title": "Performant",
       "content": "C has been consistently rated as one of the highest performing languages for energy and time."
     },
     {
-      "icon": "TODO",
+      "icon": "stable",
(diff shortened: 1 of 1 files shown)


wolf99:
The change looks good to me.
Though maybe the double use of the `fast` icon indicates that we could rethink the `efficient` feature. Either to reword to align a little with another icon, or to replace it with a different feature (maybe `portable` or `embeddable`).
What do you think?

ee7:
> Though maybe the double use of the `fast` icon indicates that we could rethink the `efficient` feature.

I agree. I couldn't see how to avoid the duplicate icon here.

> or to replace it with a different feature (maybe `portable` or `embeddable`).

I lean towards this. But perhaps `portable` overlaps too much with the existing "Cross-platform" wording?

I guess `widely-used` is possible too. Maybe just look at key features on other tracks and steal one that fits?

If you're okay with the icons otherwise - should we just merge this PR, and then you can follow-up however you want?

wolf99 (approved):

PR #676, opened 2021-08-04
beer-song: update to latest spec

### LINKED ISSUE #549: beer-song: use arrays of strings rather than new-line delimited string for output
The `beer-song` exercise is updated in #548  to align with the problem specification.
One thing was skipped during that update however.

The problem specification has changed from using a single continuous string with new-lines as sentence delimiters, to using an array of strings, for the output from the unit under test (UUT).

The implementation should change to follow the problem specification
Changing this may also allow the addition of the test that checks the full song (currently this is set to false in the exercise's `tests.toml`).

wolf99:
fixes #549

### DIFF
--- a/exercises/practice/beer-song/.meta/example.c
+++ b/exercises/practice/beer-song/.meta/example.c
@@ -1,46 +1,31 @@
 #include "beer_song.h"
+#include <stdint.h>
 #include <stdio.h>
-#include <string.h>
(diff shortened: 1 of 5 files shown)


ryanplusplus (approved):

ryanplusplus:
I like this approach :+1:

PR #683, opened 2021-08-23
Re-achitect grade-school

wolf99:
The problem specification for grade-school has some issues.
Trying to re-architect the `grade-school` _tests_ here so as to build a new problem specification based on and with referenced to a working example.
The implementation is relatively unchanged.

`gcov` tells me the new tests exercise all lines of, so I start from here and see if a problem specification that reflects these new tests is acceptable.
Having said that, there are some TODOs currently left in these tests. I plan to resolve these in later commits, so that planned additional changes are easier to follow.


wolf99:
I don't like relying on the unit under test for the `tearDown()`, but didn't see a better way.
As the `clear_roster()` is not part of the exercise, more of a part of how things need to be implemented in C, it is not tested... perhaps we should give away the implementation "for free" and then we at least know that the implementation starts of as correct for the student.
What do you think?

ryanplusplus:
> I don't like relying on the unit under test for the `tearDown()`, but didn't see a better way. As the `clear_roster()` is not part of the exercise, more of a part of how things need to be implemented in C, it is not tested... perhaps we should give away the implementation "for free" and then we at least know that the implementation starts of as correct for the student. What do you think?

Perhaps it would be better to have an init function that is called in `setup()` instead of having a clear function called in `teardown()`. It feels odd to me to assume that the component is ready to go straight away but must be cleaned up afterward. Having it be set up prior to use may be more natural for students.

wolf99:
> a part of how things need to be implemented in C

Whoops. This is complete waffle. 😆 
There a much more simple interface that I wasn't seeing the wood for the trees of.

CI link check is failing for a link that recently started to time out sporadically - I'll change that link on a separate PR.

Issue #692, opened 2021-09-09
run-length-encoding: add tests for invalid input.


braoult:
I think we could add 2 more tests with invalid input:
- encode() with a string containing digits
- decode() with a string ending with digits

If it is OK for you, I have a patch ready, and I can do a PR right now.

ryanplusplus:
Hey @braoult, since those test cases don't cover C-specific behavior, the process is to first get those tests upstreamed at:
https://github.com/exercism/problem-specifications

braoult:
@ryanplusplus, I had a look at different discussions on [problem-specification issues](https://github.com/exercism/problem-specification)s about input validation, it looks this is a recurrent subject, and _mission impossible_ to add  data validation requirements: All I have seen were rejects, for the same reason (that I don't totally agree with) : The point is to focus on a specific point, not to write a (tentatively bullet-proof) _correct_ program.

So I will do on my side the tests as I consider they should be (in this specific exercise, I want at least to always have `s == (encode(decode(s) ` and `s == decode(encode(s))` for valid input, `NULL` otherwise. I may also ask students (as side exercises, if they wish) what they could do to make their solution "more robust".

If you want, I can send you the PR anyway, that you could keep somewhere if you want, just in case. Just let me know. After that, I will close this issue.

ryanplusplus:
For now I'd rather not diverge from upstream, but if you'd like to create the PR and close it so that we have your patch available in your history that sounds like a good approach to me.

braoult:
I made the PR. Apparently there is an issue that I did not notice on my side. Which source/solution is used to run tests ? As it is the first time I play with that, I should modify this also, I suppose ?

braoult:
Now that the PR is full (#694, with correct indent as expected by `indent.sh`), what would be the next step ? In fact I am unsure what happens if I close the PR as you suggest... Won't it simply be lost ?

ryanplusplus:
> In fact I am unsure what happens if I close the PR as you suggest... Won't it simply be lost ?

Actually, once a PR is opened, even if it's closed, its contents aren't lost. GitHub will keep it around so that we can reopen it if the upstream problem specifications change or we change our mind about taking on additional test cases.

braoult:
OK, something like a `git stash`, maybe ? So I closed the PR, and deleted my branch.

ryanplusplus:
Yeah, more or less. It's special GitHub magic over and above the normal git magic.

braoult:
[...]

PR #693, opened 2021-09-10
run-length-encoding: add some invalid tests

braoult:
As discussed in #692.


braoult:
misunderstood how to make changes in tests cases (we could need to edit example.c).

PR #694, opened 2021-09-11
Run length encoding

braoult:
As discussed in #692: 2 new tests & corresponding code in .meta/example.c. I hope nothing is missing.


braoult:
new tests + example.sh + config.json + fix with `indent.sh`

braoult:
All ready (including indent.sh, that I missed too). I have one questions here, as this is the first time I do this :
What would be the next steps (to finish the commit, so it is "not accepted but not lost") ?

braoult:
Closing as discussed in #692.

PR #696, opened 2021-09-17
All your base tests fix

Nerwosolek:
I have changed two tests: test_single_zero, test_multiple_zeros.
They have expected 0 length result array but as input and output bases are correct these are just proper representations of zero number.
So the results should be zeros, which in both cases are arrays of length 1 with zero digit.


wolf99 (dismissed):
LGTM, thanks!

Nerwosolek:
As soon as I am not on mobile too :)

Nerwosolek:
OK. So it looks like there are reference solutions which are tested with updated tests and they are failing. Which is good bc now expected value is 1 instead of 0.

Nerwosolek:
I can update the example solution tommorrow.

wolf99 (changes requested):

wolf99 (approved):

wolf99:
LGTM

Issue #680, opened 2021-08-20
Check docs are up to date


ErikSchierboom:
Please check if your documentation files are still up-to-date.

The key documentation files to check are:

- `docs/ABOUT.md`
- `docs/INSTALLATION.md`
- `docs/LEARNING.md`
- `docs/RESOURCES.md`
- `docs/TESTS.md`
- `exercises/shared/.docs/help.md`
- `exercises/shared/.docs/tests.md`

There might be more.

## Link check report

To help identify invalid links, we've automatically checked the links of _all_ `*.md` files in this repo.
This is the report of that check:

<pre style="color:#bbb;white-space:pre-wrap;word-wrap:break-word;overflow-wrap:break-word">📝 Summary
---------------------
🔍 Total..........262
✅ Successful.....261
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........0
🚫 Errors...........1

Errors in ./docs/INSTALLATION.md
<span style="color:#ff5555">✗ https://msdn.microsoft.com/en-us/commandline/wsl/about (HTTP status client error (403 Forbidden) for url (https://msdn.microsoft.com/en-us/commandline/wsl/about))</span>

</pre>
## Tracking

https://github.com/exercism/v3-launch/issues/54

wolf99:
Hi @ErikSchierboom ,

What are you using to check links? 
We could add this into the CI, or perhaps it could be incorporated into the configlet utility?

ErikSchierboom:
@wolf99 https://github.com/lycheeverse/lychee I'm thinking of making it a GitHub Action.

PR #706, opened 2021-10-13
Add link checker for markdown files

wolf99:
Inspired by: https://github.com/exercism/c/issues/680#issuecomment-904353708


wolf99:
Hrm, test is not failing as it should because lychee action handles errors itself to generate the output.

wolf99:
FYI this is correctly failing. 
There are two dead links in our docs.
I intend to open a separate PR to resolve those.

PR #713, opened 2021-10-27
Add CI check for include guards

wolf99:
Inspired by #711 from @siebenschlaefer 

The bash script could probably be better but I am very rusty on bash.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -13,31 +13,31 @@ jobs:
     steps:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
 
-      - name: Shell script static analysis
-        run: shellcheck bin/fetch-configlet bin/verify-unity-version bin/check-unitybegin bin/run-tests format.sh
+      # - name: Shell script static analysis
+      #   run: shellcheck bin/fetch-configlet bin/verify-unity-version bin/check-unitybegin bin/run-tests format.sh
 
-      - name: Check practices formatting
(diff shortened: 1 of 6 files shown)


wolf99:
So it looks likes the issue is due to the version of grep available on the different Ubuntu versions used in the images Github uses on the runners.

Ubuntu 18.04 ships with grep 3.2 by default, whereas Ubuntu 20.04 ships with grep 3.4 by default.

It seems like [grep 3.3](http://savannah.gnu.org/forum/forum.php?forum_id=9334) didn't have many changes of significance, but [grep 3.4](https://savannah.gnu.org/forum/forum.php?forum_id=9640) has a fix for [a bug](https://lists.gnu.org/archive/html/bug-grep/2019-10/msg00006.html) relating to the `-L` flag, which this script uses an uses in the way described in the bug.

The resulting question then is, are we OK to update to using Ubuntu 20.04 runners in the CI?

It looks like GitHub has moved them from "tentative" to standard as the note relating to it's use has been removed from https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

If the answer is yes, then I can make a separate PR for that and then pick this one up again afterwards.
If the answer is no, it would be useful to have suggestions for other simple methods to check/grep for multi-line matches.

ryanplusplus:
I have no issue updating to 20.04 👍

wolf99:
And now this is also ready for review 🙂

ryanplusplus (approved):

PR #715, opened 2021-11-01
Fix some small markdown issues

wolf99:
Noticed some small issues reported by markdown linter

### DIFF
--- a/reference/README.md
+++ b/reference/README.md
@@ -6,7 +6,7 @@ The C concept exercises are based on language concepts. The list below contains 
 
 ### Specialities
 
-#### Speciality Concepts
+#### Specialty Concepts
 
 - [Algorithms](https://en.cppreference.com/w/c/algorithm)
 - [Bit Fields](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Bit-Fields)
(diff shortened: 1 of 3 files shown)


reference/README.md
```diff
@@ -6,7 +6,7 @@ The C concept exercises are based on language concepts. The list below contains
 
 ### Specialities
 
-#### Concepts
+#### Speciality Concepts
```
ryanplusplus, line 9: Do you want "specialty" here?

wolf99: Because there are two headings at the H3 level, Specialities and General, and these each have matching sub-headings (i.e. at the H4 level), it means there are two `Concept` headings and two `Types` headings. 

This runs afoul of [MD024/no-duplicate-heading/no-duplicate-header](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md024---multiple-headings-with-the-same-content)

Prefixing the H4 headings seemed like a quick fix that keeps the meaning clear and explicit.

wolf99: Ack, UK-Eng  vs US-Eng strikes again!

ryanplusplus (approved):

reference/README.md
ryanplusplus: Hmm, where are you seeing that it's specialities in the US? I've never seen that in use, but it may be a regional thing.

PR #723, opened 2021-12-07
Clang format

### LINKED ISSUE #716: Move from GNU indent to ...?
On Ubuntu 20.04 the version of GNU Indent available is 2.2.12.

This has [a bug](https://savannah.gnu.org/bugs/?54673) that causes the `--space-after-if`/`-sai` to consistently misbehave, meaning that Indent can never be satisfied. This argument is used in our use case as part of the `-linux` shortcut argument that we currently specify.

If we are to update to using Ubuntu 20.04 (see #713, #717 ), then we will need to overcome this issue.

As the [last release](https://git.savannah.gnu.org/cgit/indent.git) of Indent was 3 years ago (2019), after the bug was filed (2018), and the release before that 12 years ago, I don't see Indent being updated to resolve this issue any time soon. Therefore I see several bad options and one other option:

- Split the Indent task to its own workflow and continue using Ubuntu 18.04 there (not great to continue requiring old versions, especially for local development)
- Change our arguments to indent (not great as a space after an if statement is quite a widespread usage
- Try to change exercises so that this bug is never encountered (not great as function pointers are a useful part of the language and should be part of at least one exercise)
- Change from using indent to using something else (e.g. astyle)

Thoughts?

---

Information on the GNU Indent bug and how it affects this track:

In the list-ops practice exercise `example.c`, the `filter_list()` function checks that a function pointer argument is not `NULL`.
1. On a pass of Indent it seems like Indent determines that the `if` that checks the function pointer variable is actually a  function named `if()` and so moves the opening bracket to be the next character directly after the `if`.
2. On a subsequent pass of Indent it decides differently that actually it is an `if` statement and so adds the space back in before the opening bracket.
3. GOTO 1;

To reproduce in a minimum verifiable example, create a `test.c` file

```c
void foo(void (*bar)(void))
{
   if (!bar)
      return;
}
```

Then in a bash terminal run

```console
$ indent -linux -i3 -nut ./test.c  # First pass
$ diff ./test.c ./test.c~  # Shows indent has removed space after if
4c4
<    if (!bar)
---
>    if(!bar)
$ indent -linux -i3 -nut ./test.c  # Second pass
$ diff ./test.c ./test.c~  # Shows indent has added space after if
4c4
<    if(!bar)
---
>    if (!bar)

# This continues eternally
```

wolf99:
Fixes #716


wolf99:
Hmmm... well /format seems to have stopped working!

wolf99:
@ErikSchierboom any idea why the /format no longer works ?
I think we haven't used it in a while, but it should cause this workflow to create a checkout code, run the formatting script (which works per local tests) and then commit the changes.
https://github.com/exercism/c/blob/main/.github/workflows/format-code.yml

I seem to recall some organization permissions or token was needed when it was initially created, are you able to check if that is still in place?

wolf99:
Ready for review!

If merging, please squash  🤏

.github/workflows/checks.yml
```diff
@@ -4,21 +4,24 @@ on: [push, pull_request]
...
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
-      - name: Install Indent
+      - name: Install Dependencies
         run: |
           sudo apt-get -qq update
-          sudo apt-get install -y indent
+          # sudo apt-get install -y clang-format
```
ryanplusplus, line 13: Should this be commented out?

.clang-format
```diff
@@ -0,0 +1,60 @@
+---
```
ryanplusplus, line 1: :partying_face:

docs/CONTRIBUTING.md
```diff
@@ -62,23 +62,17 @@ The pull request points to that branch, not to specific commits.
...
 
-On Linux `indent` should be available for install via your distribution's package manager.
-On a Mac you can install the package `gnu-indent` using [Homebrew][].
+On Linux `clang-format` should be available for install via your distribution's package manager.
+On a Mac you can install the package `clang-format` using [Homebrew][].
 
-Indent can be run on all files by executing `indent.sh`.
-To manually run it on a single file, you can execute:
+clang-format can be run on all files by executing `format.sh`.
+To manually run it on a single file, you can execute the following from the same location as the `.clang-format` file:
```
ryanplusplus, line 72: `clang-format` actually crawls up the directory tree to find the nearest parent folder containing a `.clang-format`. Just FYI, I don't think we need to make this more confusing by adding that nuance here.

docs/CONTRIBUTING.md
```diff
@@ -199,6 +188,7 @@ Read more about [test runners].
 [Continuous Integration]: ./CONTRIBUTING.md#continuous-integration
 [convert the PR to a draft]: https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/
 [style guide]: ./C_STYLE_GUIDE.md
+[clang-formt]: https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormat.html
```
ryanplusplus, line 191:
```suggestion
[clang-format]: https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormat.html
```

.github/workflows/checks.yml
```diff
@@ -4,21 +4,24 @@ on: [push, pull_request]
...
 
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
 
       - name: Shell script static analysis
-        run: shellcheck bin/fetch-configlet bin/verify-indent bin/check-unitybegin bin/run-tests indent.sh
+        run: shellcheck bin/fetch-configlet bin/verify-unity-version bin/check-unitybegin bin/run-tests format.sh
```
ryanplusplus, line 18: Does `format.sh` have a non-zero exit code when the formatter needs to make a change?

.github/workflows/format-code.yml
```diff
@@ -29,13 +29,12 @@ jobs:
           git clone $HEAD_REPO .
           git checkout -b "$HEAD_REF" "origin/$HEAD_REF"
 
-      - name: "Install indent"
+      - name: "Update"
         run: |
           sudo apt-get -qq update
-          sudo apt-get install -y indent
```
ryanplusplus, line 35: Does `clang-format` need to be installed or is it already in the base image? If it's already in the base image, we can remove the `sudo apt-get --qq update` since it just updates the package database.

wolf99: Looks like version 11 is in the base image on the runner 🙂

docs/CONTRIBUTING.md
wolf99: Good catch!

.github/workflows/checks.yml
wolf99: When checking if a change is needed, as in `checks.yaml`, the `--dry-run` argument is used. This results in a non-zero (`123`) exist code if incorrect formatting is detected. Which should then cause the GitHub workflow to fail, as expected.

When running `format.sh` to actually change the code to be correctly formatted, the `-i` (in-place edit) argument is used instead. If nay changes are needed clang-format makes them and the exit code is `0`. I think this is also as expected. Do you agree or do I misunderstand?

```console
# Per the checks.yaml used in CI
~/c$ find exercises/practice/ -path '*/test-framework' -prune -o \
>        \( -iname '*.h' -o -iname '*.c' \) -print0 |\
>        xargs --null clang-format --dry-run --Werror 
exercises/practice/diamond/test_diamond.c:37:13: error: code should be clang-formatted [-Wclang-format-violations]
      " A ",
            ^
exercises/practice/diamond/test_diamond.c:38:13: error: code should be clang-formatted [-Wclang-format-violations]
      "B B",
            ^
~/c$ echo $?
123       # i.e. incorrect formatting was detected as shown in the above output by the clang-format

# Per the format.sh script for use by contributors
~/c$ find exercises/practice/ -path 'exercises/practice/*/test-framework' -prune -o \
>        \( -iname '*.h' -o -iname '*.c' \) -print0 |\
>        xargs --null clang-format -i
~/c$ echo $?
0        # i.e. any needed formatting changes were correctly made
```

ryanplusplus: Sorry, I totally missed that this was the `shellcheck` step and somehow convinced myself this was the list of scripts to run.

ryanplusplus (approved):

PR #725, opened 2021-12-11
Stop running too much CI

wolf99:
We only need to run CI on push when it is a push to `main`.


ryanplusplus (approved):

wolf99:
Added in a quick name change for what was the `build` workflow, so as to match the Exercism docs (https://exercism.org/docs/building/tracks/new/setup-continuous-integration)

wolf99:
Yay, CI is now a bunch quicker! 🙌

PR #733, opened 2021-12-11
Space-age: update to latest spec

wolf99:


.clang-format
```diff
@@ -28,7 +28,7 @@ ColumnLimit: 80
 ContinuationIndentWidth: 4
 Cpp11BracedListStyle: false
 # IndentCaseBlocks: false
-IndentCaseLabels: false
+IndentCaseLabels: true
```
ryanplusplus, line 31: :+1:

exercises/practice/space-age/.meta/example.c
```diff
@@ -1,21 +1,41 @@
...
+   EARTH_YR_IN_SECONDS * 0.61519726,   // Venus
+   EARTH_YR_IN_SECONDS * 1,            // Earth
+   EARTH_YR_IN_SECONDS * 1.8808158,    // Mars
+   EARTH_YR_IN_SECONDS * 11.862615,    // Jupiter
+   EARTH_YR_IN_SECONDS * 29.447498,    // Saturn
+   EARTH_YR_IN_SECONDS * 84.016846,    // Uranus
+   EARTH_YR_IN_SECONDS * 164.79132     // Neptune
 };
 
+static bool isPlanet(planet_t planet)
```
ryanplusplus, line 19:
```suggestion
static bool is_planet(planet_t planet)
```

exercises/practice/space-age/.meta/example.c
```diff
@@ -1,21 +1,41 @@
 #include "space_age.h"
+#include <stdbool.h>
 
 /* Number of seconds in a Earth year */
-#define EARTH 31557600
+#define EARTH_YR_IN_SECONDS 31557600
```
ryanplusplus, line 5: I hate to be picky, but I think it makes more sense to avoid the abbreviation here:

```suggestion
#define EARTH_YEAR_IN_SECONDS 31557600
```

.clang-format
wolf99: `true` results in:

```c
   switch (planet) {
      case MERCURY:
      case VENUS:
      case EARTH:
      //...
```

`false` results in:

```c
   switch (planet) {
   case MERCURY:
   case VENUS:
   case EARTH:
   //...
```

The latter seems harder to read to me, but I don't really care too much so I'm checking only that the latter is your intention - is that the case?

ryanplusplus: Sorry, I was trying to say that I like the change :smile:

ryanplusplus (approved):

PR #735, opened 2021-12-12
Basics concept exercise

wolf99:
I've started with adding a proposed list of concepts to work towards.

This is based on the concepts listed in https://github.com/exercism/c/tree/main/reference, but I have not add everything from there as there is quite a bit of overlap.

Still, some of the concepts listed here are things that I am unsure should be included, or could be included differently. Some I'm not sure it's even possible to test.

Would love to get some input on the listing!


iHiD:
EXCITING!

bobahop:
I suggest adding the booleans concept. Truthiness differs from some other languages, since booleans are 0 or non-zero under the covers. Some languages are strict about that and give no boolean value to numbers. Could also mention cbool.h. Also, I've seen many C solutions with something like

```c
if (/* boolean expression here*/)
    return true;
 else
    return false;
// instead of
return /* boolean expression here */;
```
so something about boolean expressions might be helpful, too.

bobahop:
- Not sure what basics would include.
- Perhaps bit-fields and bit-manipulation could be one bitwise operations? I don't feel strongly about it.

I think the rest look good to a quick lookover, though I think some more might be added...

bobahop:
Perhaps add a looping concept. Many languages have ways to loop that C doesn't have, at least not out of the box. It might be good to tell students that so they don't go crazy looking for what C doesn't have. And also to explain the different looping constructs C does have.

bobahop:
I suggest adding a char concept. Some helpful things to cover about handling chars.

bobahop:
This may not be appropriate for a language concept, but many students ask me how I benchmark, so perhaps a benchmark  concept that explains how to benchmark C might be helpful.

ErikSchierboom (commented):
Awesome!

wolf99:
I've updated the `config.json` with the suggested Booleans, looping and chars.
I've also added some new concepts that I thought of

I have not added bench-marking as this can probably be covered in the Dates & Times concept.

@iHiD or @ErikSchierboom: If we add proposed concepts to the `config.json` like this in advance of actually adding the related files for all of those concepts, will that cause the site to break?

ErikSchierboom:
> If we add proposed concepts to the config.json like this in advance of actually adding the related files for all of those concepts, will that cause the site to break?

It will. Currently, configlet does not catch this, but I'll add this check later.

wolf99:
Thanks @ErikSchierboom .

I'll split this out from this PR then.

ErikSchierboom:
I have an [open PR to configlet](https://github.com/exercism/configlet/pull/522) to detect this issue.

ErikSchierboom (approved):

PR #738, opened 2021-12-20
Add booleans and conditionals concepts

bobahop:
This will add the concepts folder along with the booleans folder and the conditionals folder and their associated meta folder and files.


concepts/conditionals/about.md
```diff
@@ -0,0 +1,178 @@
...
+## Switch
+
+
+Sometimes an `if-else-if ladder` can be better represented by a
+`switch`. Like `if`, the expression being tested is always put
+in parentheses.
+
+```c
+switch (number) {
+    case -1: return "negative one";
```
wolf99, line 86: Would be good to put the case bodies (i.e. the `return foo;`)   etc on a new line in their own block, as this is the style we use in the exercises.
I think it helps to make it clearer what is part of the case and what the case matches versus what the the case does

wolf99: Same is true for the other switch cases.

concepts/conditionals/about.md
```diff
@@ -0,0 +1,178 @@
...
+```c
+    a = 1.345f;
+    b = 1.123f;
+    c = a + b;
+    // c may equal something like 2.4679999352
+    if (c == 2.468)
+        return "I won the lottery!";
+```
+
+A tolerance value, referred to as `epsilon`, can be used to determine a range
```
wolf99, line 176: Could there be an example of using an epsilion value in an if statement?
If this paragraph came before the above example, then probably that example could contain both the existing usage and an epsilion usage?

wolf99 (commented):
Sorry review took so long, I got busy with the holidays.
I hope you've enjoyed yours. 
Happy new year!

concepts/booleans/about.md
```diff
@@ -0,0 +1,77 @@
+# About
+
+A Boolean expression is a question asked of one or more values which can be
+answered with `true` or `false`. A true expression returns `1`; a false
```
wolf99, line 4: The C track tends to follow the line-break-on-sentence-end convention - could these be updated to follow that?

bobahop: Okay, so from what I can see at looking at some other docs, it's okay to have a very wide line (80+), just break after period, correct?

concepts/booleans/about.md
```diff
@@ -0,0 +1,77 @@
...
+`false` maps to `0`; and `bool` maps to `_Bool`. So now, instead of directly using 
+`int`s or rolling their own Boolean mappings, programmers can include `stdbool.h`
+and automatically get Boolean symbols in their code.
+
+## Comparison Operators
+
+C supports six comparison operators: `==` (Equal), `!=` (Not equal), `>` (Greater
+than), `<` (Less than), `>=`(Greater than or equal to), `<=` (Less than or equal
+to). These operators can compare values to evaluate the relationship of values
+into a boolean value.
```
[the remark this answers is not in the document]
bobahop: "These operators compare the relationships between values and resolve them into boolean values."?

bobahop:
Enjoyed the holidays until I woke up with a sore throat on New Years Eve morning. 🤧  Might have covid 😷,  but haven't been tested. 🧪  I'm quarantined to my couch. 🛋️  😄

concepts/booleans/about.md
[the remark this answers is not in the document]
wolf99: I guess the bit I feel is missing in these sentences is that it is the result of the comparison operation, rather than some kind of conversion of the operands.
Maybe I am being to pedantic ?
What do you think @ryanplusplus ?

wolf99:
Goodness, I hope you aren't feeling too unwell !
At least the couch isn't too bad a spot to quarantine 😅 .

bobahop:
Sorry, I clicked ready for review on the wrong PR. I'm still not thinking 100% clearly. We still need to resolve the bit about comparison operators.

concepts/booleans/about.md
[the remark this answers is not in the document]
wolf99: Hey @bobahop , I think we should go ahead and get this merged. This is a minor point only so let's skip it for now.

wolf99 (approved):

PR #739, opened 2021-12-21
add loops and recursion concepts

bobahop:
This adds the folders and files for the loops and recursion concepts.


concepts/loops/about.md
```diff
@@ -0,0 +1,163 @@
+# About
+
+A loop is a way of repeatedly calling one or more statements in a batch. Each
```
wolf99, line 3: Could switch to line-break-on-sentence-end convention 🙂

concepts/loops/about.md
```diff
@@ -0,0 +1,163 @@
...
+    }    
+    return 0;
+}
+```
+
+It is important for there to be a condition to end the loop; otherwise it is an
+infinite loop which will block all progress of the code. If the above code had
+the `++` removed from `i` it would be an infinite loop.
+
+## Do loop
```
wolf99, line 30: Great! I somehow often forget that do-while loops exist! 🤦

concepts/loops/about.md
```diff
@@ -0,0 +1,163 @@
...
+for (int i = 0; i < 5; i++) {
+    printf("Number is %i\n", i);
+}
+```
+
+This is less verbose than the `while` loop equivalent, and yet it is still
+readable.
+
+The initializing and updating sections of the `for` loop may handle more than
+one variable.
```
wolf99, line 72: "... more than one variable by using the comma operator to separate expression within the conditions"

Could be worth specifically calling out the use of the comma operator here?
People that are new to C often don't realise that `,` is an operator in itself.

wolf99 (commented):
Great stuff, just a couple of minor things to check.

wolf99 (approved):
This is great @bobahop , thanks for contributing!

wolf99:
Apologies Bob, I didn't realise this was not yet merged!

PR #741, opened 2022-01-17
add pointers concept

bobahop:
Still very much a WIP.


bobahop:
I haven't touched this in a while. I think it's ready for some critique.

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+
+Where are you?
+
+_I am at `0x6dfed4`_.
+
+Wait, what?
+
+Variables and constants store their value in a memory location.
+That location in memory is known as the address of the variable or constant.
+
```
wolf99, line 31: Worth mentioning that quite often addresses are displayed in hexadecimal format by convention?
Just a suggestion, not sure if this is needed or not.

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+## Address operator
+
+To get the address of a variable we use the address operator, which is `&`.
+
+
```
wolf99, line 36:
```suggestion

```

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+```c
+#include <stdio.h>
+
+int main() {
+    // my_int's value is its address
+    int my_ints[] = {1, 2};
+    
+    // since my_ints is already an address
+    // we don't need the address operator to make a pointer from it.
```
wolf99, line 72: Hold on, `*` is explained above, but `&` hasn't been explained yet...
Shouldn't that also be explained a little as part of the example on ln # 39?

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+
+Now that we understand how addresses work with array indexing, what good is it if we don't get the value stored at an address?
+How do we do that?
+To get the value stored at an address we use the dereference operator, which is, you guessed it, `*`.
+
+Come on!
+Are you kidding me?
+How many uses are there for `*`?!
+Multiplication, declaring a pointer, and now derefencing!
+It's confusing!
```
wolf99, line 114:
```suggestion
It's confusing!

Yes. Yes it is.
```
🤣 😄

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+If the type is four bytes in length, the second element will be four bytes higher than the first element.
+
+This is also why pointers are declared to be of a certain type.
+If a pointer did not know what type it points to, it would not know how many bytes to advance when adding `1` to it.
+Adding `1` to a pointer does not necessariy mean we're adding one byte to the address.
+It means we're adding `1` times the length of its type in bytes.
+
+But wait, there's more! You didn't think we were finished with `*`, did you?
+
+## Dereference Operator
```
wolf99, line 104: I thought the standard called it the "indirection operator".
But I don't have a copy right ATM and Dougle is not agreed, and now I'm second guessing myself!
It is often called dereference operator and it is probably clearer to the purpose for students, but at the same time it is good to use the right names

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+In the second line `my_ptr`, which was incremented to the second element in the first line, is first dereferenced, and then the dereferenced value is incremented.
+If we add this line to the above example
+
+```c
+// prints: my_ints[1] is 43. *++my_ptr is (some garbage value).
+printf("my_ints[1] is %u. *++my_ptr is %u.\n", my_ints[1], *++my_ptr);
+```
+
+In that line we have incremented the pointer to be beyond the array.
+We have overrun the array and we are pointing at the next four bytes in memory which the pointer interprets as an int, but, in fact, those bytes could belong to another variable or variables.
```
wolf99, line 170: Maybe it's only me being pedantic, but I would treat "int" as C code (the English translation being "integer"), just the same as you have done later on with "`char`s" (the English translation being "characters").

```suggestion
We have overrun the array and we are pointing at the next four bytes in memory which the pointer interprets as an `int`, but, in fact, those bytes could belong to another variable or variables.
```

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+    return 0;
+}
+```
+
+It may not be obvious, but the array of strings is declared and initialized as a pointer to an array of `char` arrays.
+When the array of `char` arrays is passed into `print_strings`, the signature describes it as a pointer to pointer(s) to `char`.
+
+So how do we know that `char **` is really a pointer to an array of pointers to char arrays, and not a pointer to a pointer of a single char?
+We don't.
+And this is what can make using pointers confusing.
```
wolf99, line 236:
```suggestion
This is one of the may things that can make using pointers confusing.
```
😄

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+may still compile and still print gibberish.
+But there is really no value to passing in a pointer to a pointer to a single char.
+Chances are, when you see `char **`, it represents a pointer to to an array of pointers to char arrays.
+
+We've seen a pointer passed into a function.
+But why pass a pointer into a function?
+
+## Passing a Pointer to a Function
+
+In C, arguments are passed to functions by value. That means that the value is copied from the argument and that copy is local to the function. When the function ends, so does the copied value.
```
wolf99, line 274:
```suggestion
In C, arguments are passed to functions by value. 
That means that the value is copied from the argument and that copy is local to the function. When the function ends, so does the copied value.
```

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+
+int main() {
+    char *my_strings[] = {"One string", "Two string"};
+    print_strings(my_strings);
+    return 0;
+}
+```
+
+It may not be obvious, but the array of strings is declared and initialized as a pointer to an array of `char` arrays.
+When the array of `char` arrays is passed into `print_strings`, the signature describes it as a pointer to pointer(s) to `char`.
```
wolf99, line 232:
```suggestion
When the array of `char` arrays is passed into `print_strings()`, the signature describes it as a pointer to pointer(s) to `char`.
```

Typically, when referring to a C function, people include the `()`.
I think this is only widely held convention rather than any rule we have though.
Perhaps @ryanplusplus could opine on whether it matters or not.

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+int main() {
+    int a = 3, b = 4;
+    swap(a, b);
+    // prints: After 'swap', a is 3 and b is 4
+    printf("After 'swap', a is %d and b is %d.\n", a, b);
+    return 0;
+}
+```
+
+This is because, even though the variables have the same names, `a` and `b` in `swap` are not the same `a` and `b` in `main`.
```
wolf99, line 298:
```suggestion
This is because, even though the variables have the same names, `a` and `b` in `swap()` are not the same `a` and `b` in `main()`.
```

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+    return 0;
+}
+```
+
+We are able to make a persistent change to `adjective`, demonstrating that the defined char array can be accepted as a pointer to char.
+
+Now let's try defining `adjective` as a pointer to `char` directly.
+
+```c
+// Online C compiler to run C program online
```
wolf99, line 377: Not sure if this comment line is relevant?

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+#include <string.h>
+
+void describe_me(char *adjective){
+    // segmentation fault on this line
+    strcpy(adjective, "evil");
+    printf("I am %s.\n", adjective);
+}
+
+int main() {
+    char *adjective= "good";
```
wolf99, line 388:
```suggestion
    char *adjective = "good";
```

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+    
+    return 0;
+}
+```
+
+Remember how pointers are powerful, but also potentially dangerous?
+But why doesn't it work?
+An array is a pointer and a pointer worked with a char array!
+The reason is that, before, `adjective` was defined as a `char` array which is a variable belonging to `main`.
+Now, `adjective` is defined as a string _literal_, meaning that it is written to read-only static memory which belongs to the entire program.
```
wolf99, line 400: Not exactly correct wording here I feel.
`adjective` is still just a pointer, the same as any other pointer.
What it _points to_ is a string literal.

wolf99:
Some small enough comments, looks good overall!

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+
+
+```c
+int i = 5;
+int *i_address = &i;
+```
+
+Wait, what does that `*` mean before `i_address`? Doesn't that mean multiply?
+
+A type followed by a `*` followed by an identifier is the way to declare a pointer to a variable of that type.
```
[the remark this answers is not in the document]
bobahop: This is why `*` can be confusing. Used as it is here, it is the way to declare a pointer. Indirection operator is dealt with elsewhere. I think to bring up the indirection operator in the middle of explaining how to define a pointer would be confusing.

concepts/pointers/about.md
bobahop: I'll commit it for now.

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+One question, though.
+Why do we put the star next to the identifier?
+Why not put it next to the type, since it's a pointer to that type?
+Because this
+
+```c
+int* my_ptr1, my_ptr2_that_is_really_an_int;
+```
+
+does not declare two pointers to an int, but declares a pointer to an int, and an int.
```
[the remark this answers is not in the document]
bobahop: I'll go through the whole thing to check for consistency on types. Thanks!

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+    print_strings(my_ptr_to_ptr);
+    return 0;
+}
+```
+
+but `print_strings` is not being used as intended.
+Using a different signature doesn't help. Changing the signature to this
+
+```c
+void print_strings(char *passed_in_strings[]) {
```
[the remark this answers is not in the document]
bobahop: Thanks! I'll take a look...

concepts/pointers/about.md
bobahop: That was sloppy of me. It's from the programiz online c compiler that I used for testing the code. I'll get rid of it. Thanks!

concepts/pointers/about.md
bobahop: Yes, that's technically correct. I elided talk about pointers to focus on the difference between local char arrays and string literals. Rather than do something like 

```
The reason is that, before, `adjective` was defined as a pointer to a `char` array which is a variable belonging to `main`.
Now, `adjective` is defined as a pointer to a string _literal_, meaning that it is written to read-only static memory which belongs to the entire program.
```

I'm thinking to do something like

```
The reason is that, before, `adjective` was defined as a `char` array which is a variable belonging to `main`.
Now, `adjective` is defined as a string _literal_, meaning that it is written to read-only static memory which belongs to the entire program. They are both pointers, but they point to different things.
```

But I suppose the first way isn't so bad, but, then again, maybe it's not so bad because I already understand it. For a novice, I'm not sure which way would be clearer to assimilate.

bobahop:
@wolf99 , thank you very much for you exhaustive critique! I already committed almost all of your suggestions. I'll go back through it and smooth out indirection/dereferencing and backticking the types consistently. There is also the wording of the "two pointers" for me to look at, and removing the online compiler comment. And whatever else I'm not thinking of right now.

It occurs to me that I didn't cover freeing pointers in this, so perhaps a small section on that should be added?

wolf99:
Thanks for adding the concepts, it's great to get any contribution!

On freeing pointers; Im not sure what you have in mind exactly but maybe that's more like freeing dynamically allocated memory than freeing a pointer? Although you could possibly explain something about setting pointers to NULL is the equivalent of "un-setting" a pointer, as it were;
 it may be best to avoid mixing dynamic memory handling in here if at all possible.

bobahop:
@wolf99   I think you're right. That's probably why I didn't add that to begin with.

concepts/pointers/about.md
bobahop: Fixed in latest commit

concepts/pointers/about.md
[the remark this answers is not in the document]
bobahop: Added closing brace in latest commit

concepts/pointers/about.md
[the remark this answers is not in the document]
bobahop: Backticked in latest commit.

concepts/pointers/about.md
bobahop: Dealt with this in several places in latest commit. Please let me know if anything slipped through or could be better worded.

concepts/pointers/about.md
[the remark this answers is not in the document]
bobahop: I added some about indirection operator in the latest commit. Please let me know if it could be improved.

concepts/pointers/about.md
bobahop: Well, I at least fixed the spelling for "dereferencing." 😄

concepts/pointers/about.md
```diff
@@ -0,0 +1,404 @@
...
+    
+    // since my_ints is already an address
+    // we don't need the address operator to make a pointer from it.
+    // a pointer to my_ints holds the address of my_ints
+    int *my_ptr = my_ints;
+    
+    // the address of my_ints and the value of my_ptr are the same
+    printf("my_ints is %p. my_ptr is %p.\n", my_ints, my_ptr);
+    
+    // both addresses of the first element are the same as the original addresses
```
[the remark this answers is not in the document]
bobahop: I tweaked the comments in the latest commit. Please let me know if it could be improved.

bobahop:
Latest commits failed "Fail with link errors". Didn't use hyperlinks in the text, so I don't know what this is referring to.

concepts/pointers/about.md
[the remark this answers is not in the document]
wolf99: I think it is still the indirection operator even though it is used for declaration purposes.
A minor point anyway and can be left as is if you feel it is clearer.

concepts/pointers/about.md
```diff
@@ -88,8 +90,8 @@ int main() {
 
 The previous example shows why array indexing begins at `0` instead of `1`.
 The index into an array multiplies the index by the size of the type and adds it to the starting address of the array.
-Since the first element aligns with the beginning of the array, we multiply the type by zero for adding to the start address to get the first element.
-To get the second element we multiple the type by 1 for adding to the start address.
+Since the first element aligns with the beginning of the array, we multiply the type by `0` for adding to the start address to get the first element.
+To get the second element we multiple the type by `1` for adding to the start address.
```
wolf99, line 94: Apologies, I missed this first time around.

```suggestion
To get the second element we multiply the type by `1` for adding to the start address.
```

concepts/pointers/about.md
```diff
@@ -99,20 +101,22 @@ It means we're adding `1` times the length of its type in bytes.
...
 
-## Dereference Operator
+## Indirection Operator (part 2)
 
 Now that we understand how addresses work with array indexing, what good is it if we don't get the value stored at an address?
 How do we do that?
-To get the value stored at an address we use the dereference operator, which is, you guessed it, `*`.
+To get the value stored at an address we use the indirection operator, which, you may recall, is `*`.
+In this situation the indirection operator is also sometimes referred to as the dereference operator, because it dereferences the value for which the pointer has the address.
+Calling it the dereference operator when used for dereferencing distinguishes it from the indirection operator for defining a pointer, but `*` for both uses is technically named the indirection operator.
```
wolf99, line 110: Great! 👍

concepts/pointers/about.md
wolf99: "Derefencing": The taking down of a refurbished fence. 😄

concepts/pointers/about.md
bobahop: I've looked at that so many times and have missed it every time!

concepts/pointers/about.md
```diff
@@ -229,7 +233,7 @@ int main() {
 It may not be obvious, but the array of strings is declared and initialized as a pointer to an array of `char` arrays.
 When the array of `char` arrays is passed into `print_strings()`, the signature describes it as a pointer to pointer(s) to `char`.
 
-So how do we know that `char **` is really a pointer to an array of pointers to char arrays, and not a pointer to a pointer of a single char?
+So how do we know that `char **` is really a pointer to an array of pointers to `char` arrays, and not a pointer to a pointer of a single `char`?
 We don't.
 This is one of the may things that can make using pointers confusing.
```
wolf99, line 238: Maybe you intended these to render on new lines, as in other cases of this style of prose?
If so, you'll need a couple extra new lines else the markdown is still rendered on all one line IIRC.
```suggestion
So how do we know that `char **` is really a pointer to an array of pointers to `char` arrays, and not a pointer to a pointer of a single `char`?

We don't.

This is one of the may things that can make using pointers confusing.
```

concepts/pointers/about.md
wolf99: I think you've got a nice balance with it now, well done!

wolf99:
The links seem to be failing due to an unrelated link that refers to the website of the test framework that we use.
Looking at that link my browser tells me the website's TLS certificate is not valid, so it looks like a "Them" problem rather than an "Us" problem.
I'll raise an issue to fix this separately. 
So for the purposes of this PR we can ignore it as an [SEP](https://hitchhikers.fandom.com/wiki/Somebody_Else%27s_Problem_Field).

wolf99:
This looks good to me @bobahop .
If you'd like to move it from a draft to "Ready for Review" I can approve.

wolf99:
Just had a thought: what about `restrict` ?

bobahop:
I'm not very familiar with `restrict`. I've heard of it. I think I've seen it in one solution, but haven't gotten around to assimilating it.

wolf99:
Hey Bob, 
As a solution to get this merged I suggest we go ahead as-is, but if you could open an issue to add `restrict` to this concept so that you or someone else could pick it up separately.
When creating that issue it would be good to link this PR in it also, to keep the context.

wolf99 (approved):

wolf99:
Woo, thanks Bob! 🚀

PR #743, opened 2022-01-17
all exercises should include headers

joshgoebel:


joshgoebel:
I did remove the enumerable definitions for the resistor exercises in case that was giving away a bit too much.

wolf99:
I think we previously discussed if we should add Stubb's and decided only to have them for the earlier exercises.

Open to discuss the idea again, but there should be some opinions here before this is merged I feel.

joshgoebel:
> I think we previously discussed if we should add Stubb's and decided only to have them for the earlier exercises.

But that's NOT the case at all, it's *all over the map*.  Being consistent is one thing, but it's just random.  Quite a few early exercises are missing them, a lot of hard exercises have them...

So if there are criteria lets decide and then perhaps document that in the README for future reference - and also then fix any remaining inconsistencies between what we have now and what the actual policy is going to be.

ryanplusplus:
> > I think we previously discussed if we should add Stubb's and decided only to have them for the earlier exercises.
> 
> But that's NOT the case at all, it's _all over the map_. Being consistent is one thing, but it's just random. Quite a few early exercises are missing them, a lot of hard exercises have them... (at least I could discern no pattern)
> 
> So if there are criteria lets decide and then perhaps document that in the README for future reference - and also then fix any remaining inconsistencies between what we have now and what the actual policy is going to be.

What's the policy for this PR? Does this just add prototypes to all headers that were missing them or is it more nuanced?

wolf99:
I realised this is "all over the map" because the track's curriculum changed from branched to linear with the change to v3.
Which means if we would like to have the previously agreed approach to stubs, then this is a valid issue and the exercises need updating inline with the new curriculum progression.

jsannemo:
As someone who just tried exercism out to refresh syntax a bit, I find the exercises without stubs super annoying. I much prefer solving problems over having to read through all the tests just to first figure out what functions I'm even supposed to implement.

dm-alexi:
One should be able to start the exercise without having to peek into the test section. Therefore, all exercises should include headers.

wolf99:
> One should be able to start the exercise without having to peek into the test section. Therefore, all exercises should include headers.

Hi @dm-alexi
This track (and others) use TDD (test driven development) as a means to teach the interface expected. In TDD students are _supposed_ to look in the test files. This is a widespread development paradigm and the use of it on Exercise generally teaches this paradigm but also allows students the opportunity to learn to create the header files (a key part of C development) instead of just the implementation.
I thought this was explained somewhere on the Exercise site somewhere when starting the track but perhaps it has been re/moved

jsannemo:
Nobody doing TDD is handed a bunch of tests where someone else have already decided on the interface and reverse engineer it from those, so I don't really buy that this somehow teaches TDD more than offering up stubs.

Of course, I can only offer up my own experience for people to decide on whether this tradeoff ("learning to write header files") is worth it: I stopped doing exercism pretty quickly because it was tiresome reversing types and function signatures all the time.

vladimir-popov:
Personally, I think it would be better to see in the header file all types used in the exercise, than recover them from the source. The empty header file looks like accidentally pushed without commits.

wolf99:
Tha ks for adding your thoughts @vladimir-popov.

To help with this specific example we could either:

1. Add declarations to headers
2. Explain the lack of declarations in the track website documentaion

I guess those already on this issue will advocate for 1, but any thoughts on 2?

ryanplusplus:
@wolf99 my thought is that the ideal state is we have a mix of exercises with and without headers with the exercises without headers coming later in the track. Before we do this we probably need to work on the exercise ordering since it's at least partly accidental right now. Thoughts?

wolf99:
I'm ok with this @ryanplusplus .
Though honestly I can't recall how ordering is done!

ryanplusplus:
> Though honestly I can't recall how ordering is done!

Same 😞

vladimir-popov:
@wolf99 could you share an example of the exercise, where reverse engineering of the header file is reasonable, please. I think I don't get your point well.

PR #744, opened 2022-01-19
color tests

joshgoebel:
- enh: tests show results in color

Adds `-DUNITY_OUTPUT_COLOR` to all makefiles

### DIFF
--- a/exercises/practice/acronym/makefile
+++ b/exercises/practice/acronym/makefile
@@ -11,7 +11,7 @@ CFLAGS += -Wextra
 CFLAGS += -pedantic
 CFLAGS += -Werror
 CFLAGS += -Wmissing-declarations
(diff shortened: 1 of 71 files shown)


joshgoebel:
Test runner may need updates also?

wolf99 (approved):
LGTM

Test runner may need update, I can't recall. It's in its own c-test-runner repo.

ryanplusplus (approved):

siebenschlaefer:
Are you sure that doesn't break the current test runner?

There's a (IMHO correct) solution in the queue where all tests pass but it's marked as "Failed".
I just tried one of my old exercises and get the same behavior: All tests pass but I can't submit because the test runner thinks "An error occurred".

joshgoebel:
The test runner may need to be updated to use `sed` to remove the new define to get black and white.

PR #747, opened 2022-01-27
free leaked memory in test_colors function.

Abdellahmaarifa:
calling the colors() function should return an array allocated in the heap, after done testing, it should be freed.

### DIFF
--- a/exercises/practice/resistor-color/test_resistor_color.c
+++ b/exercises/practice/resistor-color/test_resistor_color.c
@@ -1,5 +1,6 @@
 #include "test-framework/unity.h"
 #include "resistor_color.h"
+#include <stdlib.h>
 
 #define ARRAY_LENGTH(A) (sizeof(A) / sizeof(A[0]))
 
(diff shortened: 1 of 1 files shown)


ryanplusplus:
The result of `colors()` is fixed so it's not necessary to dynamically allocate the returned value. The way that the test works now requires this behavior. Your proposal is also valid, but I don't think it's an improvement over the existing implementation because the existing implementation demonstrates that not all pointer values returned have to be dynamically allocated.

wolf99:
IIRC, this was implemented in the current fashion to demonstrate precisely as @ryanplusplus indicates.

There are a lot of other exercises that use dynamic allocation, so it is nice to have some that demonstrate that dynamic allocation is not a hard requirement for any C that handles array passing.

Is there somewhere appropriate that we could document such, per-exercise, decisions and intentions? @ErikSchierboom

ErikSchierboom:
@wolf99 I would use the `.meta/design.md` for that. We currently only use it for Concept Exercises, but we could just as easily use it for Practice Exercises.

wolf99:
Closing as existing design is by intention.

PR #749, opened 2022-01-31
Update test_say.c  Fixed memory leaks

bobahop:
`ans` would not be freed on failed `TEST_ASSERT_XXX`, according to `make memcheck` on my local machine. Used `test_phone_number.c` as my guide for the fix.

### DIFF
--- a/exercises/practice/say/test_say.c
+++ b/exercises/practice/say/test_say.c
@@ -12,7 +12,7 @@ void tearDown(void)
 {
    if (ans)
       free(ans);
-   ans = NULL;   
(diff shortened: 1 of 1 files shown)


bobahop:
Fixed formatting error.

wolf99:
Slightly differnt pattern used here, but same question remains as for #750 , why does function scope present a memory leakage, but file scope does not?

bobahop:
File scope would leak memory if the memory wasn't freed in the teardown.

ryanplusplus (approved):

PR #750, opened 2022-01-31
Update to test_roman_numerals.c Fixed memory leaks

bobahop:
In current configuration `make memcheck` reported direct memory leak for failed `TEST_ASSERT_EQUAL_STRING` on my local machine (Ubuntu WSL.) With the updated configuration there are no memory leaks. Specifying `if (result) free result;` generated an error for attempted double free of `result`, so it must be getting successfully freed somewhere as long as the ASSERT passes. **Edit** I had forgotten to set `result` to NULL, which seems to have led to the double-free.

### DIFF
--- a/exercises/practice/roman-numerals/test_roman_numerals.c
+++ b/exercises/practice/roman-numerals/test_roman_numerals.c
@@ -9,6 +9,9 @@ void setUp(void)
 
 void tearDown(void)
 {
+   if (result)
+      free(result);
+   result = NULL;
 }
 
 static void check_conversion(int number, char *expected)
--- a/exercises/practice/roman-numerals/test_roman_numerals.c
+++ b/exercises/practice/roman-numerals/test_roman_numerals.c
@@ -2,6 +2,7 @@
 #include "roman_numerals.h"
 #include <stdlib.h>
 
+static char *result = NULL;
 void setUp(void)
 {
 }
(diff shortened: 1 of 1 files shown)


bobahop:
This isn't ready yet. When testing with more than one test it leaks again. So looking into it...

bobahop:
Is working now for multiple tests.

wolf99 (approved):

wolf99:
I'm not clear why this would be a memory leak.
As `result` is a local variable which is created and freed each time the function is called, where is it leaked?

wolf99 (changes requested):
Sorry, I was too hasty with approval. I would like to confirm my understanding with you first per the previous comment on the PR

ryanplusplus:
> I'm not clear why this would be a memory leak. As `result` is a local variable which is created and freed each time the function is called, where is it leaked?

I'll let @bobahop respond, but I suspect it's because Unity calls `teardown()` even after a failed test whereas the body of the test after the failure is skipped. If so, we could also fix this by freeing before performing the check (though that would be less elegant and would prevent us from using the helpful check macros).

bobahop:
From the behavior I observed, if the assert fails then no more code in that function is run, thus the free is not reached. Result is not exactly a local variable. It must be allocated memory on the heap in the program being tested to be able to return a value.

bobahop:
What I've changed to is essentially the same code that is already in phone number, which reports no memory errors on failed assertions. I don't see any reason to free before the check; just to make sure we free in the teardown, which is what it's there for, I think.

ryanplusplus:
> I don't see any reason to free before the check; just to make sure we free in the teardown, which is what it's there for, I think.

Agreed. I don't think we need to do it, I was just using that example to explain the behavior.

ryanplusplus (approved):

wolf99 (approved):

wolf99:
Ah, of course! 
Great, all good then. Thanks for these fixes @bobahop ! 👌

PR #751, opened 2022-01-31
Update test_pig_latin.cs Fixed memory leaks

bobahop:
`make memcheck` reported direct memory leak on failed `TEST_ASSERT_EQUAL_STRING` on my local machine (Ubuntu WSL.) Now reports no memory leaks.

### DIFF
--- a/exercises/practice/pig-latin/test_pig_latin.c
+++ b/exercises/practice/pig-latin/test_pig_latin.c
@@ -14,180 +14,142 @@ void tearDown(void)
    res = NULL;
 }
 
+static void check_transate(const char *phrase, const char *expected)
+{
+   res = translate(phrase);
+   TEST_ASSERT_EQUAL_STRING(res, expected);
+}
+
 static void test_word_beginning_with_a(void)
 {
-   const char *phrase = "apple";
-   res = translate(phrase);
-   TEST_ASSERT_EQUAL_STRING("appleay", res);
+   check_transate("apple", "appleay");
 }
 
 static void test_word_beginning_with_e(void)
 {
-   TEST_IGNORE();   // delete this line to run test
-   const char *phrase = "ear";
-   res = translate(phrase);
-   TEST_ASSERT_EQUAL_STRING("earay", res);
+   TEST_IGNORE(); // delete this line to run test
+   check_transate("ear", "earay");
 }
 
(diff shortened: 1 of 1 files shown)


exercises/practice/pig-latin/test_pig_latin.c
```diff
@@ -2,210 +2,192 @@
...
+static char *res = NULL;
 void setUp(void)
 {
 }
 
 void tearDown(void)
 {
+   if (res)
+      free(res);
+   res = NULL;
```
wolf99, line 14: This seems like a bit of a tautology here, in that `res` is set to `NULL` even if it already is.
Would it make more sense, semantically, to set `res` to `NULL` in the `setup()` ?

exercises/practice/pig-latin/test_pig_latin.c
```diff
@@ -2,210 +2,192 @@
...
 }
 
 static void test_word_beginning_with_a(void)
 {
    const char *phrase = "apple";
-   char *res = translate(phrase);
+   res = translate(phrase);
    TEST_ASSERT_EQUAL_STRING("appleay", res);
```
wolf99, line 21: Potentially these lines could be further refactored out of each `test_` function to a  single `check_translation()` function.
Do you think this PR could cover that?

exercises/practice/pig-latin/test_pig_latin.c
bobahop: I based it on phone number. And I think it makes sense to do it here as part of a thorough cleanup. See [here](https://stackoverflow.com/questions/1025589/setting-variable-to-null-after-free#:~:text=free()%20is%20a%20library,will%20be%20set%20to%20NULL.)

exercises/practice/pig-latin/test_pig_latin.c
bobahop: Yes, I was thinking about that as I saw that done in another one of the leaking exercises. I could use that one as a template for this one and refactor. I may not get to it until tomorrow, but thanks for the go-ahead!

bobahop: pig latin now revised from roman numeral template.

wolf99 (approved):

PR #755, opened 2022-02-05
Update tests_pascals_triangle.c:  Fix memory leaks

bobahop:
Very similar to previous fixes on other tracks. A difference is that the student allocates an array of arrays and is responsible for freeing all of that. Failing test ASSERTS were leaking memory.  Now failed ASSERTS do not leak memory.  No double-free on success. Of course, if a student allocates something else we don't expect, we can't catch it and the leak is on them, so to speak. 😄

### DIFF
--- a/exercises/practice/pascals-triangle/test_pascals_triangle.c
+++ b/exercises/practice/pascals-triangle/test_pascals_triangle.c
@@ -1,8 +1,12 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stddef.h>
+#include <stdlib.h>
 #include "test-framework/unity.h"
 #include "pascals_triangle.h"
+
+uint8_t **actual = NULL;
+size_t count = 0;
 
 void setUp(void)
 {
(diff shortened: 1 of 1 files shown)


exercises/practice/pascals-triangle/test_pascals_triangle.c
```diff
@@ -29,18 +43,20 @@ static bool check_triangle(size_t count, uint8_t expected[][count],
 static void test_no_rows(void)
 {
    uint8_t expected[1][1] = { { 0 } };
-   uint8_t **actual = create_triangle(0);
-   TEST_ASSERT_TRUE(check_triangle(1, expected, actual));
+   actual = create_triangle(0);
+   TEST_ASSERT_TRUE(check_triangle((count = 1), expected, actual));
    free_triangle(actual, 1);
+   actual = NULL;
```
ryanplusplus, line 49: I think this is redundant because it's handled in the teardown. Is this included here to make it clear what the "normal" flow looks like?

bobahop: Thank you very much for looking at this!

The student's code in  `free_triangle` doesn't necessarily set `actual` to NULL. So we need to set it to NULL so we don't get a double free in the teardown. Luckily, there is no error for double setting to NULL like there is for double free. Of course, if the student does not free the memory in `free_triangle` but sets it to NULL, there's nothing we can do about that. In that case `make memcheck` would rightly report a memory leak. But from the few exercises I spot-checked, students use free but they don't set to NULL.

bobahop: Actually, before it even gets to double free, if the rows are freed by the student's `free_triangle` after successfully passing the test, trying to index into them will generate a `heap-use-after-free`. But that won't happen if we set `actual` to NULL.

ryanplusplus (approved):

PR #756, opened 2022-02-07
Update test_list_ops.c: Fixed memory leaks

bobahop:
Was leaking memory on failed ASSERTS. Students can't set lists persistently to NULL, since the delete_list function is only given the pointer to a list, not the address of the pointer, so if delete_list set the pointer to NULL it would only be setting the copy of the pointer to NULL. So now we set the list pointer to NULL in a variadic function that is passed all the lists needing to be deleted, which makes tests with multiple lists a bit less verbose. So, the student frees the list, we set it to NULL, and there is no memory leak on fail and no double-free on success.

### DIFF
--- a/exercises/practice/list-ops/test_list_ops.c
+++ b/exercises/practice/list-ops/test_list_ops.c
@@ -12,35 +12,15 @@ void setUp(void)
 {
 }
 
-static void call_delete_list(int count, ...)
-{
-   va_list ptr;
-   va_start(ptr, count);
-   for (int i = 0; i < count; i++) {
-      list_t **list = va_arg(ptr, list_t **);
-      delete_list(*list);
-      *list = NULL;
-   }
-   va_end(ptr);
-}
-
 static void release(int count, ...)
 {
    va_list ptr;
    va_start(ptr, count);
    for (int i = 0; i < count; i++) {
-      if (i < 3) {
-         list_t **list = va_arg(ptr, list_t **);
-         if (*list) {
-            free(*list);
-            *list = NULL;
-         }
-      } else {
-         char **err_msg = va_arg(ptr, char **);
-         if (*err_msg) {
-            free(*err_msg);
-            *err_msg = NULL;
-         }
+      list_t **list = va_arg(ptr, list_t **);
+      if (*list) {
+         free(*list);
+         *list = NULL;
       }
    }
    va_end(ptr);
@@ -48,7 +28,11 @@ static void release(int count, ...)
 
 void tearDown(void)
 {
-   release(4, &list, &list2, &actual, &error_message);
+   release(3, &list, &list2, &actual);
+   if (error_message) {
+      free(error_message);
+      error_message = NULL;
+   }
 }
 
 static char *print_elements(size_t length, list_element_t list_elements[])
@@ -131,6 +115,12 @@ static list_element_t fold_divide(list_element_t element,
    return (accumulator == 0) ? 0 : element / accumulator;
 }
 
+static void call_delete_list(list_t **list)
+{
+   delete_list(*list);
+   *list = NULL;
+}
+
 static void test_append_empty_lists(void)
 {
    list = new_list(0, NULL);
@@ -139,7 +129,9 @@ static void test_append_empty_lists(void)
    actual = append_list(list, list2);
    check_lists_match(0, NULL, actual);
 
-   call_delete_list(3, &list, &list2, &actual);
+   call_delete_list(&list);
+   call_delete_list(&list2);
+   call_delete_list(&actual);
 }
 
 static void test_append_list_to_empty_list(void)
@@ -153,7 +145,9 @@ static void test_append_list_to_empty_list(void)
    actual = append_list(list, list2);
    check_lists_match(expected_length, expected_elements, actual);
 
-   call_delete_list(3, &list, &list2, &actual);
+   call_delete_list(&list);
+   call_delete_list(&list2);
+   call_delete_list(&actual);
 }
 
 static void test_append_empty_list_to_list(void)
@@ -167,7 +161,9 @@ static void test_append_empty_list_to_list(void)
    actual = append_list(list, list2);
    check_lists_match(expected_length, expected_elements, actual);
 
-   call_delete_list(3, &list, &list2, &actual);
+   call_delete_list(&list);
+   call_delete_list(&list2);
+   call_delete_list(&actual);
 }
 
 static void test_append_non_empty_lists(void)
@@ -181,7 +177,9 @@ static void test_append_non_empty_lists(void)
    actual = append_list(list, list2);
    check_lists_match(expected_length, expected_elements, actual);
 
-   call_delete_list(3, &list, &list2, &actual);
+   call_delete_list(&list);
+   call_delete_list(&list2);
+   call_delete_list(&actual);
 }
 
 static void test_filter_empty_list(void)
@@ -192,7 +190,8 @@ static void test_filter_empty_list(void)
    actual = filter_list(list, filter_modulo);
    check_lists_match(0, NULL, actual);
 
-   call_delete_list(2, &list, &actual);
+   call_delete_list(&list);
+   call_delete_list(&actual);
 }
 
 static void test_filter_non_empty_list(void)
@@ -205,7 +204,8 @@ static void test_filter_non_empty_list(void)
    actual = filter_list(list, filter_modulo);
    check_lists_match(expected_length, expected_elements, actual);
 
-   call_delete_list(2, &list, &actual);
+   call_delete_list(&list);
+   call_delete_list(&actual);
 }
 
 static void test_length_empty_list(void)
@@ -217,7 +217,7 @@ static void test_length_empty_list(void)
    size_t actual = length_list(list);
    TEST_ASSERT_EQUAL(expected, actual);
 
-   call_delete_list(1, &list);
+   call_delete_list(&list);
 }
 
 static void test_length_non_empty_list(void)
@@ -229,7 +229,7 @@ static void test_length_non_empty_list(void)
    size_t actual = length_list(list);
    TEST_ASSERT_EQUAL(expected, actual);
 
-   call_delete_list(1, &list);
+   call_delete_list(&list);
 }
 
 static void test_map_empty_list(void)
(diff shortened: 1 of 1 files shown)


exercises/practice/list-ops/test_list_ops.c
```diff
@@ -1,14 +1,54 @@
-#include "test-framework/unity.h"
+#include <stdarg.h>
 #include "list_ops.h"
+#include "test-framework/unity.h"
 
 const int MAX_STRING_LEN = 100;
+list_t *list = NULL;
+list_t *list2 = NULL;
+list_t *actual = NULL;
+char *error_message = NULL;
```
ryanplusplus, line 9:
```suggestion
static const int MAX_STRING_LEN = 100;
static list_t *list = NULL;
static list_t *list2 = NULL;
static list_t *actual = NULL;
static char *error_message = NULL;
```

exercises/practice/list-ops/test_list_ops.c
```diff
@@ -1,14 +1,54 @@
...
+            free(*list);
+            *list = NULL;
+         }
+      } else {
+         char **err_msg = va_arg(ptr, char **);
+         if (*err_msg) {
+            free(*err_msg);
+            *err_msg = NULL;
+         }
+      }
```
ryanplusplus, line 44: This seems kind of convoluted. Why have a variadic function that's only called once and that has one argument handled in a special way?

exercises/practice/list-ops/test_list_ops.c
```diff
@@ -92,248 +133,234 @@ static list_element_t fold_divide(list_element_t element,
...
+   list2 = new_list(0, NULL);
 
-   list_t *actual = append_list(list1, list2);
+   actual = append_list(list, list2);
    check_lists_match(0, NULL, actual);
 
-   delete_list(list1);
-   delete_list(list2);
-   delete_list(actual);
+   call_delete_list(3, &list, &list2, &actual);
```
ryanplusplus, line 142: If I'm a student learning C and I see this, I'm not sure that it will be clear to me what's happening. This is less verbose, but for a student is this an improvement over deleting each of these and setting to `NULL`?

ryanplusplus:
I understand why it is preferable to not leak during a failure, but I think this makes the tests significantly harder to understand for a student. I'm concerned that the extra complexity and obfuscation in the test suite may have a much higher cost than having a memory leak in a failed test case (and _only_ in a failed test case).

@wolf99 what do you think?

exercises/practice/list-ops/test_list_ops.c
bobahop: Some tests only have one list to delete; some have up to three. Also, the error_msg can leak memory. So we can have this

```c
void tearDown(void)
{
   release(4, &list, &list2, &actual, &error_message);
}
```

Instead of this

```c
static void release(list_t **list)
{
   if (*list)
   {
      free(*list);
      *list = NULL;
   }
}

void tearDown(void)
{
   release(&list);
   release(&list2);
   release(&actual);
   if (error_message)
   {
      free(error_message);
      error_message = NULL;
   }
}
```

But if you prefer the latter I can change it back.

exercises/practice/list-ops/test_list_ops.c
bobahop: That the students read the tests to learn C vs. just to get the function signatures is something I hadn't considered. I can say, as a student of C myself, when I saw the three commands in a row that only differ by the argument, my first thought was "why not send them all at once into a single function that can loop through them?" And that set me onto a variadic function. If not a variadic function, I would think to at least send them into a function that would call `delete_lst` and set to NULL instead of doing it inline over and over. Then again, if they are learning C from reading the tests, then this is how they can learn about variadic functions.

exercises/practice/list-ops/test_list_ops.c
ryanplusplus: The part of this that feels clumsiest isn't the use of a variadic function, but the fact that one of the arguments is special in a non-obvious way. What is gained by including the message as one of the variadic arguments instead of as a positional parameter?

exercises/practice/list-ops/test_list_ops.c
ryanplusplus: I wasn't approaching this from the perspective of using the tests as an opportunity for a student to learn more C. My concern is instead that the student should be able to read the test to understand what is being tested and not have to learn, for example, what a variadic function is in order to reach the learning objective of the exercise. The test code should not get in the way of the student completing the exercise.

I can speak from my experience in other tracks on this. Many tracks have "clever" tests that are structured to be as DRY as possible in order to make them easier to maintain. But, as a student, this often makes it unclear what is even being tested. There may be an argument for using this approach on "real" tests (ie: tests for production code), but as a learner that is still coming to grips with the language, this almost invariably serves to obfuscate. Instead of being able to focus on the learning objectives of the exercise, I also have to wade through the complexity of the tests to understand what is expected of my code.

bobahop:
```javascript
delete_list(list1);
list1 = NULL;
delete_list(list2);
list2 = NULL;
delete_list(actual);
actual = NULL;
```

doesn't seem [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) to me, but it may be simpler for a novice to follow.

As it is now, if students develop their solutions in the browser, they won't see the memory leaks happening, but if they develop locally they will see memory leaks if they run `make memcheck`. I don't know how many of them do that instead of running `make test`. I know I do, and the first time I saw memory leaks for a failed test when I thought I had handled it properly, it was confusing, until I realized it was the test itself which was leaking memory. So I think it should be fixed, even if we simplify the fix. I think it can be embarrassing to Exercism if we don't.

bobahop: Our last comments crossed. I can revert to the simpler fixes and remove the variadics..

exercises/practice/list-ops/test_list_ops.c
bobahop: The variadic function as it is now has all cleanup in one function vs. distributing it between itself and teardown. We could have the variadic take just the lists and still do `error_message` separately, like so

```c
static void release(int count, ...)
{
   va_list ptr;
   va_start(ptr, count);
   for (int i = 0; i < count; i++) {
      list_t **list = va_arg(ptr, list_t **);
      if (*list) {
         free(*list);
         *list = NULL;
      }
   }
   va_end(ptr);
}

void tearDown(void)
{
   release(3, &list, &list2, &actual);
   if (error_message) {
      free(error_message);
      error_message = NULL;
   }
}
```

I can implement whichever way you prefer.

exercises/practice/list-ops/test_list_ops.c
bobahop: Does this seem okay, or do you think it's still too complex?

```c
static void call_delete_list(list_t **list)
{
   delete_list(*list);
   *list = NULL;
}
// code snipped
static void test_append_empty_lists(void)
{
   list = new_list(0, NULL);
   list2 = new_list(0, NULL);

   actual = append_list(list, list2);
   check_lists_match(0, NULL, actual);

   call_delete_list(&list);
   call_delete_list(&list2);
   call_delete_list(&actual);
}
```

bobahop: By "too complex" I mean is the indirection too confusing for a novice. To your point, I remember a particular exercise on another track which generated some complaints right after the v3 launch. The implementation was empty and a few students found it too hard to figure out what was required from the tests.

ryanplusplus: > So I think it should be fixed, even if we simplify the fix. I think it can be embarrassing to Exercism if we don't.

I agree. My point is just that we need to make sure that the fix isn't worse for a student than the issue we're fixing. Your last proposal looks good to me. It's cleaner without relying upon anything that will look like black magic to a beginner.

exercises/practice/list-ops/test_list_ops.c
ryanplusplus: I think your last proposal is a good way to go.

exercises/practice/list-ops/test_list_ops.c
bobahop: So, to make sure I understand what my last proposal is, the teardown will be this

```c
void tearDown(void)
{
   release(3, &list, &list2, &actual);
   if (error_message) {
      free(error_message);
      error_message = NULL;
   }
}
```

and the tests will be something like this

```c
   call_delete_list(&list);
   call_delete_list(&list2);
   call_delete_list(&actual);
```

Sorry if I got it wrong. I should know what my last proposal is, but since they're spit between two threads, I've befuddled myself!

exercises/practice/list-ops/test_list_ops.c
```diff
@@ -1,14 +1,38 @@
...
+static list_t *list = NULL;
+static list_t *list2 = NULL;
+static list_t *actual = NULL;
+static char *error_message = NULL;
 
 void setUp(void)
 {
 }
 
+static void release(int count, ...)
```
wolf99, line 15: Could this be `release_lists` so as to be a little clearer on the purpose?

exercises/practice/list-ops/test_list_ops.c
wolf99: As the `tearDown()` now does the cleaning up, and `delete_list()` is not an explicit part of the spec, do we even need to delete the lists within the test functions?
They could be removed completely and just let `tearDown()` handle it in both the success case and the error case.

wolf99:
This was my thinking also @ryanplusplus.
But I think we are now close to a pretty clear implementation 🙂

wolf99:
Closes #756

exercises/practice/list-ops/test_list_ops.c
```diff
@@ -1,14 +1,38 @@
...
+{
+   va_list ptr;
+   va_start(ptr, count);
+   for (int i = 0; i < count; i++) {
```
wolf99, line 19:
```suggestion
   for (size_t i = 0; i < count; i++) {
```

wolf99: FYI also, that `++i` rather than `i++` would be ever so slightly more efficient.

wolf99 (approved):
LGTM

bobahop:
Thank you for your help with this one. Preserving the clarity of the tests is a lesson I intend to keep in mind when moving forward.

ryanplusplus:
> Thank you for your help with this one. Preserving the clarity of the tests is a lesson I intend to keep in mind when moving forward.

It's all good. It's a bit of a unique scenario because you have to do something different than what you'd do for production code.

PR #762, opened 2022-02-22
CI for concepts

wolf99:
I realised during #735  that tooling and CI does not yet cover concept exercises.
This PR should resolve that.

### DIFF
--- a/bin/check-include-guards
+++ b/bin/check-include-guards
@@ -2,11 +2,17 @@
 
 set +e
 
+usage="check-include-guards - checks include guards for all exercises in a given directory
+Usage:
+  check-include-guards <path-to-exercises>
+
+Example:
+  check-include-guards ./exercises/concept ./exercises/practice"
+
 ERROR_COUNT=0
 
-echo "Checking include guards format"
-
 check_guard() {
+    # Check the include guards in a given file
     if [ -f "$1" ]; then
         if grep -LPzo "$2" "$1" > /dev/null; then
             printf "%s does not have correctly formatted include guards\n" "$1"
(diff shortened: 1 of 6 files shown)


bin/check-unitybegin
```diff
@@ -9,11 +16,35 @@ set +e
...
+check_exercises () {
+    for f in "${1}"/*/test_*; do
+        if [ -f "${f}" ]
+        then
+            if ! grep -q UnityBegin\(\""$(basename "$f")"\"\) "$f"; then
+                echo "$f needs correct UnityBegin line"
+                (( ERROR_COUNT++ ))
+            fi
+        else
+            echo "WARNING: file not found. Ignoring ${1}"
```
wolf99, line 28: Im not too sure if it is a good idea to leave this as a warning.
Maybe it should be an error.
It could mean that there is a exercises/concept directory but no concept exercises therein, but it could also mean that there is a concept exercise without the needed files within it.
But then - separation of concerns - should this script care about that?

bin/verify-unity-version
```diff
@@ -27,11 +34,37 @@ test_unity_version() {
...
+check_exercises () {
+    for f in "${1}"/*/test-framework/unity.h; do
+        if [ -f "${f}" ]
+        then
+            if ! test_unity_version "$f"; then
+                echo "$f unexpected unity version"
+                (( ERROR_COUNT++ ))
+            fi
+        else
+            echo "WARNING: file not found. Ignoring ${1}"
```
wolf99, line 47: Same question about the error handling here.

wolf99:
Please squash if merging!

bin/check-include-guards
```diff
@@ -15,21 +21,43 @@ check_guard() {
...
+then
+    echo "Checking include guards format"
+
+    for directory
+    do
+        if [ -d "${directory}" ]
+        then
+            check_exercises "${directory}"
+        else
+            echo "WARNING: Directory not found. Ignoring: ${1}"
```
ryanplusplus, line 52: How about making this an error? I can't think of a situation where we wouldn't want to make this obvious.

bin/check-unitybegin
ryanplusplus: I'm in favor of making this an error. It's likely that it would get caught long before this, but I can't see any reason why we wouldn't take this opportunity to call attention to something unexpected.

ryanplusplus:
This looks good. I'm happy to merge this once we've resolved the question about warnings and errors.

bin/check-include-guards
wolf99: The exercises/concept directory doesn't currently exist. So this would cause the CI to always fail until we get some concept exercises merged

We could maybe not pass that directory in yet, but then at least the first concept exercise will not have CI.
It's kind of a bootstrap problem.

ryanplusplus: My bad, I haven't been following the Concept exercises and didn't realize that none had been merged.

ryanplusplus (approved):

PR #763, opened 2022-02-23
Exercise allergies - test

karstenj:
Added test as mentioned in instructions where the program shall only report the eggs (1) allergy for a given score of 257. In addition remaining array positions in list shall be false.

### DIFF
--- a/exercises/practice/allergies/test_allergies.c
+++ b/exercises/practice/allergies/test_allergies.c
@@ -394,6 +394,13 @@ static void test_score_roll_over(void)
    allergen_list_t list = get_allergens(257);
    TEST_ASSERT_EQUAL(1, list.count);
    TEST_ASSERT_TRUE(list.allergens[ALLERGEN_EGGS]);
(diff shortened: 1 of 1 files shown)


karstenj:
Hi @wolf99,

okay, I see your point. I'm new on exercism and I was not aware, that the test are generated from a generic description.

The instructions, I mean, is the instruction page related to the exercise: https://github.com/exercism/c/blob/main/exercises/practice/allergies/.docs/instructions.md. The last statement on that page mentions a case were the score contains a bit which is not related to an allergy. But this case is not covered by the current test. 

So, the correct approach would be to modify the JSON-file? How are the tests generated from that JSON-file?

wolf99:
Hi @karstenj 

No worries everyone starts somewhere 😊.

The tests on this language track repository are manually created, but follow the spec.

The specs are in a separate repository.

I think you are correct that this is a missing test case. For a test case to be added a PR should be added on the spec itself in their repository: https://GitHub.com/exercism/problem-specifications .

Happy to help if you need any support on raising such a PR!

wolf99:
Fixed in #797

PR #765, opened 2022-03-01
Lasagna: add exercise

wolf99:
Adds Lucian's Luscious Lasagna exercise to cover the basics concept.

Ci for Configlet will fail until #735 is merged, due to missing concept config that is added in that PR.


bobahop:
I took the liberty of moving the exercise stub from the concepts array to the correct location of the exercises/concept array.

config.json
colinleach, line 1: `"status": "wip"` is included twice

siebenschlaefer:
Do you really want to introduce `extern` variables in the first exercise, as one of the fundamental basics of C?  
I am not involved here but I'd prefer keeping this a function, too, like in the C++ track.

Also, one thing that came up when vaeng wrote this exercise on the C++ track:  
Do you want to introduce `.h` files right at the start?  
As a mentor on the C track I often get asked what these `.h` files are even from students who apparently were able to solve the first few exercises easily. If `.h` files are a part of this first exercise I think you will either have to explain them, add comments that tell the students to ignore parts of the initial stub, and/or have additional tooling so that students won't face cryptic error messages.

colinleach:
Am I right in thinking C99 has `const` values? (sorry, I mostly used older versions of C, which would use a `#define` for this). `expected_minutes_in_oven` never needs to be changed, and most tracks define it as a constant if possible..

siebenschlaefer:
C99 has `const` variables. But the problem is not `const` in this case but `extern`.  
You need to declare a variable as `extern` in the `.h` file if the tests should be able to access it.  
But is that something you want to do in this first concept exercise?

BTW: In the past three years I tried three times writing a C syllabus. But I could never figure out how to start it. Most tutorials start with a standalone program that prints "Hello, World!". But with our test-centered exercises the student must at least understand what a function is, how to specify a return type, and how to return something. Depending on what the first exercise is about they will have to understand other things, too, like maybe strings or arithmetic or variables or parameters or function declarations/definitions or `.h` files or ... before they can solve the exercise.  
I'm curious how you folks will approach that. I'm rooting for you.

colinleach:
> I'm curious how you folks will approach that

Me too! It's early days...

As a personal prejudice, I think it's important to start with _something_, even if it will need fixing later. It is easy for any of us to fall into a paralysis of indecision.

FYI, I'm an experimental scientist, not a professional programmer. I spent my whole career getting most things wrong at first, then doing better.

colinleach:
> I spent my whole career getting most things wrong at first

I worked in chemistry labs, so "getting things wrong" could lead to flames or loud noises. Exercism is gentler!

PR #770, opened 2022-03-04
Adding Prerequisites & Practices

vizzv:
adding prerequisites and practices


wolf99:
Hi @vizzv 

Thanks for raising a PR.
I'm not sure where this change is coming from though.

- Where have the preconditions come from?
- Why change the casing everywhere?
- I thought the `"practices" field was supposed to list the slugs of other practice exercises, but this lists concepts, where have those come from?
- Which concepts are being referenced by the prequisites and practices values?

PR #772, opened 2022-03-14
Update config.json

bobahop:
Add bits directory to concepts. Needed for #771.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1000,6 +1000,11 @@
       "name": "Basics"
     },
     {
+      "uuid": "20948c69-20f7-498d-b743-fada00e5c79d",
+      "slug": "bits",
+      "name": "Bits"
+    },    
+    {
       "uuid": "2c087541-a5fa-4cfd-b0f1-5355f933b94b",
       "slug": "booleans",
       "name": "Booleans"


bobahop:
I don't know what the configlet problem is. It's harshing my mellow.

wolf99:
Looks like GitHub was unable to correctly run the script that fetches configlet.
No idea why this was the case but there are two possibilities

1. The fetch-configlet script is broken. Unlikely as the script has not changed in over a year and has worked correctly in that time and worked correctly just now from my local machine.
2. GitHub itself died. Possible, especially as the details show that GitHub could not report any output.

Solution: re-trigger the job to see if it works now - I will kick it off now.

wolf99:
Yep, that worked.

Now the problem from configlet itself:

> The `bits` concept is missing its required files. Please create the `/home/runner/work/c/c/concepts/bits` directory with its required files:
> /home/runner/work/c/c/config.json

So this is complaining that while this branch does have bits in the config, there is no directory.
While the job on #771 is complaining of the opposite.

Solution: close this PR and add the change to #771 instead.

bobahop:
Thank you!

The problem is that I started #771 before the latest config.json changes were checked in. But, if I can pull the latest config.json into my branch and then update it there, then that would do it, I think. So I'll look into how to do that...

PR #773, opened 2022-03-16
create dynamic memory management concept

bobahop:


bobahop:
Converting back to draft because I want to create a Linkage concept to go along with a new Static Functions concept. After that, I can link to the scope, linkage and storage duration stuff in About.md instead of re-explaining it there.

concepts/dynamic-memory-management/about.md
```diff
@@ -0,0 +1,285 @@
...
+    // prints The first element of the array is 2
+    printf("The first element of the array is %d\n", numbers[0]);
+    // treat numbers as a single value again
+    // prints numbers is 2
+    printf("numbers is %d\n", *numbers);    
+    free(numbers);
+    numbers = NULL;    
+}
+```
+If the value(s) will definitely be written to before being read, then to use `malloc` can be an efficient way to allocate memory.
```
wolf99, line 94:
````suggestion
```

If the value(s) will definitely be written to before being read, then to use `malloc` can be an efficient way to allocate memory.
````

File: concepts/dynamic-memory-management/about.md [hunk 2]
```diff
@@ -0,0 +1,285 @@
...
+
+`free` is a function declared in `stdlib.h` which is used for deallocating memory which has been dynamically allocated.
+Dynamically allocated memory stays claimed in memory until released by `free`.
+Some operating systems may automatically free dynamically allocated memory when the program finishes, but it is best practice not to depend on that.
+`free` has one parameter, which is a pointer to the memory which was dynamically allocated.
+The use or lack of use of `free` is associated with several memory issues.
+
+#### memory leak
+
+Not freeing memory after it is no longer needed is referred to as a memory leak.
```
wolf99, line 110:
```suggestion
Not freeing dynamically allocated memory after the reference to that memory reaches the end of its lifetime, or is otherwise destroyed, is referred to as a memory leak.
```

concepts/dynamic-memory-management/about.md
```diff
@@ -0,0 +1,285 @@
...
+This means the compiler can arbitrarily handle it any way it wants to.
+The following example, compiled with two different compilers, can have two different outcomes
+
+```c
+#include <stdio.h>
+#include <stdlib.h>
+
+int * my_function(int count) {
+    // pointer-to-void from malloc is converted to a pointer-to-int
+    int * numbers = (int *) malloc(sizeof(int) * count);
```
wolf99, line 165:
```suggestion
    int * numbers = malloc(sizeof(int) * count);
```

concepts/dynamic-memory-management/about.md
```diff
@@ -0,0 +1,285 @@
...
+    // prints numbers is 1
+    printf("numbers is %d\n", *numbers);
+    free(numbers);
+    // first compiler generates runtime error: double free detected in tcache 2
+    // second compiler generates no error
+    free(numbers);
+}
+```
+
+The first compiler generated no runtime error when `numbers` was set to `NULL` between each call to `free`, but that would not be guaranteed to work with a third compiler.
```
wolf99, line 182: I read this as there should be a `numbers = NULL;` line somewhere in the above example, specifically between the two calls to `free()`. But I don't see such. 
Do I misunderstand?

concepts/dynamic-memory-management/about.md
```diff
@@ -0,0 +1,285 @@
...
+    // 0 or it could be 18728 or anything else
+    printf("The sixth element of the array is %d\n", numbers[5]);
+    numbers[5] = 6;
+    // shrink the size of the array
+    numbers = my_function_2(numbers, 3);
+    // prints The sixth element of the array is [whatever is there]
+    // the removed memory is not initialized, so it could be
+    // 6 or it could be 43164 or anything else
+    printf("The sixth element of the array is %d\n", numbers[5]);
+}
```
wolf99, line 277: Should there be a `free()` somewhere in this example?

concepts/dynamic-memory-management/introduction.md
```diff
@@ -0,0 +1,162 @@
+# Introduction
+
+Dynamic memory is implemented by library functions which the programmer uses to allocate and deallocate memory at runtime.
+Since the programmer controls the allocation and deallocation of memory, dynamic memory is as safe as the programmer is free from human error.
+For that reason, dynamic memory is one of the  major sources for programming bugs and a major target for hacking attacks.
+## malloc
```
wolf99, line 6:
```suggestion
For that reason, dynamic memory is one of the  major sources for programming bugs and a major target for hacking attacks.

## malloc
```

concepts/dynamic-memory-management/about.md
bobahop: Yes, there absolutely should. Good catch! Thanks!

concepts/dynamic-memory-management/about.md
bobahop: Good catch! I just put it in a commented line, since the program was run both with and without setting `numbers = NULL`.

concepts/dynamic-memory-management/about.md
```diff
@@ -0,0 +1,285 @@
...
+The `sizeof` operator is commonly used to get the size of each element's data type.
+The following example demonstrates this
+
+```c
+#include <stdio.h>
+#include <stdlib.h>
+
+int * my_function(int count) {
+    // pointer-to-void from malloc is converted to a pointer-to-int
+    int * numbers = malloc(sizeof(int) * count);
```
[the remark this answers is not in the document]
bobahop: Thanks, I think I want to leave it as is to be a bit more straightforward.

bobahop:
@wolf99 Thank you very much for the attention to detail and suggestions for improvement!

wolf99:
Is this ready to go @bobahop ?

bobahop:
@wolf99 , if you approve it, it is.

wolf99:
Ah, apologies for the long delay then!

wolf99 (approved):

cjsutton77:
Would it be possible to add a "free" for the <code>realloc</code> example?  Also a <code>return</code> at the end of main as well.

Can we contribute and edit?

PR #775, opened 2022-03-23
create blocks concept

bobahop:


concepts/blocks/about.md
```diff
@@ -0,0 +1,59 @@
+# About
+
+A block is a set of statements and declarations grouped into one syntactic unit.
+Blocks are often, but not always, defined within curly braces.
+Blocks can be nested within functions and within other blocks.
+
+In the following example, `number` is defined within the `main` function.
```
wolf99, line 7: Do you mean to imply that a function itself constitutes a block? (As well as the signature)
If yes, then it may be best to explain that explicitly.
This would then make clearer the meaning of "inner block" as used in following paragraphs.

concepts/blocks/introduction.md
```diff
@@ -0,0 +1,59 @@
+# Introduction
```
wolf99, line 1: I'm not really clear on why the about.md is an exact copy of the introduction.md. Maybe there is some reasoning for this, but I don't recall. Is it expected?

concepts/blocks/about.md
```diff
@@ -0,0 +1,59 @@
...
+The error message may be confusing, as `flag` _is_ used earlier in the function.
+However, `flag` is defined within a block inside the function.
+The inner block can "see" the variable `number` in its outer function, but the outer function can't "see" the `flag` variable defined within its inner block.
+Once execution leaves the block, the `flag` variable is automatically destroyed and is no longer available to the program.
+The `main` function itself sees the `flag` identifier for the first time in the last `printf` statement.
+If the last `printf` statement is removed, the program will compile and print `1` when run.
+That is because the `flag` variable is valid in the block containing the first `printf` statement.
+
+As of `C99`, single-line `if`, `for`, `while` and `do` statements are also considered blocks.
+In the following example, although `i` is not defined within curly braces, it exists only in the block defined by the `for` statement, and is not visible to the outer `main` function.
```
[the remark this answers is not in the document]
bobahop: Scope is related to block in that a variable defined within a block has block scope, It may have automatic or static storage, but it won't have external linkage.

concepts/blocks/about.md
bobahop: Yeah, thanks, I've been thinking about whether to further distinguish blocks vs functions. For instance, two variables can have the same name in different blocks in the same function, but two labels cannot, but that really relates more to function scope. I can probably think a bit more objectively about how to put it with a little time to switch from creating mode to editing mode.

concepts/blocks/introduction.md
bobahop: Usually with more complex topics there is a bigger difference between introductory stuff and details. I usually write the about first and then take a subset for the introduction. Here, though, there didn't seem to be much to take away. Perhaps that means the "about" should be more detailed. But my concern is not to get too much into the weeds with elaborations on scope and such just for the concept of block. A lot of that material is already in the dynamic memory concept.

bobahop: Thank you for looking all of this over!

concepts/blocks/about.md
bobahop: For the "about" I added the bit about labels to explain the difference between function and block. I left that detail out of the introduction.

concepts/blocks/introduction.md
bobahop: For "about" I added a bit explaining how multiple single-line statements and comments could be one block without needing curly braces. I left that detail out of the introduction.

concepts/blocks/about.md
```diff
@@ -57,3 +57,104 @@ int main() {
...
+            goto start;
+        }
+    }
+
+}
+```
+
+Single-line statements can be "stacked" and still be within one block without the use of curly braces.
+In the example below, the `for` statement contains an `if` statement which in turn contains a `printf` statement.
+Although there are two total statements in the `for` block, it "sees" only its single-line `if` statement.
```
wolf99, line 150: `"sees"` is kind of how I think of it too... but I feel like there is a more correct way to phrase this.

Something like "Although there are two total statements in the `for` block, only its single-line `if` statement is evaluated as the body of the `for` loop.".
What do you think?

bobahop: Followed your recommendation and changed to this

```
Although there are two total statements in the `for` block, only its single-line `if` statement is evaluated as the body of the `for` loop.
The `if` statement's single-line `printf` statement in turn is evaluated as its body.
If nested blocks are each only a single line, none of the blocks need curly braces.
```

wolf99:
LGTM Bob, just need to resolve the conflict and I'll approve.

bobahop:
Conflict is resolved.

wolf99 (approved):

PR #777, opened 2022-03-27
create calling functions through function pointers concept

bobahop:


concepts/calling-functions-through-function-pointers/about.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+The function signature of `bool (*func) (int)` for the first argument to `test_number` matches the signature of the functions passed in.
+Both `is_even` and `is_odd` take an `int` parameter and return a `bool`.
+In `(*func)`, `func` represents the name of the function to be used.
+if `(*fn)` were used, then it would have been called like so: `printf(text, number, fn(number)? "true": "false");`.
+To satisfy the function pointer parameter, we pass the name of a function having the expected signature when calling `test_number`.
+Thus, we see that a function name points to the address of that named function.
+
+The following example compiles but crashes when calling a function without the expected signature
```
wolf99, line 42: The rest from here onward seems  like unnecessary level of detail, maybe better to just have a few lines that both the signature and implementation must be correct?

config.json
```diff
@@ -1004,6 +1004,11 @@
       "slug": "booleans",
       "name": "Booleans"
     },
+    {
+      "uuid": "1721552b-9c55-432d-a776-05e618c25bb4",
+      "slug": "calling-functions-through-function-pointers",
+      "name": "Calling Functions Through Function Pointers"
```
wolf99, line 1010: How about just "Function pointers" ?

bobahop: Well, I took it directly from your document. 😄 I like that it's specific, but I don't feel strongly about it.

concepts/calling-functions-through-function-pointers/about.md
bobahop: That's definitely why I left it out of the Introduction. Let me mull on it. I think I was experimenting with my own what-ifs for my own learning, but you make a good point that it may be too much detail.

bobahop: Okay, I slashed it all to just

```
Calling a function pointer with the wrong signature results in undefined behavior (UB.)
Depending on the compiler, it may not compile, or it may compile with or without warnings but then suffer a runtime crash.
```

config.json
wolf99: 😄 

Both approaches have fairly clear meaning I think. The main advantage of "Function pointers" is that it is shorter.

This helps a couple of things:

1. Rendering on the site in different forms, e.g. in lists, tables, etc (especially on mobile)
2. Maintenance of the repository (less typing for maintainers!)

bobahop: Good points. Okay, it's all renamed.

wolf99 (approved):

config.json
bobahop: What's left to do: another review or just squash and merge?

Issue #758, opened 2022-02-12
Diamond leaks memory even when all tests pass


bobahop:
I noticed when solving Diamond that `make memcheck` detected memory leaks even for passing tests. In case it was an issue with my solution, I tried the most-starred solution that currently passes and got essentially the same results of `2680 byte(s) leaked in 62 allocation(s).`

I've only just started looking into it, and probably won't really dig into it until sometime next week or the week after.

Here is the `make memcheck` output

```
Compiling memcheck
test_diamond.c:155:test_rows_degenerate_case_with_a_single_a_row:PASS
test_diamond.c:156:test_rows_degenerate_case_with_no_row_with_3_distinct_groups_of_spaces:PASS
test_diamond.c:158:test_rows_smallest_non_degenerate_case_with_odd_diamond_side_length:PASS
test_diamond.c:160:test_rows_smallest_non_degenerate_case_with_even_diamond_side_length:PASS
test_diamond.c:162:test_rows_largest_possible_diamond:PASS

-----------------------
5 Tests 0 Failures 0 Ignored
OK

=================================================================
==248==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1300 byte(s) in 25 object(s) allocated from:
    #0 0x7fd30969ddc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
    #1 0x7fd30a00d13e in write_row diamond.c:23
    #2 0x7fd30a00d307 in make_diamond diamond.c:47
    #3 0x7fd30a00de19 in test_rows_largest_possible_diamond test_diamond.c:146
    #4 0x7fd30a00ce4c in UnityDefaultTestRun test-framework/unity.c:1837
    #5 0x7fd30a00df91 in main test_diamond.c:162
    #6 0x7fd3093b70b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)

Direct leak of 1300 byte(s) in 25 object(s) allocated from:
    #0 0x7fd30969ddc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
    #1 0x7fd30a00d352 in make_diamond diamond.c:53
    #2 0x7fd30a00de19 in test_rows_largest_possible_diamond test_diamond.c:146
    #3 0x7fd30a00ce4c in UnityDefaultTestRun test-framework/unity.c:1837
    #4 0x7fd30a00df91 in main test_diamond.c:162
[...]

PR #780, opened 2022-04-08
Improve "diamond" by adding a cleanup function.

siebenschlaefer:
As discussed in issue #758 this exercise currently requires the function `make_diamond()` to return an array of `char *` where the first element is a dynamically allocated string, and the other elements point to substrings of that first element.
That might be clever because it avoids unnecessary allocations, but it's not obvious.

This PR adds a second function `free_diamond()` as part of the exercise.
That way the students can choose how the result is allocated and are responsible for deallocating it completely.
That should be a little bit easier to understand. And it still allows the "clever" solution.

### DIFF
--- a/exercises/practice/diamond/.meta/example.c
+++ b/exercises/practice/diamond/.meta/example.c
@@ -45,3 +45,10 @@ char **make_diamond(const char letter)
 
    return diamond;
 }
+
+void free_diamond(char **diamond)
+{
+    free(diamond[0]);
+    free(diamond);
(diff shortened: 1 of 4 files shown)


ryanplusplus:
@siebenschlaefer this looks good to me. I am going to merge and format afterward since the format workflow doesn't work on your fork.

ryanplusplus (approved):

siebenschlaefer:
Thank you!

PR #785, opened 2022-04-15
Remove Broken URL

orsenthil:
Remove Broken URL

### DIFF
--- a/exercises/practice/etl/.meta/config.json
+++ b/exercises/practice/etl/.meta/config.json
@@ -23,5 +23,6 @@
       ".meta/example.h"
     ]
   },
-  "source": "The Jumpstart Lab team"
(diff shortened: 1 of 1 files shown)


ryanplusplus:
I think this comes from upstream. Can you apply the fix [here](https://github.com/exercism/problem-specifications/blob/main/exercises/etl/metadata.toml) instead?

wolf99:
Correct URL might be http://tutorials.jumpstartlab.com/  ?

orsenthil:
https://github.com/exercism/problem-specifications/pull/2014 

- Should we need fixes at both places?

wolf99:
If you fix it there, it will eventually get sync-ed to this repository.

wolf99:
This can be closed as the PR on the problem-specs repo will cover this.

PR #790, opened 2022-05-02
create linkage and static functions concepts

bobahop:


wolf99:
Should either/both of these concepts cover `extern`?

bobahop:
> Should either/both of these concepts cover `extern`?

Added a bit about `extern` in the About for Linkage.

wolf99 (approved):

PR #793, opened 2022-05-19
Updated call to assert to provide a better error

JamesMConroy:
The order of the expected and the actual strings were revered giving a
error message where the output of the function was referred to as
expected.

Ex:

```
test_word_beginning_with_a:FAIL: Expected 'apple' Was 'appleay'
```

Now the message is:

```
test_word_beginning_with_a:FAIL: Expected 'appleay' Was 'apple'
```

Note: in both these examples the function under test just retunes the
input string.


ryanplusplus (approved):

ryanplusplus:
Thanks, great catch!

PR #794, opened 2022-05-23
Use (void) for create_reactor prototype

ryanplusplus:
I was reviewing a submission for `reactor` and noticed that it used an empty argument list (`()`) instead of `(void)` for the `create_reactor` prototype and traced this back to the provided template.

This changes the template and sample solution to use the more correct `(void)`.

For more background, please see https://stackoverflow.com/questions/51032/is-there-a-difference-between-foovoid-and-foo-in-c-or-c


wolf99 (approved):
Good catch!

PR #807, opened 2022-07-02
improve starting point of program

Aliwahid17:
it will help beginner c programmers to know where to start writing code.

### DIFF
--- a/exercises/practice/armstrong-numbers/armstrong_numbers.c
+++ b/exercises/practice/armstrong-numbers/armstrong_numbers.c
@@ -1,1 +1,8 @@
 #include "armstrong_numbers.h"
+#include <stdio.h>
+#include <math.h>
+
+bool is_armstrong_number(int candidate)
+{
+  // start writing your code here
+}


siebenschlaefer:
I'm slightly in favor of that change. armstrong-numbers seems to be the first exercise for many students, let's keep the entry barriers low.  
But I don't like this two includes, the exercise can be solved with or without functions from `<stdio.h>` or `<math.h>`.  
(Disclaimer: I'm a mentor, not a maintainer. I can only voice my opinion.)

ryanplusplus:
> I'm slightly in favor of that change. armstrong-numbers seems to be the first exercise for many students, let's keep the entry barriers low. But I don't like this two includes, the exercise can be solved with or without functions from `<stdio.h>` or `<math.h>`. (Disclaimer: I'm a mentor, not a maintainer. I can only voice my opinion.)

I agree. I don't mind the addition of the skeleton function, but the includes aren't necessary and I think they should be removed.

exercises/practice/armstrong-numbers/armstrong_numbers.c
```diff
@@ -1 +1,8 @@
 #include "armstrong_numbers.h"
+#include <stdio.h>
+#include <math.h>
```
ryanplusplus, line 3:
```suggestion
```

wolf99:
Is hello-world no longer the first exercise?

If this is not obvious, should there be some update to the docs?

Sorry last time I mentored was pre-v3!

siebenschlaefer:
Yes, hello-world is still the first one. But since it can be solved by replacing `NULL` with `"Hello, world!"`, that doesn't force students to understand what functions are, where they are defined/declared, why there's no `main()`, etc.  
I've talked to quite a few students (in mentoring sessions or on gitter) who struggled with the first "real" exercise: armstrong-numbers.
BTW: This is what the C track currently looks like:
c_track_exercises

PR #811, opened 2022-08-19
grains: Improve the tests with an _EQUAL_ macro.

siebenschlaefer:
Currently the tests compare the expected and actual result with `==` and
call `TEST_ASSERT()` which leads to error messages like
"FAIL: Expression Evaluated To FALSE".
With `TEST_ASSERT_EQUAL_UINT64()` the error message is much easier to
understand:
"FAIL: Expected 18446744073709551615 Was 18446744073709551614"

### DIFF
--- a/exercises/practice/grains/test_grains.c
+++ b/exercises/practice/grains/test_grains.c
@@ -11,61 +11,61 @@ void tearDown(void)
 
 static void test_square_1(void)
 {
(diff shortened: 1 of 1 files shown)


ryanplusplus (approved):

siebenschlaefer:
That was extremely quick. Thanks!

Issue #812, opened 2022-09-04
Newer C standard


jtepe:
Is there a reason the code is restricted to C99? Most common compilers ([gcc](https://gcc.gnu.org/wiki/C11Status), [clang](https://clang.llvm.org/c_status.html)) are already supporting C17.

This issue #53 is back from 2016. Today, learning the language means learning the newer versions. I can't see a reason for sticking to such old standards.

wolf99:
Hi!
Thanks for the good question and for finding and referencing the past question!

GCC and Clang do indeed support recent C standards in _hosted_ environments (e.g. environments with an OS).

However C is still used heavily in non-hosted environments (bare-metal, RTOS, near-hosted, etc). These environments usually only support much older versions of C. Additionally Microsoft's MSVC only supports C to C99 (last I looked, the situation may have changed since; but C is always second citizen to C++ for MS).

Thus having an older standard caters to the lowest common denominator of students that may use the track.

That said, I don't know if this is something we specifically aim for. FWIW I have no specific objection to moving to newer standards. 

Off the top of my head however I can't think of any currently existing exercise on the track that would make use any of the features of the newer standards (I'm not super familiar with the new features either!). 

Are there any specific features that you had in mind?

siebenschlaefer:
No objections from my side, I'm fine with either.

Just out of curiosity: Would that have any impact on the solutions?
I'm not aware of any exercise where a C11 or C17 solution would look significantly different.

jtepe:
@wolf99 Non-hosted environments are a valid point. However, I look at it from the other side. Learning C as it is now should be the goal. Later you can easily restrict the features you use and adapt to a more limited environment. IMO, learning the language should not be restricted because of that.

@siebenschlaefer No, the exercises (haven't solved all yet) are also solvable with older versions. But I found using newer features often made it easier.

siebenschlaefer:
> But I found using newer features often made it easier.

That's what I wanted to know. Do you have an example for that?

jtepe:
I thougth that most of the "quality of life" improvements I use came from C11 onwards, but looking it up it appears most of them have been introduced by C99 already.

For example, compound literals:
```c
int *p = (int [2]){1, 2};
```
Or array initialization with designators:
```c
int a[] = { [2] = 3 };
```
For C11 I can only point to anonymous struct members:
```c
struct foo {
    int i;
    struct {
        int a;
        int b;
    };
};
...
[...]

PR #815, opened 2022-10-09
armstrong-numbers: update tests.toml

wolf99:
Adds 2 new tests as excluded in the exercise's `tests.toml` using `include = false`.

The tests are excluded because the track currently uses C99 which does not have an integer data type large enough to handle the Armstrong numbers these tests require (128 bits).

C23 will, or does (I am not sure of the current status of C23), have `_BitInt(N)` which would allow 128 bits. So this may be a (admittedly minor) case for #812.

Another alternative would be to change the interface to either allow the value to be split across multiple parameters, or to allow the value split across multiple members of a single struct parameter.

IMHO, these alternatives involve more complication than is expected in this exercise so I have choseen to simply exclude the tests for now.


wolf99:
Draft until #814 is merged.

ryanplusplus (approved):

PR #831, opened 2022-11-21
Binary Search - Do not expect a non-const pointer from a const pointer

ryanplusplus:
Currently in `binary-search` we pass an array via a `const` pointer but then expect a non-`const` pointer back. This requires an illegal cast (removal of `const`) as seen in our example.

This changes the return type from `int *` to `const int *` so that no cast is required.

### DIFF
--- a/exercises/practice/binary-search/.meta/example.c
+++ b/exercises/practice/binary-search/.meta/example.c
@@ -1,6 +1,6 @@
 #include "binary_search.h"
 
-int *binary_search(const int value, const int *arr, const size_t length)
+const int *binary_search(const int value, const int *arr, const size_t length)
 {
    if (0 == length || NULL == arr) {
       return NULL;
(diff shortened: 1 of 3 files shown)


exercises/practice/binary-search/.meta/example.c
```diff
@@ -14,7 +14,7 @@ int *binary_search(const int value, const int *arr, const size_t length)
       } else if (*mid < value) {
          low = mid + 1;
       } else {
-         return (int *)mid;
```
ryanplusplus, line 17: This casted away the `const` qualifier. Whoops!

ryanplusplus:
> And casting the `const` away is only illegal if the program then tries to modify the (originally `const`) object.

My issue is that this implementation can't control that, so returning a non-`const` pointer is inviting undefined behavior. IMO if the caller wants to take on that responsibility, it should be the one making the dangerous cast.

ryanplusplus:
> It follows the example of [bsearch(3)](https://en.cppreference.com/w/c/algorithm/bsearch) which does the exact same thing (take a `const` pointer, do a binary search, return a non-`const` pointer).

Interestingly, C23 makes this `const` safe with a type-generic macro. Neat!

wolf99 (approved):

PR #851, opened 2023-03-08
Binary Search approach

bobahop:
Approach for Binary Search

### DIFF
--- a/exercises/practice/binary-search/.approaches/while-with-if-statements/content.md
+++ b/exercises/practice/binary-search/.approaches/while-with-if-statements/content.md
@@ -1,0 +1,68 @@
+# `while` loop with `if` statements
+
+**binary_search.h**
+
+```c
+#ifndef BINARY_SEARCH_H
+#define BINARY_SEARCH_H
+
+#include <stddef.h>
+
+int* binary_search(int value, const int *arr, size_t length);
+
+#endif
+```
+
+**binary_search.c**
+
+```c
+#include "binary_search.h" 
+
+int* binary_search(int value, const int *arr, size_t length){
+    if (!arr) return NULL;
+    size_t left = 0, right = length;
+    while (left < right) {
+        size_t mid = left + ((right - left) / 2);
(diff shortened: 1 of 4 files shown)


exercises/practice/binary-search/.approaches/while-with-if-statements/content.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+This approach starts by checking for a `NULL` input array.
+It then defines the [`size_t`][size_t] variables to keep track of the `left` and `right` positions being searched for in the array.
+The `left` is initialized to `0` and the `right` is initialized to the length passed in.
+
+The while loop will loop as long as the `left` position is less than the `right` position.
+
+Inside the loop, the middle value is set by `left` plus ((`right` - `left`) divided by `2`).
+- For example, if `left` is `0` and `right` is `10`, then the middle is calculated to `5`.
```
wolf99, line 42: Need blank line before list

wolf99:
Just the one nit, otherwise it looks great! 👍

bobahop:
@wolf99: Thanks, the suggested change was committed.

wolf99:
Sorry bob, the comments have been just comments, not changes.
I'm on mobile app so just jotting comments as I see things.

I think you've been marking comments as resolved rather than committing changes.

Either way it looks like there's no change since I reviewed last.

exercises/practice/binary-search/.approaches/while-with-if-statements/content.md
bobahop: Whoa, sorry! Not used to reviewers doing that. Okay, change made.

Luckily, I think other changes I may have missed on previous exercises were fairly minor formatting issues.

bobahop:
@wolf99: Made the change here and made a couple PRs for the changes I missed in previous approaches.

wolf99 (approved):

wolf99:
All merged, thanks!

Issue #840, opened 2023-01-09
Exercise ResistorColor is confusing or too difficult


jhaand:
Hi,

For the C track I found the last test of this test very confusing. 
It was very easy to map the mnemonics to an enum for the first 3 tests.

https://exercism.org/tracks/c/exercises/resistor-color

But the last test has no signature for the colors() function. And the solution requires to return an array of values on the heap. (static array) I find these quite advanced topics for an easy assignment without any hints.

Which invites people to look up solutions.

It would actually have been easier for me to calculate the resistor value based on 4 random colors.

ryanplusplus:
@jhaand the last test doesn't require you to return an array of values from the heap. The tests communicate this by not `free`ing the result of `colors()`.

Can you help me understand why creating the `colors()` signature is challenging? The function takes no arguments and the line in the tests where it is used give the result type so I would have guessed that it would be straightforward to construct the signature by copying the result's type.

siebenschlaefer:
@ryanplusplus I've experienced the same with other students in this exercise. Quite a few were asking how to solve the last test, others returned a pointer to a local variable asking why they got an error, a third group dynamically allocated the array with `malloc()`.  
I guess many of them are coming from languages with automatic memory management, or are just new to the language and not yet used to object lifetimes and the difference between local and `static` (or global) variables. That makes this last test quite a challenge when they have no idea where to start or get an error that is hard to understand and/or research.

wolf99:
> Which invites people to look up solutions.

I don't think this is a bad thing? Researching how others have solved a common problem is a valid part of software engineering.

I'll agree that having the signature could make this easier, but students need to learn it t some point. With the concepts part of the C track not yet in place we don't have a good way of teaching new things aside from, "here's something new, have a go at it".

Or to put it another way, students will need to face each hurdle at some point.

Having said that I would be interested in any proposals of an alternative `resistor_color` interface!

jhaand:
> @jhaand the last test doesn't require you to return an array of values from the heap. The tests communicate this by not `free`ing the result of `colors()`.
> 
> Can you help me understand why creating the `colors()` signature is challenging? The function takes no arguments and the line in the tests where it is used give the result type so I would have guessed that it would be straightforward to construct the signature by copying the result's type.

I would expect to use an application to determine the actual resistor value. By testing the actual bands of a resistor. If you add the multiplier colors, that wouldn't be so hard. A test of: 
```
[...]

PR #856, opened 2023-03-10
Move resistor color to later in the track and bump the difficulty

ryanplusplus:
Addresses https://github.com/exercism/c/issues/840

### DIFF
--- a/config.json
+++ b/config.json
@@ -64,19 +64,6 @@
         ]
       },
       {
-        "slug": "resistor-color",
-        "name": "Resistor Color",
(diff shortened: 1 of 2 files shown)


fapdash:
Looks fine for me regarding https://github.com/exercism/c/issues/840 but it might make sense to improve the ordering overall, maybe in an extra PR. There is currently a thread in the forum about practice exercise ordering: https://github.com/exercism/c/issues/840 I know that reordering all the exercises can be a lot of work but as a small improvement, for me two-fer and leap are two obvious candidates to put very early since you can solve both of them with a single line statement.

wolf99 (approved):

PR #858, opened 2023-03-10
Hamming approach

bobahop:
Approach for Hamming

### DIFF
--- a/exercises/practice/isogram/.approaches/bitfield/content.md
+++ b/exercises/practice/isogram/.approaches/bitfield/content.md
@@ -1,0 +1,94 @@
+# Bit field
+
+**isogram.h**
+
+```c
+#ifndef ISOGRAM_H
+#define ISOGRAM_H
+
+#include <stdbool.h>
+
+bool is_isogram(const char phrase[]);
+
+#endif
+```
+
+**isogram.c**
+
+```c
+#include "isogram.h"
+#include <stdint.h>
+
+bool is_isogram(const char phrase[])
+{
+   if (!phrase)
+      return false;
+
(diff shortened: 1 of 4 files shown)


wolf99:
Um, seems like these are the files for Bob, not hamming?

bobahop:
Ugh! I don't know how I did it, but the files for Bob and Hamming are both in the same PR. I made the change in this PR you asked for in the original Bob PR (`has_letter`). Perhaps the changes for Hamming can be made in this PR, and when okay this can be merged and the original Bob PR can just be closed without merging? I'm wary of deleting the Bob files from here in case they somehow get deleted from the original Bob branch, too, since one may branch off the other. But I can try that if you want to merge the Bob PR first. Sorry for the mess!

exercises/practice/hamming/.approaches/increment-pointers/content.md
```diff
@@ -0,0 +1,55 @@
...
+This approach starts by checking if either the `lhs` or `rhs` strings are `NULL`.
+If so, then the function returns `-1`.
+
+```exercism/note
+There is a difference between a `NULL` pointer and a valid pointer pointing to a null character (`'\0'`).
+```
+
+The variable to keep track of the distance count is initialized to `0`.
+
+The `for` loop begins with no initialization statement.
```
wolf99, line 42: Would be useful to explain why this is the case here.

wolf99 (changes requested):

wolf99:
Yeesh, didn't see your comment here until after the bob or was merged. If there's issues I'll guess there'll be a merge conflict to sort out, but otherwise it may be ok.

PR #861, opened 2023-03-11
Hamming approach

bobahop:
Approach for Hamming


bobahop:
@wolf99: Original Hamming PR with Bob files closed. Hamming approach recreated with change to explain not initializing `for` loop.

Thanks for all of your help!

wolf99 (approved):

PR #864, opened 2023-03-11
Difference of Squares approach

bobahop:
Approach for Difference of Squares


bobahop:
@wolf99: C approach for Difference of Squares ready for review.

wolf99 (approved):

PR #865, opened 2023-03-13
Linked List approach

bobahop:
Approach for Linked List


exercises/practice/linked-list/.approaches/circular-doubly-linked-list/content.md
```diff
@@ -0,0 +1,302 @@
...
+This approach adds a `length` field which is initialized to `0` on list creation and updated as nodes are added and removed.
+
+A circular doubly-linked list populated with at least one node does not have `NULL` values for any node's `prev` or `next` fields.
+
+- If there is only one node in the list, its `prev` and `next` fields will point to itself.
+- If there are two nodes in a list, node `A` will point to node `B` for both its `prev` and `next` fields.
+Likewise, node `B` will point to node `A` for both its `prev` and `next` fields.
+- If there are three nodes in a list, node `A`'s `prev` will point at node `C` and node `A`'s `next` will point at node `B`.
+Node `B`'s `prev` will point at node `A` and node `B`'s `next` will point at node `C`.
+Node `C`'s `prev` will point at node `B` and node `C`'s `next` will point at node `A`.
```
wolf99, line 176: Should these all be part of the same list item?
Hard to tell on mobile but it looks like the latter two lines will be separate paragraphs after the list?

wolf99, line 176: This explanation is good!
But the concept may be confusing for people not familiar with linked lists in general... It seems like an image would help; does exercism render graphs or images within markdown?

bobahop: One-node description is in its own bullet point, two-nodes description is in its own bullet point, and three-nodes description is in its own bullet point.

bobahop:
Removed the cast from malloc abnd added some explanation about setting a freed pointer to `NULL`.

"Pointing at" and "pointing to" are synonymous in English.

There are some other valid points (I especially like the idea of a graphic, but I'm not the one to implement, as I'm a porr visualizer), but I'm not sure they are in keeping with the concept of optimistic merging as explained in this [blog post](https://exercism.org/blog/freeing-our-maintainers).

>The one exception is that we will continue to accept PRs for Approaches and Articles, and will implement an org-wide optimistic merging policy that aims to populate a baseline of Approaches across exercism, and allow for incremental improvements, using the following rules:

>If the code solves the exercise and is syntactically and semantically idiomatic (i.e. it looks like $LANG code) it should be merged. If not, it should be fixed by the PR author.
If a maintainer wants to make changes to the content (e.g. to improve the advice, tweak things, highlight better/alternative/more-idiomatic approaches) then that should be done in a follow up PR.

bobahop:
Ah! Hold on, I was editing in two sperate browser instances...

siebenschlaefer:
> I'm not sure they are in keeping with the concept of optimistic merging

Oh, I see. I will not comment on these PRs again. Sorry.

bobahop:
@siebenschlaefer: Thanks for your time and attention. I think I do want to make the `pop` change for `list_destroy`...

bobahop:
@wolf99: Made some changes in response to suggestions.

wolf99 (approved):

PR #866, opened 2023-03-14
Beer Song approach

bobahop:
Approach for Beer Song

### DIFF
--- a/exercises/practice/beer-song/.approaches/complexity-in-data/content.md
+++ b/exercises/practice/beer-song/.approaches/complexity-in-data/content.md
@@ -1,0 +1,183 @@
+# Complexity in Data
+
+**beer_song.h**
+
+```c
+#ifndef BEER_SONG_H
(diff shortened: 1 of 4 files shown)


bobahop:
@wolf99: C approach for Beer Song ready for review.

wolf99 (commented):
Looks good!
One thing though, the code refers to the sets of words as `LINE`s, whereas the text refers to verses.
It may reduce confusion for the code to also use the verses term?

bobahop:
@wolf99: Thanks!

The text refers to LINEs and verses, as there are two lines in each verse, as it states here

```
Arrays are defined to hold the varying values for the first and second lines of each verse. 
```

Do you have a suggestion on how to further clarify?

wolf99:
Ah, I misunderstood.
Then all seems good to me, approved!

wolf99 (approved):

PR #868, opened 2023-03-15
Pig Latin approach

bobahop:
Approach for Pig Latin.


bobahop:
@wolf99: C Pig Latin approach ready for review. Whew!

bobahop:
@wolf99: I made a it of a hash with the `strncat`, so let me tweak that...

ErikSchierboom (approved):

PR #893, opened 2023-05-01
fix: formatting

tosiek88:
in https://exercism.org/tracks/c/exercises/saddle-points formatting looks wird
```
|----------- 1 | 9 8 7 8 2 | 5 3 2 4 <--- potential tree house at row 2, column 1, for tree with height 5 3 | 6 6 7 1
```


ryanplusplus:
Hey @tosiek88, this comes from https://github.com/exercism/problem-specifications. It looks like it's already fixed upstream so all we need to do is regenerate the docs in this repo:
https://raw.githubusercontent.com/exercism/problem-specifications/main/exercises/saddle-points/instructions.md

Issue #907, opened 2023-07-21
Expected and actual assertion arguments reversed in some tests


jkersell:
Two tests in the High Scores exercise have the expected and actual arguments reversed which causes the assertion failure message to read incorrectly.

jkersell:
I've created a pull request to resolve this, however, I'm not sure if I should be assigning a reviewer or not. If so, I don't know who that would be.

wolf99:
Fixed in #908

PR #908, opened 2023-07-21
Correct order of assertion arguments in high-scores (#907)

jkersell:
This change reverses the order of the actual and expected arguments for two tests so the assertion failure message makes more sense.


wolf99 (approved):

wolf99:
Fixes #907

PR #918, opened 2023-09-26
Add test. Negative numbers are not armstrong numbers

pfmaggi:


ryanplusplus:
Hey @pfmaggi, these tests come from upstream in https://github.com/exercism/problem-specifications/blob/main/exercises/armstrong-numbers/canonical-data.json. If you think this should be added, the first step is to make a change upstream.

pfmaggi:
Thanks for the comment!
I've opened https://github.com/exercism/problem-specifications/pull/2332 and will close this PR.

Issue #319, opened 2018-07-26
gigasecond: simply adding a billion second fails in TZs without DST.


NobbZ:
I can't explain why, but this is a fact we discovered today.

A simple `n + 1e9` fails with an our off ins such timezones if it falls into a timespan where the creator of the tests had DST.

I'm not sure if this is related to the canonical data or to the implementation of ``time.h`` in C, but I have not yet seen similar problems or heard thereof in other languages.

wolf99:
Is there a bug report on this? I'd look into this but it seems to work OK on my system. I'm in UTC-0 with DST of +1 running tests on Ubuntu 18.04 with GNU GCC with stdlib.
I'd need further information to figure it out I think

NobbZ:
Activate yourself a timezone that has no DST at all. The student in question I talked about this problem earlier today was in Moscow Time, where no DST applies since Oct'14 (https://en.wikipedia.org/wiki/Moscow_Time).

In his timezone the example solution fails.

ryanplusplus:
@NobbZ good catch. This probably affects almost every track, right?

Do you know of an elegant solution to this? I'd imagine that we can test the current locale for DST and adjust the test based on this runtime test, but that will probably be confusing for a student.

NobbZ:
As I said, I have never seen similar problem s on other tracks, I had posted at the problem-specifications repository if I were thinking it were more global.

I will try to reproduce the issue in other languages over the weekend. At least if deploy today doesn't break anything ;)

ZapAnton:
When I ran into the problem, simply commenting the `date.tm_isdst = 0;` line in the test suite was
enough to pass the tests.

Could it be a solution for the issue on this track or would you prefer a more general one?

ryanplusplus:
@ZapAnton I think that is just letting garbage from the stack fill that value. In your case, it probably was something non-zero, but this won't be a good general solution because it will vary for every student.

DrJPizzle:
Firstly: Hi.

Secondly:
I think the issue here is mktime();

This will always make some assumptions about the daylight savings. I think a negative number in the tm_isdst field (make best guess as to if to apply DST) is generally the best approach if you are just given a system time and need to deal with it. This may have been what happened to @ZapAnton. 

[...]

PR #925, opened 2023-11-04
Rework gigasecond to be slightly more challenging

### LINKED ISSUE #924: Rework the gigasecond exercise?
I was highly underwhelmed by the gigasecond exercise, where (spoiler) the solution is adding a constant to the function parameter.

I'd like to do something to make it a bit more challenging, something to force the student to use time.h

Are maintainers open to that?

glennj:
I'm not terribly experienced in C. Have I implemented an OK way to handle strings without malloc/free?

Closes #924

### DIFF
--- a/exercises/practice/gigasecond/.meta/example.c
+++ b/exercises/practice/gigasecond/.meta/example.c
@@ -1,6 +1,10 @@
 #include "gigasecond.h"
 
-time_t gigasecond_after(time_t start)
-{
-   return start + 1000000000;
+#define GIGASECOND 1000000000
+
+void gigasecond(time_t start, char *buffer, size_t size) {
+    time_t after = start + GIGASECOND;
+    struct tm *t = gmtime(&after);
+    strftime(buffer, size, "%Y-%m-%d %H:%M:%S", t);
 }
+
--- a/exercises/practice/gigasecond/.meta/example.c
+++ b/exercises/practice/gigasecond/.meta/example.c
@@ -2,9 +2,9 @@
 
 #define GIGASECOND 1000000000
 
-void gigasecond(time_t start, char *buffer, size_t size) {
-    time_t after = start + GIGASECOND;
-    struct tm *t = gmtime(&after);
-    strftime(buffer, size, "%Y-%m-%d %H:%M:%S", t);
+void gigasecond(time_t start, char *buffer, size_t size)
+{
+   time_t after = start + GIGASECOND;
+   struct tm *t = gmtime(&after);
+   strftime(buffer, size, "%Y-%m-%d %H:%M:%S", t);
 }
-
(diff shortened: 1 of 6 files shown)


glennj:
> Hmm I could be wrong but IIRC the intention of the gigasecond exercise is to be relatively simple.
> Introducing time.h In any significant way may add needless complexity and could lead to different results depending on platform.

@wolf99 how does this submission meet your concerns?

exercises/practice/gigasecond/test_gigasecond.c
```diff
@@ -1,6 +1,8 @@
 #include "test-framework/unity.h"
 #include "gigasecond.h"
 
+#define BUFFER_SIZE 20   // "YYYY-mm-dd HH:MM:SS"
```
ryanplusplus, line 4:
```suggestion
#define BUFFER_SIZE (strlen("YYYY-mm-dd HH:MM:SS") + 1)
```

siebenschlaefer: Alternatively: `#define BUFFER_SIZE sizeof("YYYY-mm-dd HH:MM:SS")`

ryanplusplus: > Alternatively: `#define BUFFER_SIZE sizeof("YYYY-mm-dd HH:MM:SS")`

That's much better! I don't know why that never occurred to me.

ryanplusplus:
Can you bump up the difficulty for `gigasecond` in `config.json`?

exercises/practice/gigasecond/test_gigasecond.c
glennj: Should it be a `static size_t` instead of a #define so it only needs to be calculated once?

ryanplusplus: It will be calculated at compile time even if it's a `#define`.

exercises/practice/gigasecond/.meta/example.c
```diff
@@ -1,6 +1,10 @@
 #include "gigasecond.h"
 
-time_t gigasecond_after(time_t start)
+#define GIGASECOND 1000000000
+
+void gigasecond(time_t start, char *buffer, size_t size)
 {
-   return start + 1000000000;
+   time_t after = start + GIGASECOND;
+   struct tm *t = gmtime(&after);
```
wolf99, line 8: The implementation changes look good to me.
I _think_ this usage does not run afoul of #319 but maybe @ryanplusplus can double-check?

wolf99 (approved):

exercises/practice/gigasecond/.meta/example.c
ryanplusplus: I _think_ it's still okay because of the fix we have in place.

ryanplusplus (approved):

glennj:
I was totally Chesterton's Fencing with that construct_date function. Didn't understand why it's there, didn't complain about it ;)

PR #927, opened 2023-11-11
Fix MAX_STR_LEN constant value in clock exercise

toy:


exercises/practice/clock/.approaches/introduction.md
```diff
@@ -24,7 +24,7 @@ thus keeping the code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yoursel
 ```c
 #include <stdbool.h>
 
-#define MAX_STR_LEN 14 // "##:##\0" for int values
+#define MAX_STR_LEN (5 + 1) // "##:##\0" for int values
```
ryanplusplus, line 27:
```suggestion
#define MAX_STR_LEN sizeof("##:##")
```

toy: Today I learned

ryanplusplus (approved):

exercises/practice/clock/.approaches/introduction.md
ryanplusplus: I just learned this the other day. Very useful.

exercises/practice/clock/.approaches/introduction.md
```diff
@@ -24,7 +24,7 @@ thus keeping the code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yoursel
 ```c
 #include <stdbool.h>
 
-#define MAX_STR_LEN 14 // "##:##\0" for int values
```
wolf99, line 27: This was previously a value of 14.
Now it's a value of 6 I think.
Is that correct?

ryanplusplus: I checked the tests to confirm and it looks like the comment was correct and the code was wrong. I'm not sure how we got to 14 🤷.

wolf99: Thanks @ryanplusplus .
Hard for me to check properly on mobile.
It seems this change is for one of the docs in the exercise. Should the exercise header stub be updated also?
(Currently it seems to use the `(5+1)` approach)

ryanplusplus: Good catch, it should also be updated.

toy: The dig-deeper/approaches variant had `14` since it was added in #874 and exercise had `5+1` since it was added in #90. Maybe @bobahop can remember the reason

Created #928 to fix other places

Issue #931, opened 2023-12-04
 The order of the results actually matters for the Word Count exercise


elmq0022:
The instructions state that the order of the unique words does not matter when in fact they do.  

https://github.com/exercism/c/blob/main/exercises/practice/word-count/.docs/instructions.md?plain=1#L21

The tests assume the word order is the order they are first encountered in.  

Either the instructions should be updated, or the expected and actual results could just be sorted before the check to ensure a standard order of comparison.  Personally, I would prefer the later option, and I could do some work in the future to make that happen if people are interested.

elmq0022:
I can update the tests to use the stdlib qsort. We just need a custom sort function for the struct with text/string and number of occurrences.  That shouldn't be too bad to implement.   TBH I just want to see some traction on the issue before doing the work.  I'll try to open a PR tonight.

elmq0022:
Sorry about the delay.  I'll try to look at any feedback on #932 as soon as possible.  Thankfully this is a rather small PR.

elmq0022:
closed by #932

PR #932, opened 2023-12-10
sorts solutions prior to check

elmq0022:
This PR addresses issue #931 by sorting the expected and actual lists of word_count_word_t prior to comparing them.

### DIFF
--- a/exercises/practice/word-count/test_word_count.c
+++ b/exercises/practice/word-count/test_word_count.c
@@ -17,8 +17,7 @@ void tearDown(void)
 {
 }
 
-
-int cmp_word_count_word(const void *a, const void *b){
+static int cmp_word_count_word(const void *a, const void *b){
     word_count_word_t *w1 = (word_count_word_t *)a;
     word_count_word_t *w2 = (word_count_word_t *)b;
     return strcmp(w1->text, w2->text);
--- a/exercises/practice/word-count/test_word_count.c
+++ b/exercises/practice/word-count/test_word_count.c
@@ -17,10 +17,11 @@ void tearDown(void)
 {
 }
 
-static int cmp_word_count_word(const void *a, const void *b){
-    word_count_word_t *w1 = (word_count_word_t *)a;
-    word_count_word_t *w2 = (word_count_word_t *)b;
-    return strcmp(w1->text, w2->text);
+static int cmp_word_count_word(const void *a, const void *b)
+{
+   word_count_word_t *w1 = (word_count_word_t *)a;
+   word_count_word_t *w2 = (word_count_word_t *)b;
+   return strcmp(w1->text, w2->text);
 }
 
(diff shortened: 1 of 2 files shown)


exercises/practice/word-count/word_count.h
```diff
@@ -26,4 +26,7 @@ typedef struct word_count_word {
 //           words will contain the results up to that point.
 int count_words(const char *sentence, word_count_word_t *words);
 
+// used to sort the words before comparison
+int cmp_word_count_word(const void *a, const void *b);
+
```
elmq0022, line 31: I am not sure if this definition should live here, or if I create a new header file for the test.

exercises/practice/word-count/test_word_count.c
```diff
@@ -17,6 +17,13 @@ void tearDown(void)
 {
 }
 
+
+int cmp_word_count_word(const void *a, const void *b){
```
ryanplusplus, line 21:
```suggestion
static int cmp_word_count_word(const void *a, const void *b){
```

exercises/practice/word-count/word_count.h
ryanplusplus: It would be best to leave it as a `static` function (basically a private function) in the test file. If you mark the definition private in `test_word_count.c` then you can remove this.

ryanplusplus:
```suggestion
```

exercises/practice/word-count/test_word_count.c
ryanplusplus, line 21:
```suggestion
static int cmp_word_count_word(const void *a, const void *b){
```

ryanplusplus (changes requested):

exercises/practice/word-count/test_word_count.c
```diff
@@ -17,6 +17,12 @@ void tearDown(void)
 {
 }
 
+static int cmp_word_count_word(const void *a, const void *b){
```
wolf99, line 20: *nit: this function could be named just `compare_word()`, the type of word is not super important to indicate because the function is private.

wolf99 (approved):

exercises/practice/word-count/test_word_count.c
elmq0022: Thanks, I've made the update.

ryanplusplus (approved):

elmq0022:
Yay!

ryanplusplus:
Thanks!

elmq0022:
How long until the site reflects this change?

ryanplusplus:
Not long. I'm pretty sure it begins deployment right after it hits `main`.

wolf99:
Tha ks for contributing @elmq0022 !

PR #939, opened 2024-01-21
Add used types to the pythagorean_triplet.h

vladimir-popov:
It would be better to see all used in the exercise types in the header file, than recover them from the source.


vladimir-popov:
It's duplicate of [743](https://github.com/exercism/c/pull/743)

Issue #940, opened 2024-01-22
Exercises for #48in24


siebenschlaefer:
Dear maintainers,

Jeremy posted an overview of the exercises that will be featured in #48in24:
http://forum.exercism.org/t/which-exercises-for-48in24/8970/19

The following featured exercises have not been translated to C yet:

- [ ] parallel-letter-frequency
- [ ] zebra-puzzle
- [ ] dnd-character
- [ ] yacht
- [ ] spiral-matrix
- [ ] kindergarten-garden
- [ ] knapsack
- [ ] rotational-cipher
- [ ] food-chain
- [ ] bank-account
- [ ] pop-count

I'm posting them here as a reminder.
I hope I will find the time to translate a few of them, and perhaps others will help, too.
Feel free to close or remove this issue if you do not find it helpful.

ryanplusplus:
Thanks for the heads up. I'll pick up some as I get time. If anyone would like to grab one, please leave a note here first so that no one else starts work on the same exercise.

ahans:
I will try to contribute kindergarten-garden.

ryanplusplus:
I'm going to grab spiral-matrix

ryanplusplus:
I'm also going to grab protein-translation

siebenschlaefer:
> I'm also going to grab protein-translation

[Somebodyon the forum](https://forum.exercism.org/t/c-protein-translation-exercise/9448) wanted to help with that. Although they didn't reply in the last two days.

ryanplusplus:
> > I'm also going to grab protein-translation
> 
> [Somebodyon the forum](https://forum.exercism.org/t/c-protein-translation-exercise/9448) wanted to help with that. Although they didn't reply in the last two days.

Thanks for the heads up. I'll give them another few days to respond.

ahans:
Opened a PR for `pop-count` as well since it was so trivial.

ryanplusplus:
I'm going to grab knapsack

siebenschlaefer:
I've given the zebra-puzzle a shot (#959)

ahans:
Put me down for yacht. I will probably get around to implement something next week.

ahans:
This was quicker than expected: https://github.com/exercism/c/pull/972

ahans:
I also gave rotational-cipher a shot: https://github.com/exercism/c/pull/975

ahans:
Feel free to also out me down for dnd-character and food-chain.

[...]

PR #942, opened 2024-01-27
Add spiral-matrix

ryanplusplus:
Related to https://github.com/exercism/c/issues/940


exercises/practice/spiral-matrix/test_spiral_matrix.c
```diff
@@ -0,0 +1,120 @@
...
+                           const spiral_matrix_t *actual)
+{
+   if (expected_size != actual->size)
+      return false;
+
+   for (int i = 0; i < expected_size; i++) {
+      for (int j = 0; j < expected_size; j++) {
+         if (expected_matrix[i][j] != actual->matrix[i][j]) {
+            printf("expected[%d][%d] = %d\n", i, j, expected_matrix[i][j]);
+            printf("actual[%d][%d] = %d\n", i, j, actual->matrix[i][j]);
```
ryanplusplus, line 26: Because the check is pretty opaque (it tells you whether it matches, but that's it), I added in these prints to show where the first difference is when there's a failure.

exercises/practice/spiral-matrix/test_spiral_matrix.c
```diff
@@ -0,0 +1,130 @@
...
+
+static void test_spiral_of_size_5(void)
+{
+   TEST_IGNORE();
+
+   int expected_matrix[5][5] = {
+      { 1, 2, 3, 4, 5 },       //
+      { 16, 17, 18, 19, 6 },   //
+      { 15, 24, 25, 20, 7 },   //
+      { 14, 23, 22, 21, 8 },   //
```
ryanplusplus, line 109: These `//`s prevent the formatter from wrapping these with more than one row on a line. I only need these when there are 5 rows.

exercises/practice/spiral-matrix/.meta/example.c
```diff
@@ -0,0 +1,50 @@
...
+   for (int i = 0; i < size; i++) {
+      spiral_matrix->matrix[i] = calloc(size, sizeof(int));
+   }
+
+   int x = 0;
+   int y = 0;
+   int dx = 1;
+   int dy = 0;
+
+   for (int i = 1; i <= size * size; i++) {
```
wolf99, line 23: `size_t i` maybe?

exercises/practice/spiral-matrix/.meta/example.c
```diff
@@ -0,0 +1,50 @@
...
+      x += dx;
+      y += dy;
+   }
+
+   return spiral_matrix;
+}
+
+void spiral_matrix_destroy(spiral_matrix_t *spiral_matrix)
+{
+   for (int i = 0; i < spiral_matrix->size; i++) {
```
wolf99, line 41: `size_t i` maybe?

exercises/practice/spiral-matrix/spiral_matrix.h
```diff
@@ -0,0 +1,4 @@
+#ifndef SPIRAL_MATRIX_H
+#define SPIRAL_MATRIX_H
+
```
wolf99, line 3: Given the recent feedbac in other issues, should we have at least the struct definition here?

exercises/practice/spiral-matrix/test_spiral_matrix.c
wolf99: How does this look when the tests run and have their standard output?
Would it interfere with how the web IDE detects pass/fail ?

exercises/practice/spiral-matrix/.meta/example.c
```diff
@@ -0,0 +1,50 @@
+#include <stdlib.h>
+#include "spiral_matrix.h"
+
+spiral_matrix_t *spiral_matrix_create(int size)
+{
+   spiral_matrix_t *spiral_matrix = calloc(1, sizeof(spiral_matrix_t));
```
wolf99, line 6: Small nit:
Consider refactoring type and variable names from `spiral_matrix` to `spiral` so that we then have a nice `spiral->matrix` during use?

wolf99 (commented):
Sorry that it took me so long to review this!
Just a few small nits

exercises/practice/spiral-matrix/.meta/example.c
ryanplusplus: Sure, good idea

exercises/practice/spiral-matrix/.meta/example.c
ryanplusplus: I started with `size_t`, actually. The reason I changed to `int` is that I want to compare to `x + dx` and `y + dy` to `size` with a minimum of awkwardness and with `size_t` there's a signedness conflict. Casting `x + dx` to `size_t` isn't safe because `size_t` can't represent when `x + dx` is negative, casting `size` up to a larger signed type isn't an option, and casting `size` down to an `int` isn't much different than using `int` in the first place. I would use `ssize_t`, but this isn't included in standard C. C++ has standardized `std::ssize` and widespread use of unsigned size types is considered to have been a mistake in the C++ standard library.

In other words, I agree with you in principle, but this felt like the best of the bad options and least likely to cause awkward situations for students.

exercises/practice/spiral-matrix/spiral_matrix.h
ryanplusplus: Sure, I will add it.

exercises/practice/spiral-matrix/test_spiral_matrix.c
ryanplusplus: With this, the output for a failure is:

```
expected[0][0] = 1
actual[0][0] = 0
./test_spiral_matrix.c:52:test_trivial_spiral:FAIL: Expected TRUE Was FALSE
```

Without it, the output is:

```
./test_spiral_matrix.c:52:test_trivial_spiral:FAIL: Expected TRUE Was FALSE
```

>  Would it interfere with how the web IDE detects pass/fail ?

I don't think it will since the failure is still present, but I'm not sure whether it will show the console output in addition to the failure text.

Looking through the Unity docs I think I should be able to compare each row in the array. This should give enough context to understand where the failure is because each row is unique.

ryanplusplus: With Unity's array matcher, the output is:

```
./test_spiral_matrix.c:22:test_spiral_of_size_5:FAIL: Element 2 Expected 25 Was 0
```

I think that's sufficient to work out where the error is even without knowing the row number.

ryanplusplus:
> Sorry that it took me so long to review this! Just a few small nits

No worries, thanks for the review!

wolf99 (approved):
LGTM 👍

PR #943, opened 2024-01-27
Add kindergarten-garden (#940)

ahans:
This adds `kindergarten-garden` from #940's list.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2,14 +2,14 @@
   "language": "C",
   "slug": "c",
   "active": true,
-  "blurb": "C is a small, general-purpose, imperative programming language with a static type system, scopes, and structures. It's typically used as an alternative to assembly programming, such as in operating systems.",
-  "version": 3,
   "status": {
     "concept_exercises": false,
     "test_runner": true,
     "representer": false,
     "analyzer": false
   },
+  "blurb": "C is a small, general-purpose, imperative programming language with a static type system, scopes, and structures. It's typically used as an alternative to assembly programming, such as in operating systems.",
+  "version": 3,
   "online_editor": {
     "indent_style": "space",
     "indent_size": 4,
@@ -36,7 +36,6 @@
     ]
   },
   "exercises": {
-    "concept": [],
     "practice": [
       {
         "slug": "hello-world",
@@ -1015,6 +1014,19 @@
           "performance_optimizations",
           "strings"
         ]
+      },
+      {
+        "slug": "kindergarten-garden",
+        "name": "Kindergarten Garden",
+        "uuid": "9484317e-d638-4714-afba-2a10baccff16",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2,
+        "topics": [
+          "arrays",
+          "control_flow_loops",
+          "strings"
+        ]        
       }
     ]
   },
@@ -1024,7 +1036,7 @@
       "slug": "basics",
       "name": "Basics"
     },
-    {    
+    {
       "uuid": "20948c69-20f7-498d-b743-fada00e5c79d",
       "slug": "bits",
       "name": "Bits"
@@ -1043,7 +1055,7 @@
       "uuid": "1721552b-9c55-432d-a776-05e618c25bb4",
       "slug": "function-pointers",
       "name": "Function Pointers"
-    },    
+    },
     {
       "uuid": "e5725455-3ce3-48e9-a3c6-ceb522434b5b",
       "slug": "conditionals",
@@ -1058,7 +1070,7 @@
       "uuid": "e39f5003-8ed9-4d82-ab9b-f79bbca78679",
       "slug": "linkage",
       "name": "Linkage"
-    },    
+    },
     {
       "uuid": "3d79388b-e909-4e1b-99e9-d28e30e54fde",
       "slug": "loops",
@@ -1083,7 +1095,7 @@
       "uuid": "f4946fe8-afb3-4110-8fca-b1476f19be6f",
       "slug": "static-functions",
       "name": "Static Functions"
-    },    
+    },
     {
       "uuid": "c065ddca-627c-45b6-a92f-7144dd852111",
       "slug": "storage-class-specifiers",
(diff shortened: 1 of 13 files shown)


config.json
```diff
@@ -2,14 +2,14 @@
...
   "active": true,
-  "blurb": "C is a small, general-purpose, imperative programming language with a static type system, scopes, and structures. It's typically used as an alternative to assembly programming, such as in operating systems.",
-  "version": 3,
   "status": {
     "concept_exercises": false,
     "test_runner": true,
     "representer": false,
     "analyzer": false
   },
+  "blurb": "C is a small, general-purpose, imperative programming language with a static type system, scopes, and structures. It's typically used as an alternative to assembly programming, such as in operating systems.",
```
ahans, line 11: `bin/configlet create` made those extra changes. Not sure we want to keep them as part of this PR. I can for sure revert them and only keep the addition of the entry for this exercise.

exercises/practice/kindergarten-garden/.meta/example.c
```diff
@@ -0,0 +1,46 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static int student_to_index(const char *student)
+{
+   static const char *students[] = { "Alice",  "Bob",    "Charlie", "David",
+                                     "Eve",    "Fred",   "Ginny",   "Harriet",
+                                     "Ileana", "Joseph", "Kincaid", "Larry" };
+   for (size_t i = 0; i < sizeof(students); ++i) {
```
siebenschlaefer, line 12: `sizeof` gives you the size of `students` in bytes, not the number of elements.

exercises/practice/kindergarten-garden/test_kindergarten_garden.c
```diff
@@ -0,0 +1,164 @@
...
+
+static void test_partial_garden_garden_with_single_student(void)
+{
+   const plant_t expected[] = { RADISHES, CLOVER, GRASS, GRASS };
+   test_kindergarten_garden("RC\nGG", "Alice", expected);
+}
+
+static void test_partial_garden_different_garden_with_single_student(void)
+{
+   const plant_t expected[] = { VIOLETS, CLOVER, RADISHES, CLOVER };
```
siebenschlaefer, line 30: In the tests for almost all other exercises there is a line TEST_IGNORE(); (except in the first test), with an explaining comment in the second test.
see https://github.com/exercism/c/pull/944#discussion_r1468699220

exercises/practice/kindergarten-garden/test_kindergarten_garden.c
```diff
@@ -0,0 +1,164 @@
...
+   test_kindergarten_garden("VC\nRC", "Alice", expected);
+}
+
+static void test_partial_garden_garden_with_two_students(void)
+{
+   const plant_t expected[] = { CLOVER, GRASS, RADISHES, CLOVER };
+   test_kindergarten_garden("VVCG\nVVRC", "Bob", expected);
```
siebenschlaefer, line 37: You could consider formatting the "diagram" in the tests in two lines like this:
```c
   test_kindergarten_garden(
         "VVCG\n"
         "VVRC",
         "Bob", expected);
```
to make the tests a little bit easier to read.

exercises/practice/kindergarten-garden/.meta/example.c
ryanplusplus:
```suggestion
   for (size_t i = 0; i < sizeof(students)/sizeof(students[0]); ++i) {
```

exercises/practice/kindergarten-garden/kindergarten_garden.c
ryanplusplus, line 1: We usually have the `#include` for the header in the provided C file

exercises/practice/kindergarten-garden/test_kindergarten_garden.c
```diff
@@ -0,0 +1,164 @@
...
+
+void tearDown(void)
+{
+}
+
+static void test_kindergarten_garden(const char *diagram, const char *student,
+                                     const plant_t *expected)
+{
+   plant_t actual[4];
+   plants(diagram, student, actual);
```
ryanplusplus, line 17: What if you just returned struct containing an array of size 4? Since it's a known size it's strange to require it to be allocated by the client and then passed as an output parameter.

exercises/practice/kindergarten-garden/test_kindergarten_garden.c
```diff
@@ -0,0 +1,164 @@
...
+static void test_full_garden_for_larry_last_student_s_garden(void)
+{
+   const plant_t expected[] = { GRASS, VIOLETS, CLOVER, VIOLETS };
+   test_kindergarten_garden(
+       "VRCGVVRVCGGCCGVRGCVCGCGV\nVRCCCGCRRGVCGCRVVCVGCGCV", "Larry", expected);
+}
+
+int main(void)
+{
+   UnityBegin("test_kindergarten-garden.c");
```
ryanplusplus, line 143:
```suggestion
   UnityBegin("test_kindergarten_garden.c");
```

siebenschlaefer: Question for the maintainers: Why do we use `UnityBegin` with an explicit filename and `UnityEnd` instead of `UNITY_BEGIN()` without the filename and `UNITY_END()`?
(see https://github.com/ThrowTheSwitch/Unity/blob/master/docs/UnityConfigurationGuide.md#main)

ryanplusplus: This is the only project where I've used Unity and it was done like this when I became a maintainer of the track. In other words, I have no idea, and I didn't realize that there was another option.

exercises/practice/kindergarten-garden/.meta/example.c
ahans: Thanks for the suggestion @ryanplusplus , that's exactly what I had in mind. Just too little practice with C, I actually _know_ what `sizeof` does ....

exercises/practice/kindergarten-garden/test_kindergarten_garden.c
ahans: I fixed the name now. You want me to remove it? Another PR could change the existing exercises to make things consistent.

ahans:
I hope I addressed all comments now.

exercises/practice/kindergarten-garden/test_kindergarten_garden.c
ahans: Makes sense, although I'm not entirely happy with the naming. I wanted to return a fixed size array at first, but that doesn't work. Wrapping it in a struct is a good workaround!

exercises/practice/kindergarten-garden/test_kindergarten_garden.c
ryanplusplus: @ahans another PR to change this would be great.

ryanplusplus (approved):

PR #944, opened 2024-01-27
Add pop-count (#940)

ahans:
Implements one of the missing exercises from #940.

### DIFF
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -28,7 +28,7 @@ execute_test () {
         echo "Running tests for ${exercise_name}";
 
         #remove the ignore line
-        sed -i -e 's/TEST_IGNORE();//' ./test_*.c
+        sed -i 's/TEST_IGNORE();//' ./test_*.c
 
         # Copy the examples with the correct name for the exercise
         if [ -e ".meta/example.c" ]
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -28,7 +28,7 @@ execute_test () {
         echo "Running tests for ${exercise_name}";
 
         #remove the ignore line
-        sed -i 's/TEST_IGNORE();//' ./test_*.c
+        sed -i -e 's/TEST_IGNORE();//' ./test_*.c
 
         # Copy the examples with the correct name for the exercise
         if [ -e ".meta/example.c" ]
--- a/config.json
+++ b/config.json
@@ -2,14 +2,14 @@
   "language": "C",
   "slug": "c",
   "active": true,
-  "blurb": "C is a small, general-purpose, imperative programming language with a static type system, scopes, and structures. It's typically used as an alternative to assembly programming, such as in operating systems.",
-  "version": 3,
   "status": {
     "concept_exercises": false,
     "test_runner": true,
     "representer": false,
     "analyzer": false
   },
+  "blurb": "C is a small, general-purpose, imperative programming language with a static type system, scopes, and structures. It's typically used as an alternative to assembly programming, such as in operating systems.",
+  "version": 3,
   "online_editor": {
     "indent_style": "space",
     "indent_size": 4,
@@ -36,7 +36,6 @@
     ]
   },
   "exercises": {
-    "concept": [],
     "practice": [
       {
         "slug": "hello-world",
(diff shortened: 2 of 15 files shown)


config.json
```diff
@@ -2,14 +2,14 @@
   "language": "C",
   "slug": "c",
   "active": true,
-  "blurb": "C is a small, general-purpose, imperative programming language with a static type system, scopes, and structures. It's typically used as an alternative to assembly programming, such as in operating systems.",
```
ahans, line 5: Same as for the other PR applies: This was modified by `configlet create`. Let me know if I should revert it or put it into a different PR.

exercises/practice/pop-count/test_pop_count.c
```diff
@@ -0,0 +1,46 @@
...
+
+static void test_0_eggs(void)
+{
+   const int expected = 0;
+   TEST_ASSERT_EQUAL_INT(expected, egg_count(0));
+}
+
+static void test_1_eggs(void)
+{
+   const int expected = 1;
```
siebenschlaefer, line 20: In the tests for almost all other exercises there is a line `TEST_IGNORE();` (except in the first test), with an explaining comment in the second test.
see https://github.com/exercism/c/blob/main/exercises/practice/grains/test_grains.c#L19
That allows the tests to be enabled one by one so that the student can focus on one test at a time, similarly to Test-Driven Development.

exercises/practice/pop-count/test_pop_count.c
```diff
@@ -0,0 +1,46 @@
...
+
+static void test_13_eggs(void)
+{
+   const int expected = 13;
+   TEST_ASSERT_EQUAL_INT(expected, egg_count(2000000000));
+}
+
+int main(void)
+{
+   UnityBegin("test_prime_factors.c");
```
siebenschlaefer, line 38: This should be `"test_pop_count.c"`

siebenschlaefer (commented):
*Personally* I would prefer an unsigned type for the argument because the behavior of right-shifting a negative number is implementation-defined.
But that's probably a subjective choice to some degree.

ryanplusplus:
> _Personally_ I would prefer an unsigned type for the argument because the behavior of right-shifting a negative number is implementation-defined. But that's probably a subjective choice to some degree.

I agree -- it should be unsigned.

exercises/practice/pop-count/test_pop_count.c
ahans: I was wondering about all of those `TEST_IGNORE()`, the explanation makes sense. Adapted it now.

ahans:
> _Personally_ I would prefer an unsigned type for the argument because the behavior of right-shifting a negative number is implementation-defined. But that's probably a subjective choice to some degree.

If you pass in only positive numbers, it doesn't make a difference. But you're right of course, unsigned is cleaner here. Adapted it now.

ahans:
I hope I addressed all comments, would be great if you could have another look!

bin/run-tests
```diff
@@ -28,7 +28,7 @@ execute_test () {
         echo "Running tests for ${exercise_name}";
 
         #remove the ignore line
-        sed -i 's/TEST_IGNORE();//' ./test_*.c
+        sed -i -e 's/TEST_IGNORE();//' ./test_*.c
```
ryanplusplus, line 31: What's the reason for this change?

exercises/practice/pop-count/test_pop_count.c
```diff
@@ -0,0 +1,50 @@
...
+   const unsigned int expected = 13;
+   TEST_ASSERT_EQUAL_INT(expected, egg_count(2000000000));
+}
+
+int main(void)
+{
+   UnityBegin("test_pop_count.c");
+
+   RUN_TEST(test_0_eggs);
+
```
ryanplusplus, line 44:
```suggestion
```

bin/run-tests
ryanplusplus: I'm going to revert this and merge. If you think this change is necessary, please open another PR.

```suggestion
        sed -i 's/TEST_IGNORE();//' ./test_*.c
```

ahans: Thanks for cleaning that up! I committed that accidentally, it's required for BSD's `sed` (I'm partially on macOS). But there are apparently more issues with that script that I wasn't able to fix yet. I've been meaning to do a separate PR with those fixes.

PR #945, opened 2024-01-28
Add knapsack

ryanplusplus:
Related to https://github.com/exercism/c/issues/940

### DIFF
--- a/config.json
+++ b/config.json
@@ -1026,7 +1026,7 @@
           "arrays",
           "control_flow_loops",
           "strings"
-        ]        
+        ]
       },
       {
         "slug": "pop-count",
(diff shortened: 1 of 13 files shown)


siebenschlaefer (approved):
Just in case you were waiting for a review: Looks great!
(I don't think I have the power to approve here on the C track.)

The only tiny suggestion from me is this:
In the first test you could consider defining variables `items` and `item_count`, just like in the other tests, because the students have to extract the signature and semantics of the function `maximum()` from the tests and I think that's easier with `maximum(maximum_weight, items, item_count)` than with `maximum(maximum_weight, NULL, 0)`.

ryanplusplus:
> Just in case you were waiting for a review: Looks great! (I don't think I have the power to approve here on the C track.)
> 
> The only tiny suggestion from me is this: In the first test you could consider defining variables `items` and `item_count`, just like in the other tests, because the students have to extract the signature and semantics of the function `maximum()` from the tests and I think that's easier with `maximum(maximum_weight, items, item_count)` than with `maximum(maximum_weight, NULL, 0)`.

Good point, I'll change that.

ryanplusplus:
> Just in case you were waiting for a review: Looks great! (I don't think I have the power to approve here on the C track.)

Thanks for the review!

wolf99 (approved):

PR #948, opened 2024-01-28
Fix run-test for macOS standard tools

ahans:
BSD sed requires an argument with `-i` (backup file extension, disabling
backup file creation by providing an empty string; this is the default for GNU
sed when `-i` is specified without an argument).

Further, `/bin/bash` shipping with macOS 14 doesn't like `${*}` when the
argument list is empty (at least in the context it is used in `run-tests`).
The fix of having a distinct message for the case where `${*}` is empty also
improves the behavior for cases where a version of Bash is used that does not
complain about an empty `${*}`.

This has been tested on macOS Sonoma 14.3, but probably also applies to earlier
versions.

### DIFF
--- a/bin/run-tests
+++ b/bin/run-tests
@@ -28,7 +28,7 @@ execute_test () {
         echo "Running tests for ${exercise_name}";
 
         #remove the ignore line
-        sed -i 's/TEST_IGNORE();//' ./test_*.c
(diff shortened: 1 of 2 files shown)


ryanplusplus (approved):
Thanks!

PR #949, opened 2024-01-29
Return `int` in pop-count example

ryanplusplus:
I noticed this when doing the exercise. Our example returns an `unsigned int`, but the tests check for `int`.


ahans:
Is the issue that the test uses `TEST_ASSERT_EQUAL_INT` instead of `TEST_ASSERT_EQUAL_UINT`? In that case I think it'd be much better to change the test. Requiring students to implement their function returning `int` would be quite unexpected, given that in the tests `expected` is of type `unsigned int`. Since this exercise was only published yesterday, there are probably not many solutions that would require retesting.

ryanplusplus:
> Is the issue that the test uses `TEST_ASSERT_EQUAL_INT` instead of `TEST_ASSERT_EQUAL_UINT`? In that case I think it'd be much better to change the test. Requiring students to implement their function returning `int` would be quite unexpected, given that in the tests `expected` is of type `unsigned int`. Since this exercise was only published yesterday, there are probably not many solutions that would require retesting.

Yep, good point. I'll switch back to my initial approach.

wolf99 (approved):

PR #959, opened 2024-02-25
add exercise zebra-puzzle

siebenschlaefer:
Feel free to criticize anything, big or small.


exercises/practice/zebra-puzzle/zebra_puzzle.h
```diff
@@ -0,0 +1,10 @@
+#ifndef ZEBRA_PUZZLE_H
+#define ZEBRA_PUZZLE_H
+
+/// Determine the nationality of the resident who drinks water.
+const char *drinks_water(void);
+
+/// Determine the nationality of the resident who owns the zebra.
+const char *owns_zebra(void);
+
+#endif // ZEBRA_PUZZLE_H
```
ryanplusplus, line 10:
```suggestion
#endif
```

Removed for consistency with other exercises

config.json
```diff
@@ -1050,6 +1050,14 @@
...
       },
       {
+        "slug": "zebra-puzzle",
+        "name": "Zebra Puzzle",
+        "uuid": "2cae2796-5c05-4159-b3e7-9f5653f0bc9b",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 6
+      },
+      }
```
ryanplusplus, line 1060:
```suggestion
      {
```

My bad, I messed this up fixing the conflict

ryanplusplus (approved):
This is really slick! This is my first time seeing a solution to this exercise.

exercises/practice/zebra-puzzle/.meta/example.c
```diff
@@ -0,0 +1,203 @@
...
+   switch (owns_zebra) {
+      case ENGLISH:
+         *out_owns_zebra = "English";
+         break;
+      case JAPANESE:
+         *out_owns_zebra = "Japanese";
+         break;
+      case NORWEGIAN:
+         *out_owns_zebra = "Norwegian";
+         break;
```
wolf99, line 179: Why solve the same problem twice?
Is it possible to solve it once and then just query the results (possibly in some generic way) ?

wolf99 (commented):
I feel like there may be some way to implement a large part of this with x macros. But I can't see it well enough on mobile to know for sure

exercises/practice/zebra-puzzle/.meta/example.c
ryanplusplus: Do you mean to take an object oriented approach to the exercise where solving and querying the solution are separate steps?

siebenschlaefer: I agree, and don't like it either.  
I was under the impression that these two separate functions would be a what the author of the [problem-specification](https://github.com/exercism/problem-specifications/blob/main/exercises/zebra-puzzle/canonical-data.json) intended, with the two "properties" `drinksWater` and `ownsZebra`.  
I looked at the [C# version](https://github.com/exercism/csharp/blob/main/exercises/practice/zebra-puzzle/ZebraPuzzle.cs) where those are two `static` methods and the [Python version](https://github.com/exercism/python/blob/main/exercises/practice/zebra-puzzle/zebra_puzzle.py) where those are two free functions.

But now that you asked I dug a little further and found the [Go version](https://github.com/exercism/go/blob/main/exercises/practice/zebra-puzzle/zebra_puzzle.go) that has a single function `SolvePuzzle()` which I like more.

I could change the code to something like
```c
struct solution
{
   const char *drinks_water;
   const char *owns_zebra;
};

struct solution solve_puzzle(void);
```

PR #960, opened 2024-02-28
Improve the zebra-puzzle.

siebenschlaefer:
Following the discussion at <https://github.com/exercism/c/pull/959#discussion_r1501899860>, this PR changes the API of the solution. Instead of two formally independent functions for the two sub-tasks the tests now require a single function that returns a `struct` with the two answers. That should reduce the boilerplate that students will have to write.

It also improves some small details of the example solution.

### DIFF
--- a/exercises/practice/zebra-puzzle/zebra_puzzle.h
+++ b/exercises/practice/zebra-puzzle/zebra_puzzle.h
@@ -1,10 +1,12 @@
 #ifndef ZEBRA_PUZZLE_H
 #define ZEBRA_PUZZLE_H
 
-/// Determine the nationality of the resident who drinks water.
-const char *drinks_water(void);
+struct solution
+{
+   const char *drinks_water;
+   const char *owns_zebra;
+};
 
-/// Determine the nationality of the resident who owns the zebra.
-const char *owns_zebra(void);
+struct solution solve_puzzle(void);
 
 #endif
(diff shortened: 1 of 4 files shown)


exercises/practice/zebra-puzzle/zebra_puzzle.h
```diff
@@ -1,10 +1,11 @@
 #ifndef ZEBRA_PUZZLE_H
 #define ZEBRA_PUZZLE_H
 
-/// Determine the nationality of the resident who drinks water.
-const char *drinks_water(void);
+struct solution {
+   const char *drinks_water;
+   const char *owns_zebra;
+};
```
ryanplusplus, line 7:
```suggestion
typedef struct {
   const char *drinks_water;
   const char *owns_zebra;
} solution_t;
```

I think this would be more consistent with how we've handled structs elsewhere in the track

exercises/practice/zebra-puzzle/zebra_puzzle.h
```diff
@@ -1,10 +1,11 @@
 #ifndef ZEBRA_PUZZLE_H
 #define ZEBRA_PUZZLE_H
 
-/// Determine the nationality of the resident who drinks water.
-const char *drinks_water(void);
+typedef struct {
+   const char *drinks_water;
+   const char *owns_zebra;
+} solution_t;
```
ryanplusplus, line 7: Can you update the type elsewhere? Now this doesn't match the prototype below or the tests.

siebenschlaefer: Sorry

ryanplusplus: No worries!

ryanplusplus (approved):
Thanks!

PR #969, opened 2024-04-04
Amend the return type of function in `raindrop`

### LINKED ISSUE #968: Return type for `convert(char result[], int drops)` in "Raindrop" should be of type `void`
Since the test cases do not make use of returned value, and it is clear from the overview that the argument is used to pass the string by reference, it makes sense to use `void` instead. Learners might be confuse this behavior.

Ashvith10:
Return type is amended from `char*` to `void` for the function `convert(char result[], int drops)`

Fixes #968

### DIFF
--- a/exercises/practice/raindrops/.approaches/data-driven/content.md
+++ b/exercises/practice/raindrops/.approaches/data-driven/content.md
@@ -6,7 +6,7 @@
 #ifndef RAINDROPS_H
 #define RAINDROPS_H
 
-char *convert(char result[], int drops);
+void convert(char result[], int drops);
 
 #endif
 ```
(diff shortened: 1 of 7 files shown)


wolf99 (changes requested):
Good for this file, but changes will be needed in the header and source in the exercises `.meta` directory also.

Ashvith10:
@wolf99 my bad. Looks like there's a whole lot of stuff to change - but since the `ls` command in my GuixSD instance wasn't configured with hidden files, I missed out on that. I'll be pushing out the new changes soon.

Ashvith10:
I tried checking out for all instances of incorrect definitions using `grep -nr "char *convert(.*)"`. Hopefully, that should check for all the strings in the directory recursively.

ryanplusplus (approved):

wolf99 (approved):
LGTM, let's merge!

PR #972, opened 2024-04-13
Add exercise yacht

ahans:


ahans:
I will obviously fix the GCC warning.

exercises/practice/yacht/test_yacht.c
```diff
@@ -0,0 +1,250 @@
...
+   TEST_IGNORE();
+   const dice_t dice = { 2, 3, 4, 5, 6 };
+   TEST_ASSERT(score(dice, SIXES) == 6);
+}
+
+static void test_full_house_two_small_three_big(void)
+{
+   TEST_IGNORE();
+   const dice_t dice = { 2, 2, 4, 4, 4 };
+   TEST_ASSERT(score(dice, FULL_HOUSE) == 16);
```
wolf99, line 93: For the test we tend to use the type specific test macros from Unity rather than TEST_ASSERT(). Can you check if there are any suitable for this exercise?

exercises/practice/yacht/.meta/example.c
```diff
@@ -0,0 +1,108 @@
...
+
+typedef struct {
+   uint8_t faces[6];
+} count_t;
+
+static count_t count(const dice_t dice)
+{
+   count_t counts = { 0 };
+   for (int i = 0; i < 5; ++i) {
+      assert(dice.faces[i] >= 1 && dice.faces[i] <= 6);
```
wolf99, line 25: I'm not sure that we've used assert() in an exercise before. Do you know what kind of output or results this shows the user when running the tests?

Normally we tend to try and output some error value, though I don't know if there is a good way to do that in this exercise?

wolf99 (changes requested):
Mostly good stuff, a few small items to look at

exercises/practice/yacht/.meta/example.c
ryanplusplus: On other tracks it's assumed that inputs are not invalid. This isn't production so I like to try to keep these exercises focused and not worry about making them bulletproof against all types of misuse.

exercises/practice/yacht/test_yacht.c
```diff
@@ -0,0 +1,250 @@
...
+
+static void test_yacht(void)
+{
+   const dice_t dice = { 5, 5, 5, 5, 5 };
+   TEST_ASSERT(score(dice, YACHT) == 50);
+}
+
+static void test_not_yacht(void)
+{
+   TEST_IGNORE();
```
siebenschlaefer, line 21: In most exercises there's a comment in this first `TEST_IGNORE()` line that explains that this should be deleted to enable the tests, e.g. in [test_difference_of_squares.c](https://github.com/exercism/c/blob/main/exercises/practice/difference-of-squares/test_difference_of_squares.c#L19)
Do you mind adding such comment?

exercises/practice/yacht/.meta/config.json
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    "example": [
+      ".meta/example.c",
+      ".meta/example.h"
+    ]
+  },
+  "blurb": "Score a single throw of dice in the game Yacht.",
+  "source": "James Kilfiger, using wikipedia",
+  "source_url": "https://en.wikipedia.org/wiki/Yacht_(dice_game)"
+}
```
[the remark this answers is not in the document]
ahans: Hm, looks like auto-format on save removes it for some reason. It's back now. Sorry for this, I was under the impression that's the way how `configlet` creates the file and it's how you want it to be.

exercises/practice/yacht/.meta/example.c
ahans: This is not the test, but the example implementation. I'm using `i - 1` as an index further down, so this is also supposed to be sort of a comment reminding the reader that the faces go from 1 to 6, while we use indices 0 to 5 for them. I would find an `assert` good practice here and that's why I put it. If you insist, I can of course remove it.

ahans: > Do you know what kind of output or results this shows the user when running the tests?

Just to be really clear, when a student does the exercise, they will not interact with this file in any way.

exercises/practice/yacht/test_yacht.c
ahans: Done!

config.json
```diff
@@ -1077,6 +1077,20 @@
...
           "pointers"
         ]
+      },
+      {
+        "slug": "yacht",
+        "name": "Yacht",
+        "uuid": "dd360788-1dff-4a5d-89af-5bbd1199ae86",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2,
```
ryanplusplus, line 1087: How about a 4? We have Eliud's eggs rated as a 2 and that's a fair bit less complicated.

```suggestion
        "difficulty": 4,
```

exercises/practice/yacht/.meta/example.c
```diff
@@ -0,0 +1,108 @@
...
+   int sum = 0;
+   for (int i = 0; i < 5; ++i) {
+      if (dice.faces[i] == x)
+         sum += x;
+   }
+   return sum;
+}
+
+typedef struct {
+   uint8_t faces[6];
```
ryanplusplus, line 18:
```suggestion
   int faces[6];
```

There's definitely nothing wrong with a fixed sized integer, but it's inconsistent with the typing of the rest of the solution. If we change this to `int` we should be able to remove `stdint.h` above.

exercises/practice/yacht/.meta/example.c
```diff
@@ -0,0 +1,108 @@
+#include "yacht.h"
+
+#include <assert.h>
+#include <stdbool.h>
```
ryanplusplus, line 4:
```suggestion
```

I think this is unused

exercises/practice/yacht/.meta/example.c
ryanplusplus: If you'd rather use fixed-width types consistently, that's also fine with me.

ryanplusplus (changes requested):

config.json
ahans: Whatever you find reasonable. Difficulties are all over the place. `spiral-matrix` is also a 4 and I would say that's more difficult than this. I updated it to 4 now.

ryanplusplus: Agreed, many of these need to be revisited

exercises/practice/yacht/.meta/example.c
ahans: Where exactly is this inconsistent? I use `uint8_t` everywhere except for the final score. Technically, that could also be a `uint8_t`, but some small change in the specification (let's say yacht is is not 50, but 300 points) would break this.

My main motivation here, though, was efficiency. 6 bytes will easily be passed via a register when doing function calls. 24 bytes could be a different story.

I can change it to `int` if you prefer. But I'd keep the `uint8_t` here.

exercises/practice/yacht/.meta/example.c
ahans: A leftover from when `has_count` returned a `bool`. Removed that now.

exercises/practice/yacht/.meta/example.c
ryanplusplus: There are several places in this that use `int` instead of a fixed-width type. I understand your reasons for making the choice you made, but this is a learning aid, not production code. This optimization will have little meaning to someone scoring a dice game to learn the fundamentals of C. My thought is that seeing a mix of fixed-width and non-fixed-width types will be more surprising to someone solving the problem than a missed optimization.

exercises/practice/yacht/.meta/example.h
```diff
@@ -0,0 +1,27 @@
+#ifndef YACHT_H
+#define YACHT_H
+
+#include <stdint.h>
```
wolf99, line 4: I think this include can probably be removed now?
Then I think this PR is good to go!

wolf99 (changes requested):

exercises/practice/yacht/yacht.h
```diff
@@ -0,0 +1,27 @@
...
+   BIG_STRAIGHT,
+   CHOICE,
+   YACHT
+} category_t;
+
+typedef struct {
+   uint8_t faces[5];
+} dice_t;
+
+int score(dice_t dice, category_t category);
```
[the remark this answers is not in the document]
ahans: In the structs I care about the size (I want them to be as small as possible). Where I use `int`, I'll gladly accept whatever's the platform's default (it will be at least `int16_t`, so good enough). I do find that consistent. But I'm following @ryanplusplus's argument that this is too surprising for students. I changed it to `int` now. Although I do think that there should not be a hard distinction between "production" and "practice" code. Code quality matters, always. However, this is a micro-optimization that even in production code could lead to some debates. It would depend on the context if the performance benefit warranted the increased potential for confusion. Here, they probably do not.

exercises/practice/yacht/.meta/example.h
ahans: Ah, right! Hadn't had a chance to look at the diff myself yet! This one is gone now as well.

wolf99 (approved):

wolf99:
LGTM, well done @ahans !

ryanplusplus (approved):

ryanplusplus:
Thanks!

Issue #971, opened 2024-04-09
Jumped right from Hello World to Armstrong Numbers?


dotrosedotnet:
Did I miss something? This is a massive jump, and left me entirely clueless. I tried to build a solution using google, but it was just so far ahead of anything that I'd encountered that I think I'm going to give up with this one and go back to using a coding book. I have encountered this problem with a few of the tracks, where it jumps levels without building information incrementally.

For clarity: I *love* that I get to learn using my own tooling on my own computer in Exercism. I am wildly grateful that you offer that functionality. Thank you. Don't stop. <3

wolf99:
Seems like a fair comment!

The order on the website follows the older of the exercises listed in the tracks `config.json`.

However, I think it is a long while since we considered the orderring there. Any thoughts @ryanplusplus ?

@ErikSchierboom are there any resources or suggestions to assist tracks to know how to order exercises?

ErikSchierboom:
> @ErikSchierboom are there any resources or suggestions to assist tracks to know how to order exercises?

Not really no. Most maintainers have traditionally used their gut feeling, which is admittedly not much in terms of guidance. Looking at the difficulties, it might be that the difficulty of `armstrong-numbers` could be bumped a bit, as it is currently the lowest difficulty.

https://exercism.org/tracks/c/build can be really helpful to identify problems too. If you go to the practice exercises section, you'll quickly see that the completion rate of Armstrong Numbers is very low compared to other exercises. Same goes for Isogram.

siebenschlaefer:
We discussed the order shortly in issue #740 but did not follow up on that.

dotrosedotnet:
Y’all are the bestOn Apr 10, 2024, at 9:16 AM, Matthias ***@***.***> wrote:﻿
We discussed the order shortly in issue #740 but did not follow up on that.

ryanplusplus:
@wolf99 I think this is definitely worth reworking. The resource that @ErikSchierboom shared looks like it has exactly the info we need to do this correctly.

Issue #911, opened 2023-08-13
More documentation to `Grade School`


MathieuSoysal:
## Introduction

Currently, I'm doing the exercise [Grade School](https://exercism.org/tracks/c/exercises/grade-school).

## Question 

I wanted to know why we don't have a method inside [grade_school.h](https://github.com/exercism/c/blob/main/exercises/practice/grade-school/grade_school.h)?

siebenschlaefer:
This was done on purpose.  

Here at Exercism the instructions give some motivation and describe the task from a high-level perspective. All the details, the exact requirements, are specified by the tests. Reading the tests is part of the task.  
For *some* tasks the `.h` is provided completely or (like in this case) just the data structures but in general it shouldn't be hard to extract the required data structures and functions from the tests.

Concrete: Did you find it difficult to read the tests and understand what you're supposed to do?

Whether a complete `.h` file should be provided or not has been discussed a few times in the past. The maintainers decided that they like the current state the most. If you want you can try to convince them otherwise.

Issue #740, opened 2022-01-16
Inconsistency in headers provided


joshgoebel:
Most exercises provide at least the header to get you started (so you know what API you are implementing)... but some give you *nothing*.  Example: `circular_buffer`... this makes it much harder to get started.

Is this somehow intentional?  Could we agree on what amount of boilerplate exercises should start with and then would PRs be welcome here?

ryanplusplus:
When we've discussed this in the past, the goal was to use empty headers as part of the exercise. Of course, the goal is to create a learning opportunity, not put up a brick wall. I don't mentor the C track so I don't have a good feel for where people are struggling and it's entirely possible that we've missed the mark in what we provide for some of the exercises.

PRs are welcome.

wolf99:
I think we previously discussed if we should add Stubb's and decided only to have them for the earlier exercises.

Open to discuss the idea again though.

I raised the same concern on the PR and Josh validly pointed out that this policy is not apparent in the implementation.

IIRC the main stumbling block that we had before was that for some exercises providing the stubs "gave the game away". With concepts starting Ng to be created perhaps this concern is less of an issue?

joshgoebel:
> providing the stubs "gave the game away". 

I feel like I didn't include stubs where the stubs were the whole exercise... (resistor enums, etc)... but most of the time it's just the headers and then you still have to implement the actual algorithm... so I guess the question is how hard are we trying to make it...

bobahop:
Having done 44 solutions myself and having mentored 200+ solutions, I can't say I recall the lack of forward declarations for some exercises to have been much of a problem for myself or others. I may have had a very few students express they had difficulty, but I don't remember any specifically. When lacking forward declarations, my own approach has been to look at the tests to see what function signature(s) they expect and take it from there. It may be a bit challenging to have to do that, but it's also rewarding when, hey! I figured it out! 😄

joshgoebel:
> I can't say I recall the lack of forward declarations for some exercises to have been much of a problem for myself or others

I can only say I found it a problem - perhaps it was more the inconsistency though?  I was/am still a bit hazy on the difference between things like `*`, `**`, `[]`, `[][]` or what not... and just looking at the tests (which does work well in many other languages) I sometimes still found it challenging to parse those differences.  

Having some/most exercises I could jump right into and then only some where I hit the no headers wall - friction and frustration.  Perhaps if they all forced me to do the work it would have been better (in the long run), at least it wouldn't be inconsistent.

[...]

PR #974, opened 2024-04-14
Re-rank and re-order exercises

ryanplusplus:
This definitely isn't perfect, but I think it's a pretty significant improvement over our current state. What I did was:
- Re-assigned difficulties based on gut feel (from my own experience and by looking at our solutions), completion percentages (see https://exercism.org/tracks/c/build), and topics covered (for example, anything with memory management was bumped up to at least a 4 or 5... we had several at 1 and 2 :neutral_face:)
- Ordered all of the exercises by difficulty
- Within the lower difficulty levels, I ordered exercises such that not too much was introduced at a time.

As a next step, I think we can pick a difficulty cut-off point and start adding more complete skeletons to earlier exercises. This would address:
- https://github.com/exercism/c/issues/911
- https://github.com/exercism/c/issues/740

Fixes https://github.com/exercism/c/issues/971

### DIFF
--- a/config.json
+++ b/config.json
@@ -427,7 +427,6 @@
         "difficulty": 4,
         "topics": [
           "control_flow_if_statements",
-          "preprocessor_x_macros_in_test",
           "strings",
           "text_formatting"
         ]
@@ -491,7 +490,12 @@
         "uuid": "f1ad3d6e-7fc2-4461-8ea2-554c2fa9f1f0",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 4
+        "difficulty": 4,
+        "topics": [
+          "control_flow_if_else_statements",
+          "control_flow_loops",
+          "strings"
+        ]
       },
       {
         "slug": "beer-song",
--- a/config.json
+++ b/config.json
@@ -49,44 +49,27 @@
         ]
       },
       {
-        "slug": "armstrong-numbers",
-        "name": "Armstrong Numbers",
-        "uuid": "a4658ee9-b71c-4764-9652-7ed83d46516b",
+        "slug": "leap",
+        "name": "Leap",
+        "uuid": "9b047dda-b85e-4378-9487-9f9cce086e4f",
         "practices": [],
         "prerequisites": [],
         "difficulty": 1,
         "topics": [
           "control_flow_if_else_statements",
+          "logic"
+        ]
+      },
+      {
+        "slug": "difference-of-squares",
+        "name": "Difference of Squares",
+        "uuid": "9802cbc0-eb70-4840-b8ee-a228f1fc29ea",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1,
+        "topics": [
           "control_flow_loops",
-          "logic",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -103,220 +86,246 @@
...
+          "control_flow_if_else_statements",
+          "logic",
           "math"
         ]
       },
       {
-        "slug": "grade-school",
-        "name": "Grade School",
-        "uuid": "68b1de28-7c3d-4e8c-ab51-7abe6779e187",
+        "slug": "resistor-color-duo",
```
wolf99, line 141: I would have thought that this would come directlay after `resistor-color`, but it is a while since I've looked at the exercise.
Can you share why it isn't'?

config.json
```diff
@@ -335,401 +344,426 @@
...
+        "uuid": "b868eeff-400c-4e46-8675-10d85c14b503",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 1,
+        "difficulty": 4,
         "topics": [
-          "arrays",
-          "enums"
+          "control_flow_if_statements",
+          "preprocessor_x_macros_in_test",
```
wolf99, line 430: I think there are no X-macros here.
Perhaps there was at some point in the past.
Should this topic be removed? If so, will it affect tje difficulty rating or ordering?

config.json
```diff
@@ -335,401 +344,426 @@
...
+          "memory_management",
+          "strings",
+          "text_formatting"
         ]
       },
       {
-        "slug": "high-scores",
-        "name": "High Scores",
-        "uuid": "2c81f99a-991a-4914-aae7-d5bf72456119",
+        "slug": "protein-translation",
```
wolf99, line 489: Are there no topics for this exercise?
(Or perhaps I'm mis-reading the diff)

wolf99: Maybe fixing this is not in the scope of this PR?

wolf99 (commented):
Looks good!
I guess we try it and see if the stats improve for students' linear progression in the track
Some small question-ey comments only 🙂

config.json
ryanplusplus: No reason. I'll move it around.

config.json
ryanplusplus: Yep, I'll remove it. I don't think it makes much of a difference for difficulty. I rated it a 4 because it has string parsing and formatting, both of which are a bit of a pain for a C newbie.

config.json
ryanplusplus: Actually, because I rated them 1, 2, and 3 I can't get them to be right next to each other without having the difficulties out of order. I guess a benefit of them not being in a row is that people won't tired of resistors.

config.json
ryanplusplus: Good catch. While I'm here I'll add some in.

config.json
wolf99: Seems like a good enough reason, maybe leave them as they are then and we can see if we get any feedback.

wolf99 (approved):

PR #975, opened 2024-04-15
Add exercise rotational-cipher

ahans:


exercises/practice/rotational-cipher/test_rotational_cipher.c
```diff
@@ -0,0 +1,100 @@
...
+{
+}
+
+static void test_rotate(const char *expected, const char *text, int shift_key)
+{
+   char *rotated = rotate(text, shift_key);
+
+   TEST_ASSERT_EQUAL_STRING(expected, rotated);
+
+   free(rotated);
```
wolf99, line 21: Is it good to have this in the tests rather than the implementation?

It seems to imply a specific implementation or restrict the options for implementations.

wolf99 (changes requested):

exercises/practice/rotational-cipher/test_rotational_cipher.c
ahans: I'm not following. This exercise is about taking a string and returning a modified version of it. I think it's a standard pattern that the function allocates memory for the returned string and it's the caller's responsibility to free that memory. Similar exercises such as reverse-string, atbash-cipher, acronym, or pig-latin handle it in the exact same way.

config.json
```diff
@@ -1091,6 +1091,20 @@
...
           "structs"
         ]
+      },
+      {
+        "slug": "rotational-cipher",
+        "name": "Rotational Cipher",
+        "uuid": "3adbc85c-7e7e-4122-a37d-c843196bc8af",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2,
```
ryanplusplus, line 1101: How about a 4? I know this isn't actually difficult, but I'm redoing the track ordering and rated everything that uses *alloc as at least a 4 so that people don't run into memory management early on in the track.

```suggestion
        "difficulty": 4,
```

ryanplusplus: Never mind, I'll need to merge this into my branch anyway, I'll figure out a good place to stick it at that point.

exercises/practice/rotational-cipher/test_rotational_cipher.c
ryanplusplus: @ahans I think what @wolf99 means is that we should add a function in the implementation for freeing the result as we've done in other exercises (see `linked-list`, for example). I don't think that applies here since there's no freedom in the result type.

ryanplusplus (approved):

wolf99 (approved):

PR #980, opened 2024-04-17
Add exercise dnd-character

ahans:
Implements a missing exercise from #48in24 (#940).

### DIFF
--- a/exercises/practice/dnd-character/.meta/example.c
+++ b/exercises/practice/dnd-character/.meta/example.c
@@ -1,0 +1,43 @@
+#include "dnd_character.h"
+
+#include <stdlib.h>
+#include <math.h>
+
+static int dice_roll(void)
+{
+   // NB: This suffers from modulo bias, but probably good enough for the
+   // purpose of this exercise.
+   return 1 + rand() % 6;
+}
+
+int ability(void)
+{
+   const int rolls[4] = { dice_roll(), dice_roll(), dice_roll(), dice_roll() };
+   int min_index = 0;
+   for (int i = 1; i < 4; ++i) {
+      if (rolls[i] < rolls[min_index])
+         min_index = i;
+   }
+   int sum = 0;
+   for (int i = 0; i < 4; ++i) {
+      sum += rolls[i];
+   }
+   return sum - rolls[min_index];
+}
+
+int modifier(int score)
+{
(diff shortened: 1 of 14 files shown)


exercises/practice/dnd-character/.meta/tests.toml
```diff
@@ -0,0 +1,73 @@
...
+description = "random character is valid"
+
+[2ca77b9b-c099-46c3-a02c-0d0f68ffa0fe]
+description = "each ability is only calculated once"
+include = false
+
+[dca2b2ec-f729-4551-84b9-078876bb4808]
+description = "each ability is only calculated once"
+reimplements = "2ca77b9b-c099-46c3-a02c-0d0f68ffa0fe"
+include = false
```
wolf99, line 73: Are neither of these to be used?

exercises/practice/dnd-character/.meta/example.c
```diff
@@ -0,0 +1,43 @@
+#include "dnd_character.h"
+
+#include <stdlib.h>
+#include <math.h>
+
+static int dice_roll(void)
+{
+   // NB: This suffers from modulo bias, but probably good enough for the
+   // purpose if this exercise.
```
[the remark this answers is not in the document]
ahans: Already noticed myself and fixed it.

exercises/practice/dnd-character/.meta/tests.toml
ryanplusplus: I think that because the character is returned as a struct this test would be meaningless. There's no way for an ability to change after the fact like there would be if it were an opaque object with abilities accessible only via a getter.

ryanplusplus (approved):

exercises/practice/dnd-character/.meta/tests.toml
ahans: The second is a new version of the first, so we're not supposed to implement the first either way. But in general those tests don't make much sense in C. We return a struct, so there's no way things are computed multiple times. The C++ [implementation](https://github.com/exercism/cpp/tree/main/exercises/practice/dnd-character) uses plain attributes as well, although there we could have methods. In a language supporting actual properties like Python it makes more sense, and they do [implement](https://github.com/exercism/python/blob/main/exercises/practice/dnd-character/dnd_character_test.py#L76) those tests as well.

PR #984, opened 2024-04-25
Reorder new exercises

ryanplusplus:
Related to https://github.com/exercism/c/pull/974

These hit around the same time as the original re-ordering but were not ordered according to their score.

### DIFF
--- a/config.json
+++ b/config.json
@@ -237,6 +237,20 @@
         ]
       },
       {
+        "slug": "dnd-character",
+        "name": "D&D Character",
+        "uuid": "aafdaf60-1b68-41f7-92ae-f2d548738be9",
+        "practices": [],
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -431,6 +445,20 @@
           "text_formatting"
         ]
       },
+      {
+        "slug": "rotational-cipher",
+        "name": "Rotational Cipher",
+        "uuid": "3adbc85c-7e7e-4122-a37d-c843196bc8af",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4,
```
ryanplusplus, line 454: I bumped this difficulty up since it uses `malloc`. This keeps these exercises away from the beginning of the track.

wolf99 (approved):

PR #1001, opened 2024-07-04
Refactor binary to decimal conversion tests for clarity and organization

Imran-imtiaz48:


ryanplusplus:
Hey @Imran-imtiaz48 these tests are separate so that they track the canonical tests defined in the upstream problem-specifications repo. Additionally, having separate tests means that all failures are reported, not just the first failed check in a long super test.

Issue #1011, opened 2024-09-10
queen-attack test that prevents incorrectly accepted answers due to unsigned type.


edu-bm7:
Hello everyone, this is the first time I'm opening an Issue and trying to make a PR, if anything I'm doing is wrong, please let me know.

The issue:

most solutions on the queen-attack exercise uses something like `abs(queen_1.row - queen_2.row) == abs(queen_1.column - queen_2.column);` The problem is when we have a underflow in the first abs() and not on the second one. Leading the students forgetting to check for underflow since row and column are Unsigned Int.

There isn't a test to test this position(when both queens are on the opposite side of their initial position, but still on the same diagonal, which is a common chest position and the result should be `CAN_ATTACK`). I've made a simple test that checks this position and ensure the students don't commit this mistakes.

siebenschlaefer:
The tests for this exercise come from the language-agnostic [problem-specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/queen-attack/canonical-data.json).
Any changes to those tests should be discussed in the [forum](https://forum.exercism.org/c/exercism/bugs-and-features/126) first where maintainers of all tracks have a chance to participate in the discussion.

Also, I have to admit I understand neither the problem nor the proposed fix.

`abs(queen_1.row - queen_2.row) == abs(queen_1.column - queen_2.column)` looks fine to me, especially if you consider that the operands of the subtractions get promoted to `int` and therefore no integer overflow can happen.

And in your new test the queens on squares `{.column = 2, .row = 7 }` and `{.column = 6, .row = 1}` are not on the same diagonal but the tests expects `CAN_ATTACK`.
```text
         0   1   2   3   4   5   6   7
       ┌───┬───┬───┬───┬───┬───┬───┬───┐
      0│   │   │   │   │   │   │   │   │0
       ├───┼───┼───┼───┼───┼───┼───┼───┤
      1│   │   │   │   │   │   │ B │   │1
       ├───┼───┼───┼───┼───┼───┼───┼───┤
      2│   │   │   │   │   │   │   │   │2
       ├───┼───┼───┼───┼───┼───┼───┼───┤
      3│   │   │   │   │   │   │   │   │3
       ├───┼───┼───┼───┼───┼───┼───┼───┤
      4│   │   │   │   │   │   │   │   │4
       ├───┼───┼───┼───┼───┼───┼───┼───┤
      5│   │   │   │   │   │   │   │   │5
[...]

PR #1012, opened 2024-09-10
 queen-attack test that prevents incorrectly accepted answers due to unsigned type.

edu-bm7:
[Issue: 1011](https://github.com/exercism/c/issues/1011#issue-2515284606)

A lot of user solutions that I've seen use some type of  `abs(queen_1.row - queen_2.row) == abs(queen_1.column - queen_2.column);`. This test make sure that students are taking into account underflow that happens because of the Unsigned int type.

Tests if Queens can attack each other even when they crossed sides of the board, which is a common position on chest. Also it ensure that the students start to take into account overflows/underflows that happens often in C/C++


edu-bm7:
> or, for example, adds an edge case that is not worth rerunning all tests for

I do believe that it is a case where re-testing should be consider since the test enforces good practice when coding in C/C++ also makes them search why their solution is wrong.

PR #1022, opened 2024-12-01
minesweeper: add rows argument to free function

rivnakm:
At the moment there isn't a way to get this exercise to pass `make memcheck` since there isn't enough information for the `annotation_free()` function to free memory. Other similar exercises can get by since the matrix is square, but there isn't the same constraint for the minesweeper exercise. 

Changing the function signature to specify the number of rows explicitly.

### DIFF
--- a/exercises/practice/minesweeper/minesweeper.h
+++ b/exercises/practice/minesweeper/minesweeper.h
@@ -3,6 +3,6 @@
 #include <stddef.h>
 
 char **annotate(const char **minefield, const size_t rows);
-void free_annotation(char **annotation);
+void free_annotation(char **annotation, const size_t rows);
 
 #endif
--- a/exercises/practice/minesweeper/test_minesweeper.c
+++ b/exercises/practice/minesweeper/test_minesweeper.c
@@ -28,7 +28,7 @@ static void test_annotate_no_columns(void)
    const size_t rows = ARRAY_SIZE(expected);
    char **actual = annotate(minefield, rows);
    TEST_ASSERT_EQUAL_STRING_ARRAY(expected, actual, rows);
-   free_annotation(actual);
+   free_annotation(actual, rows);
 }
 
 static void test_annotate_no_mines(void)
@@ -51,7 +51,7 @@ static void test_annotate_no_mines(void)
    const size_t rows = ARRAY_SIZE(expected);
    char **actual = annotate(minefield, rows);
    TEST_ASSERT_EQUAL_STRING_ARRAY(expected, actual, rows);
-   free_annotation(actual);
+   free_annotation(actual, rows);
 }
 
 static void test_annotate_minefield_with_only_mines(void)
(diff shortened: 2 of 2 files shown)


ryanplusplus:
There are at least two ways to pass `make memcheck`:
- Our [example](https://github.com/exercism/c/blob/284fa41c8804af490244b772ccb21c6cf8adb519/exercises/practice/minesweeper/.meta/example.c) makes one large allocation to hold all of the rows and the array of rows makes several pointers into the allocation of all rows.
- My solution (see below) allocates a container that stores the matrix and the number of rows. When it comes time to free, it's possible to backtrack from the matrix to this container storing the matrix to determine the number of rows.

```c
#include <stdlib.h>
#include <string.h>
#include <stdint.h>
#include <stdbool.h>
#include "minesweeper.h"

#define container_of(container_type, member_name, member_pointer) \
  (container_type*)(((char*)member_pointer) - offsetof(container_type, member_name))

typedef struct {
  size_t count;
  char* rows[];
} annotation_t;

static annotation_t* create_annotation(const char** minefield, const size_t rows)
{
  annotation_t* annotation = malloc(sizeof(annotation_t) + sizeof(char*) * rows);
  annotation->count = rows;

  for(size_t i = 0; i < rows; i++) {
    annotation->rows[i] = malloc(strlen(minefield[i]) + sizeof('\0'));
    strcpy(annotation->rows[i], minefield[i]);
  }

  return annotation;
}

char** annotate(const char** m, const size_t rows)
{
  if(!rows) return NULL;

  annotation_t* annotation = create_annotation(m, rows);

  const size_t columns = strlen(m[0]);

  for(size_t r = 0; r < rows; r++) {
    for(size_t c = 0; c < columns; c++) {
      if(m[r][c] == '*') continue;

      uint8_t mines = 0;

      const bool r0 = true;
      const bool c0 = true;
      const bool rp = r + 1 < rows;
      const bool cp = c + 1 < columns;
      const bool rn = !!r;
      const bool cn = !!c;

      if(r0 && cn && m[r + 0][c - 1] == '*') mines += 1;
      if(rn && cn && m[r - 1][c - 1] == '*') mines += 1;
      if(rn && c0 && m[r - 1][c + 0] == '*') mines += 1;
      if(rn && cp && m[r - 1][c + 1] == '*') mines += 1;
      if(r0 && cp && m[r + 0][c + 1] == '*') mines += 1;
      if(rp && cp && m[r + 1][c + 1] == '*') mines += 1;
      if(rp && c0 && m[r + 1][c + 0] == '*') mines += 1;
      if(rp && cn && m[r + 1][c - 1] == '*') mines += 1;

      if(mines) annotation->rows[r][c] = mines + '0';
    }
  }

  return annotation->rows;
}

void free_annotation(char** _annotation)
{
  annotation_t* annotation = container_of(annotation_t, rows, _annotation);

  for(size_t i = 0; i < annotation->count; i++) {
    free(annotation->rows[i]);
  }

  free(annotation);
}
```

That said, neither of these is particularly obvious. I think it's fair to say that we should consider adding an append to the instructions so that we can expose at least one of these solutions to students.

What do you think about adding an append to the instructions?

rivnakm:
As someone relatively new to C these didn't occur to me. I think the example solution is perfectly reasonable, so I think just updating the instructions to add a hint to try to use a single allocation would be helpful.

rivnakm:
Although reading through community submitted solutions should be sufficient to figure this out so I think this is fine as-is.

PR #1026, opened 2024-12-11
Update instructions.md

mattshu:
original instructions made it seem extra step was needed (signal an error rather than return empty string). also updated expected output


wolf99:
Hi @mattshu 
Thank you for making a commit to improve exercism!

However, these instructions are synced for all languages tracks from a central source. 
If you would like to change these it is best to make the change there: https://github.com/exercism/problem-specifications

PR #1032, opened 2025-01-09
Updating tests for phone-number

jagdish-15:
### Pull Request: Syncing Tests for Phone-Number Exercise

I've synced the `toml` file with the problem specification repository. Currently, the tests are functioning as expected, with the program returning a "0000...0" phone number for invalid inputs. However, according to the `canonical-data.json` file in the `problem-specification` repository, the suggested behavior is for the program to throw an error when an invalid phone number is passed as an argument.

To align with the problem specification, we would need to:

1. Add the error handling where required in the test file.
2. Update the example answers accordingly.

Since the program is currently working with the existing approach (returning "0000...0"), I'm asking whether you'd prefer to proceed with the new specification of throwing an error for invalid phone numbers. Let me know how you'd like to proceed.

### DIFF
--- a/exercises/practice/phone-number/.meta/config.json
+++ b/exercises/practice/phone-number/.meta/config.json
@@ -7,6 +7,7 @@
     "Gamecock",
     "gea-migration",
     "h-3-0",
+    "jagdish-15",
     "mikewalker",
     "patricksjackson",
     "QLaille",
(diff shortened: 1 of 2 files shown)


wolf99:
Tha k you for all the updates @jagdish-15 !

Without thinking this through very deeply, I feel like changing this to add more complex error handling would result in changing the FUT interface such that the exercise becomes more about string memory management rather than string interpretation and mutation.

For that reason I would err on the side of keeping the existing behavior for errors.

I'm open to correction though!

jagdish-15:
You're welcome, @wolf99! I completely agree with your observation. Keeping the approach consistent with exercises like `protein-translation` is a great choice, as introducing complex errors could unnecessarily complicate things without adding much value.

I've updated all the necessary files accordingly, and since no changes are needed to the test file, everything should be good to go. Let me know if there's anything else you'd like me to address! 😊

wolf99 (approved):

PR #1033, opened 2025-01-09
Updating tests for Pig-Latin

jagdish-15:
### Pull Request: Syncing Tests for Pig-Latin Exercise

This pull request updates the test suite for the **Pig-Latin** exercise to sync it with the problem specification repository.

### DIFF
--- a/exercises/practice/pig-latin/.meta/tests.toml
+++ b/exercises/practice/pig-latin/.meta/tests.toml
@@ -1,69 +1,79 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [11567f84-e8c6-4918-aedb-435f0b73db57]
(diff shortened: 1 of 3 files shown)


exercises/practice/pig-latin/.meta/tests.toml
```diff
@@ -1,69 +1,79 @@
...
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [11567f84-e8c6-4918-aedb-435f0b73db57]
-description = "word beginning with a"
+description = "ay is added to words that start with vowels -> word beginning with a"
```
wolf99, line 13: These descriptions are updated (correctly). In this track the test function names usually match the descriptions, but the test functions names are not updated.
Should they be?

I understand that the descriptions are changed by additional only. And I'm not sure if our formatter handles function names this long very well or not.
Thoughts @ryanplusplus ?

ryanplusplus: The formatter doesn't handle very long function names gracefully. I've seen this in at least one test suite.

I think that if our test suite supported it, the new additions to the test names would be part of test sub-groups (ie: a nested `describe` in an RSpec-like framework). I'm okay with leaving the "short" names since the alternative is going to get pretty messy.

ryanplusplus (approved):

exercises/practice/pig-latin/test_pig_latin.c
```diff
@@ -14,142 +14,148 @@ void tearDown(void)
    res = NULL;
 }
 
-static void check_transate(const char *phrase, const char *expected)
+static void check_translate(const char *phrase, const char *expected)
```
ryanplusplus, line 17: I noticed this misspelling when reviewing the new test

PR #1036, opened 2025-01-09
Updating tests for Anagram

jagdish-15:
### Pull Request: Syncing Tests for Anagram Exercise

This pull request updates the test suite for the **Anagram** exercise to sync it with the problem specification repository.

### DIFF
--- a/exercises/practice/anagram/.meta/config.json
+++ b/exercises/practice/anagram/.meta/config.json
@@ -9,6 +9,7 @@
     "h-3-0",
     "hintjens",
     "JacobMikkelsen",
+    "jagdish-15",
     "kytrinyx",
     "lpil",
     "patricksjackson",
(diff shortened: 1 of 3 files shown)


ryanplusplus (changes requested):
I think that we should omit these new tests for the reasons that @wolf99 gave in https://github.com/exercism/c/pull/1037.

jagdish-15:
Hi @ryanplusplus,

I've implemented the same changes as those in #1037. I’d appreciate it if you could review the PR. However, it seems there's an issue with the check step after incorporating the suggestions from @wolf99. Could you please take a look and advise?

wolf99 (approved):

ryanplusplus (approved):

ryanplusplus:
Thanks!

PR #1042, opened 2025-03-02
Added line to include test requirement

mattshu:
I'm still learning C (and PRs) so my skills might not be there yet — but I stumbled hard on the very first test (`test_reading_empty_buffer_fails()`) because I didn't realize I had to set `errno` to a special value.

I thought this change could serve as a small hint for newbies like me to consider how errors are raised in C. I might be overthinking it, though!

### DIFF
--- a/exercises/practice/circular-buffer/.docs/instructions.md
+++ b/exercises/practice/circular-buffer/.docs/instructions.md
@@ -34,7 +34,9 @@ If the buffer has 7 elements then it is completely full:
 [5][6][7][8][9][3][4]
 ```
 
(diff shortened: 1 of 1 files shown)


ryanplusplus:
Hey @mattshu, these docs are generated from the upstream/common problem specifications. To add track specific data we need to either add an append. Here's an example of an append on this track:
https://github.com/exercism/c/blob/c8bad6385d229d33696a39f45d185513a59bc34c/exercises/practice/gigasecond/.docs/instructions.append.md#L4

mattshu:
Thank you for the info @ryanplusplus!

PR #1043, opened 2025-03-03
Create instructions.append.md

mattshu:
Third time is the charm--let's see if I actually read the manual this time :+1: Added hint regarding setting `errno`

### DIFF
--- a/exercises/practice/circular-buffer/.docs/instructions.append.md
+++ b/exercises/practice/circular-buffer/.docs/instructions.append.md
@@ -1,6 +1,6 @@
 # Set `errno` to a special value
 If the buffer is empty during a read attempt, an error is raised to alert the client.
 
-These error values can be found in the `errno.h` header file.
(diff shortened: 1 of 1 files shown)


exercises/practice/circular-buffer/.docs/instructions.append.md
```diff
@@ -0,0 +1,6 @@
+# Set `errno` to a special value
+If the buffer is empty during a read attempt, an error is raised to alert the client.
+
+These error values can be found in the `errno.h` header file.
+
+[Wikipedia](https://en.wikipedia.org/wiki/Errno.h)
```
ryanplusplus, line 6:
```suggestion
These error values can be found in the [`errno.h`][errno.h] header file.

[errno.h]: https://en.wikipedia.org/wiki/Errno.h
```

This switches the link to be inline. This is hopefully a bit more usable and is consistent with how we've handled similar links elsewhere in the track.

ryanplusplus (approved):
Thanks!

PR #1044, opened 2025-03-08
Corrects the instructions so that the binary representation is accurate

armahillo:
In the existing instructions, it indicates that the word 0000 1000 in binary would be 16 in decimal. This is incorrect. The correct value for that word is 8.

Since the binary representation is also in an SVG, it is easier to just change the decimal value. The solution is unchanged with this value, but understanding the binary representation is key to solving the exercise.

### DIFF
--- a/exercises/practice/eliuds-eggs/.docs/introduction.md
+++ b/exercises/practice/eliuds-eggs/.docs/introduction.md
@@ -58,7 +58,7 @@ The position information encoding is calculated as follows:
 
 ### Decimal number on the display
 
-16
+8
 
 ### Actual eggs in the coop
 


ryanplusplus:
Hey @armahillo, these instructions are generated from the upstream problem specifications. Specifically, they come from this file:
https://github.com/exercism/problem-specifications/blob/main/exercises/eliuds-eggs/introduction.md

I agree that this is wrong. Can you file a PR against the problem specifications repo so that this gets fixed for all tracks?

BNAndras:
This was just updated in the problem-specifications repo so you’ll just need to do a sync at this point.

ryanplusplus:
Synced in https://github.com/exercism/crystal/pull/711

PR #1050, opened 2025-04-24
Correct typo in introduction.md

dpflug:
Straightforward change


dpflug:
I should also mention that my linter complains about `abs()` and `bool` being used, but I don't know which C standard you're targeting.

ryanplusplus (approved):

wolf99:
> I should also mention that my linter complains about `abs()` and `bool` being used, but I don't know which C standard you're targeting. 

IIRC we target C99 as the lowest common denominator. Interesting that you see issues with abs() and bool. Can you share what tooling your using and which versions that tooling is targeting?

PR #1051, opened 2025-05-04
Disable lychee-action

ryanplusplus:
The lychee checks have been failing often for gnu.org and now they seem to fail 100% for GitHub links. I tried updating the lychee action with the same result and found [an issue](https://github.com/lycheeverse/lychee/issues/733) that seems to describe a similar issue that is blamed on a dependency of lychee that can't be easily fixed.

It seems like this check has gotten to be more trouble than it's worth. What does everyone think about disabling it? @wolf99 @siebenschlaefer

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -36,14 +36,3 @@ jobs:
 
       - name: Check include guards
         run: ./bin/check-include-guards ./exercises/concept ./exercises/practice
-
-      - name: Check links
-        uses: lycheeverse/lychee-action@1d97d84f0bc547f7b25f4c2170d87d810dc2fb2c
-        id: lychee
-        with:
-          args: --user-agent "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.75 Safari/537.36" --no-progress **/*.md **/*.html
-        env:
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-
-      - name: Fail if there were link errors
-        run: exit ${{ steps.lychee.outputs.exit_code }}


wolf99:
The errors for GitHub links look like rate limiting. We could add `--cache` for that.

ryanplusplus:
> The errors for GitHub links look like rate limiting. We could add `--cache` for that.

I have been seeing issues on the first run (ie: runs that aren't quickly following another) in addition to subsequent runs. I assume that it's because of requests from other workflows on other repos. I did a quick check to see whether `--cache` would help us and got rate limited before the first run completed (which makes sense because I think it only helps on subsequent runs):
https://github.com/exercism/c/actions/runs/14841633794/job/41665758426?pr=1052

ahans:
We could make it a check that does not block merging or allow one to skip it with a comment or tag in the title. Then it would still do its thing and when it flags something it shouldn't, one can still manually check and override when a link is in fact valid.

ahans:
> > The errors for GitHub links look like rate limiting. We could add `--cache` for that.
> 
> I have been seeing issues on the first run (ie: runs that aren't quickly following another) in addition to subsequent runs. I assume that it's because of requests from other workflows on other repos. I did a quick check to see whether `--cache` would help us and got rate limited before the first run completed (which makes sense because I think it only helps on subsequent runs): https://github.com/exercism/c/actions/runs/14841633794/job/41665758426?pr=1052

Did you also include a cache action? AFAIU, GH runners are ephemeral, so whatever a tool caches will not be available to another run, unless extra measures are taken (such as using the cache action).

ryanplusplus:
> Did you also include a cache action? AFAIU, GH runners are ephemeral, so whatever a tool caches will not be available to another run, unless extra measures are taken (such as using the cache action).

I did not, good call. The issue is that these checks are, at this point, never passing so there wouldn't be a successful check to cache.

> We could make it a check that does not block merging or allow one to skip it with a comment or tag in the title. Then it would still do its thing and when it flags something it shouldn't, one can still manually check and override when a link is in fact valid.

We could do this, but how much value is there to having a check that always fails with many false positives? I'm skeptical that anyone is going to look through the results to see if any of the 50+ failures are real.

ahans (approved):
I will try to look into it if I don't forget. Let's disable for now.

PR #1053, opened 2025-05-24
[Exercise] implement variable-length-quantity exercise

Wiguwbe:
Hello,

A few days ago I was trying to create a encoder/decoder for SQLite's `Varint`s. Instead of implementing the tests by hand, I searched for already implemented tests.

I stumbled upon _Exercism_ and the VLQ exercise in Python. As it was not implemented in C, I decided to port it. Hopefully someone will find it fun/useful as well.

Note that this is my first time contributing to (or even using) Exercism.

The test generation was automated using the `canonical-data.json` as input. It's a relatively specific script (and implemented in Perl), but I'm attaching here still:
[gen-tests.pl.gz](https://github.com/user-attachments/files/20422427/gen-tests.pl.gz) (gzip-ed so that github accepts it)

I have a couple questions for the reviewers:

- The `example.c` may no be very reader-friendly. Is it expected to be read by someone or is it just to test the tests?
- I'm using `uint64_t` for the decoded integers, but it's not expected to have integers bigger than 32-bits. Is it preferable to use 32 bit integers?

### DIFF
--- a/exercises/practice/variable-length-quantity/.meta/example.h
+++ b/exercises/practice/variable-length-quantity/.meta/example.h
@@ -1,12 +1,18 @@
-#ifndef VLQ_H
-#define VLQ_H
+#ifndef VARIABLE_LENGTH_QUANTITY_H
+#define VARIABLE_LENGTH_QUANTITY_H
 
 #include <stdint.h>
+#include <stddef.h>
 
-// return output_len
-int encode(uint64_t *integers, int integers_len, uint8_t *output);
+/// Encode an array of integers into Variable Length Quantity.
+///
+/// The function returns the size of the output buffer used.
+int encode(const uint32_t *integers, size_t integers_len, uint8_t *output);
 
-// return output_len, -1 on error
-int decode(uint8_t *buffer, int buffer_len, uint64_t *output);
+/// Decode a Variable Length Quantity byte array into an array of integers.
+///
+/// The function returns the size of the output buffer used, or -1 in an error
+/// occurred.
+int decode(const uint8_t *bytes, size_t buffer_len, uint32_t *output);
 
 #endif
--- a/exercises/practice/variable-length-quantity/.meta/example.h
+++ b/exercises/practice/variable-length-quantity/.meta/example.h
@@ -11,7 +11,7 @@ int encode(const uint32_t *integers, size_t integers_len, uint8_t *output);
 
 /// Decode a Variable Length Quantity byte array into an array of integers.
 ///
-/// The function returns the size of the output buffer used, or -1 in an error
+/// The function returns the size of the output buffer used or -1 if an error
 /// occurred.
 int decode(const uint8_t *bytes, size_t buffer_len, uint32_t *output);
 
--- a/exercises/practice/variable-length-quantity/.meta/example.h
+++ b/exercises/practice/variable-length-quantity/.meta/example.h
@@ -1,0 +1,12 @@
+#ifndef VLQ_H
+#define VLQ_H
+
+#include <stdint.h>
+
+// return output_len
+int encode(uint64_t *integers, int integers_len, uint8_t *output);
+
+// return output_len, -1 on error
+int decode(uint8_t *buffer, int buffer_len, uint64_t *output);
+
+#endif
--- a/exercises/practice/variable-length-quantity/variable_length_quantity.h
+++ b/exercises/practice/variable-length-quantity/variable_length_quantity.h
@@ -1,12 +1,18 @@
-#ifndef VLQ_H
-#define VLQ_H
+#ifndef VARIABLE_LENGTH_QUANTITY_H
+#define VARIABLE_LENGTH_QUANTITY_H
 
 #include <stdint.h>
+#include <stddef.h>
 
-// return output_len
-int encode(uint64_t *integers, int integers_len, uint8_t *output);
+/// Encode an array of integers into Variable Length Quantity.
+///
+/// The function returns the size of the output buffer used.
+int encode(const uint32_t *integers, size_t integers_len, uint8_t *output);
 
-// return output_len, -1 on error
-int decode(uint8_t *buffer, int buffer_len, uint64_t *output);
+/// Decode a Variable Length Quantity byte array into an array of integers.
+///
+/// The function returns the size of the output buffer used, or -1 in an error
+/// occurred.
+int decode(const uint8_t *bytes, size_t buffer_len, uint32_t *output);
 
 #endif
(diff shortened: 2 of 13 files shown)


siebenschlaefer:
Hi, 

thank you for translating this exercise to C! I just noticed a few things:

- The include guard is `VLQ_H`.  
    In all other exercises it's the name of the directory, in uppercase, hyphens replaced by underscores, with `_H` at the end (e.g. `RUN_LENGTH_ENCODING_H` in the [run-length-encoding](https://github.com/exercism/c/blob/main/exercises/practice/run-length-encoding/run_length_encoding.h#L1-L2) exercise).
    I'd suggest using `VARIABLE_LENGTH_QUANTITY_H` instead of `VLQ_H`.
- The comments in the `.h` file explain the return value of the two functions.  
    Usually when a `.h` file contains declarations of the functions with comments, they are similar to "docstrings" (e.g. in the exercises [anagram](https://github.com/exercism/c/blob/main/exercises/practice/anagram/anagram.h#L20-L23), [high-scores](https://github.com/exercism/c/blob/main/exercises/practice/high-scores/high_scores.h#L13-L14), [linked-list](https://github.com/exercism/c/blob/main/exercises/practice/linked-list/linked_list.h#L9), [series](https://github.com/exercism/c/blob/main/exercises/practice/series/series.h#L14-L28), [sieve](https://github.com/exercism/c/blob/main/exercises/practice/sieve/sieve.h#L7-L10)). They explain what the function does, not only one aspect of it. The only exception so far with comments similar to this one is in the [react](https://github.com/exercism/c/blob/main/exercises/practice/react/react.h#L11) exercise.
    Instead of just explaining the result I'd prefer a more complete comment that reads as if it were directed to the users of the function. Or, if this comment is meant for the students, move it to the `.c` file.
- The first parameter of both functions in the stubs is a `uint64_t *`.  
    If we want to provide declarations of the two functions in the `.h` file I'd prefer the first parameter being a pointer to `const`. I'd argue that this is best practice for pointers to read-only arguments.
- The second parameter of both functions in the stubs is an `int`.  
    In most other exercises we use `size_t` for parameters that represent a length.  
    IMHO a signed type is fine, I just wanted to point that out.
- The stub of `decode()` uses the term "buffer" for the input.  
    *Personally* I'd prefer something like `bytes` because in my experience "buffer" is most commonly used for output.
- All tests are enabled, right from the start.  
    Most exercises have only the first tests enabled, the other tests are disabled with the ` TEST_IGNORE();` macro, and the first of those lines is followed by the comment "// delete this line to run test" (e.g. in the [knapsack](https://github.com/exercism/c/blob/main/exercises/practice/knapsack/test_knapsack.c#L24) exercise.)  
    I'd suggest doing that in this exercise, too.
- The tests use decimal integer literals and call `TEST_ASSERT_EQUAL_UINT8_ARRAY` and `TEST_ASSERT_EQUAL_UINT64_ARRAY` to compare the actual and expected result.  
    A [comment](https://github.com/exercism/problem-specifications/blob/main/exercises/variable-length-quantity/canonical-data.json#L7) in the tests file recommends displaying numbers as hex literals. IMHO that applies to the literals in the test file and the output of the tests.    
    I'd suggest using hexadecimal integer literals in the tests and calling `TEST_ASSERT_EQUAL_HEX8_ARRAY` and `TEST_ASSERT_EQUAL_HEX64_ARRAY` (see the [documentation](https://github.com/ThrowTheSwitch/Unity/blob/master/docs/UnityAssertionsReference.md#naming-and-parameter-conventions)).
- In the example solution there's a section for decoding `SQLite_VARINT` with a maximal length of 9 bytes and a special treatment of the 9th byte.  
    I would really prefer not to include that. It's not relevant for this exercise, it makes the code longer, it increases the amount of code that has to be maintained, and we would need additional tests just for that section to ensure the code is correct.

Otherwise this looks good!

---

> The `example.c` may no be very reader-friendly. Is it expected to be read by someone or is it just to test the tests?

AFAIK we haven't discussed that yet for the C track. Its main purpose is to prove that the exercise can be solved. But in my opinion it should be a "good" solution, that's correct, somewhat easy to read, idiomatic, and reasonably efficient.

> I'm using `uint64_t` for the decoded integers, but it's not expected to have integers bigger than 32-bits. Is it preferable to use 32 bit integers?

*Personally* I would prefer 32-bit integers, I think that would match the instructions a little bit better ("* for this exercise we will restrict ourselves to only numbers that fit in a 32-bit unsigned integer*"). But I can live with 64-bit integers.

But (disclaimer) I'm not a maintainer of the C track, I have no vote here.

---

Just to spare the maintainers some time: The makefile and test framework are identical to the other exercises. The `.docs` and tests are complete. I was able to solve the exercise using the provided files. I have not found any issues or problems other than what I wrote above.

wolf99:
Good review @siebenschlaefer !

Wiguwbe:
Many thanks @siebenschlaefer for the very good review indeed.

I believe I have addressed all of the items. I kept it on a separate commit for ease of reading, I can squash it into the previous one if it preferable later. Let me know if the changes satisfy.

> (...) it should be a "good" solution, that's correct, somewhat easy to read, idiomatic, and reasonably efficient.

I never really did any C code _professionally_, or have it reviewed, so I'm not sure what "good"/idiomatic/easy to read code looks like :smile: . I'll see about improving tomorrow, hopefully I won't make it worse, but if it's not a blocker I'd say to not wait for me.

exercises/practice/variable-length-quantity/variable_length_quantity.h
```diff
@@ -0,0 +1,18 @@
...
+#include <stddef.h>
+
+/// Encode an array of integers into Variable Length Quantity.
+///
+/// The function returns the size of the output buffer used.
+int encode(const uint32_t *integers, size_t integers_len, uint8_t *output);
+
+/// Decode a Variable Length Quantity byte array into an array of integers.
+///
+/// The function returns the size of the output buffer used, or -1 in an error
```
ryanplusplus, line 14:
```suggestion
/// The function returns the size of the output buffer used or -1 if an error
```

exercises/practice/variable-length-quantity/.meta/example.h
```diff
@@ -0,0 +1,18 @@
...
+#include <stddef.h>
+
+/// Encode an array of integers into Variable Length Quantity.
+///
+/// The function returns the size of the output buffer used.
+int encode(const uint32_t *integers, size_t integers_len, uint8_t *output);
+
+/// Decode a Variable Length Quantity byte array into an array of integers.
+///
+/// The function returns the size of the output buffer used, or -1 in an error
```
ryanplusplus, line 14:
```suggestion
/// The function returns the size of the output buffer used or -1 if an error
```

ryanplusplus (approved):

PR #1062, opened 2025-06-14
Add intergalactic-transmission

ahans:
Reviewed this over at [arm64-assembly](https://github.com/exercism/arm64-assembly/pull/161) and think it's a nice exercise, so I figured I'd add it here!

The test file is taken from the arm64 track with minor modifications. Over there we have generators for all test files. What would other reviewers think of introducing something like that here as well?

### DIFF
--- a/exercises/practice/intergalactic-transmission/.meta/example.c
+++ b/exercises/practice/intergalactic-transmission/.meta/example.c
@@ -1,0 +1,73 @@
+#include "intergalactic_transmission.h"
+
+int transmit_sequence(uint8_t *buffer, const uint8_t *message,
+                      int message_length)
+{
+   uint8_t cur = 0;
+   int parity = 0;
+   int byte_count = 0;
+   int bit_count = 0;
+   // go over each input byte
+   for (int i = 0; i < message_length; ++i) {
+      // read bits of input byte, starting from most significant bit
+      for (int j = 0; j < 8; ++j) {
+         uint8_t const bit = (message[i] >> (7 - j)) & 1;
+         // keep track of number of set bits
+         parity += bit;
+         // set bit in output byte and shift to make room for next bit
+         cur = (cur | bit) << 1;
+         // if 7 bits have been written, output byte is ready
+         if (++bit_count == 7) {
+            // set parity bit
+            cur |= parity & 1;
+            // and write to output buffer
+            buffer[byte_count++] = cur;
+            // reset state for next output byte
+            parity = 0;
+            bit_count = 0;
+            cur = 0;
+         }
+      }
+   }
+   // make sure to write remaining state
+   if (bit_count > 0) {
+      // shift to the left to fill up with zero bits
+      cur <<= 7 - bit_count;
+      // set parity bit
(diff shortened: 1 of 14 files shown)


exercises/practice/intergalactic-transmission/intergalactic_transmission.h
```diff
@@ -0,0 +1,13 @@
+#ifndef INTERGALACTIC_TRANSMISSION_H
+#define INTERGALACTIC_TRANSMISSION_H
+
+#include <stdint.h>
+
+#define WRONG_PARITY -1
+
+int transmit_sequence(uint8_t *buffer, const uint8_t *message,
```
ahans, line 8: The focus of this exercise is on the encoding/decoding logic and not so much on figuring out function signatures. As a student I always find it annoying having to figure those out myself when i actually want to focus on the problem at hand, so I kept them here. Let me know if you think they should be removed.

ryanplusplus:
> Over there we have generators for all test files. What would other reviewers think of introducing something like that here as well?

As long as we can keep complexity in check, I'm in favor.

exercises/practice/intergalactic-transmission/test_intergalactic_transmission.c
```diff
@@ -0,0 +1,347 @@
+#include "intergalactic_transmission.h"
+
+#include "test-framework/unity.h"
+
+#include <stdint.h>
+
+#define BUFFER_SIZE 100
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
+#define WRONG_PARITY -1
```
ryanplusplus, line 9: I think this already comes from `intergalactic_transmission.h`

exercises/practice/intergalactic-transmission/.meta/example.c
```diff
@@ -0,0 +1,73 @@
...
+{
+   uint8_t cur_dec = 0;
+   int num_bits_left = 8;
+   int num_decoded = 0;
+   // iterate over input message: each byte has 7 bits payload and parity bit
+   for (int i = 0; i < message_length; ++i) {
+      uint8_t const payload = message[i] >> 1;
+      uint8_t const parity = message[i] & 1;
+      // check parity bit and return early if mismatch found
+      if ((__builtin_popcount(payload) & 1) != parity)
```
ryanplusplus, line 54: :+1:

ryanplusplus (changes requested):

exercises/practice/intergalactic-transmission/test_intergalactic_transmission.c
ahans: Good catch! Removed.

ahans:
> > Over there we have generators for all test files. What would other reviewers think of introducing something like that here as well?
> 
> As long as we can keep complexity in check, I'm in favor.

Alright, I will experiment a bit and create an issue so we can discuss (or maybe it'll be a PR already).

ryanplusplus (approved):

ryanplusplus:
Thanks!

PR #1070, opened 2025-07-10
Secret Handshake: Revised the tests and .meta/example.c.

borderite:
Per our discussion in the [forum](http://forum.exercism.org/t/the-secret-handshaking-exercise-in-c-does-not-check-the-number-of-actions/18173/5), I have made the following changes to the Secret Handshake exercise.

- The previous version of tests let solutions with extra actions pass.

- The tests now require that each solution returns an array of size 4, and the unused slots of the array be filled with NULL.

- .meta/example.c has been revised suitably so that it passes all revised tests.

- Given the nature of the revision, I don't think that the existing solutions need to be re-checked.

[no important files changed]

### DIFF
--- a/exercises/practice/secret-handshake/.meta/example.c
+++ b/exercises/practice/secret-handshake/.meta/example.c
@@ -2,19 +2,20 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-const char *actions[] = {"wink", "double blink", "close your eyes", "jump"};
+const char *actions[] = { "wink", "double blink", "close your eyes", "jump" };
 
-const char **commands(const size_t number) {
-  const char **action_sequence = calloc(4, sizeof(char *));
-  size_t k = 0;
-  if (number & 16) {
-    for (int i = 3; i >= 0; i--)
-      if (number & 1 << i)
-        action_sequence[k++] = actions[i];
-  } else {
-    for (int i = 0; i <= 3; i++)
-      if (number & 1 << i)
-        action_sequence[k++] = actions[i];
-  }
-  return action_sequence;
+const char **commands(const size_t number)
+{
+   const char **action_sequence = calloc(4, sizeof(char *));
+   size_t k = 0;
+   if (number & 16) {
+      for (int i = 3; i >= 0; i--)
+         if (number & 1 << i)
+            action_sequence[k++] = actions[i];
+   } else {
+      for (int i = 0; i <= 3; i++)
+         if (number & 1 << i)
+            action_sequence[k++] = actions[i];
+   }
+   return action_sequence;
 }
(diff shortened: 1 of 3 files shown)


exercises/practice/secret-handshake/.meta/example.c
```diff
@@ -1,60 +1,20 @@
...
-      size_t k = reverse ? (CMD_COUNT - 1 - i) : i;
-      if ((number & cmds[k].num) == cmds[k].num)
-         cmd_array[pos_cmd_array++] = cmds[k].cmd;
-   }
-
-   return cmd_array;
+const char* actions[] = {"wink", "double blink", "close your eyes", "jump"};
+
+const char** commands(const size_t number) {
+  const char** action_sequence = (const char**)calloc(4, sizeof(char*));
```
ryanplusplus, line 8:
```suggestion
  const char** action_sequence = calloc(4, sizeof(char*));
```

https://stackoverflow.com/questions/605845/should-i-cast-the-result-of-malloc-in-c

ryanplusplus:
Can you run the `format.sh` script in the root of this repo? It will fix the formatting in your code so that CI passes.

exercises/practice/secret-handshake/.meta/example.c
borderite: @ryanplusplus Your suggestion has been incorporated.

borderite:
I have run format.sh. Besides the codes I modified, the script modified two other files:

- exercises/practice/grade-school/test_grade_school.c
- exercises/practice/pascals-triangle/test_pascals_triangle.c

But I did not include these changes in my commit, because they are outside the scope of the current PR.

exercises/practice/secret-handshake/test_secret_handshake.c
```diff
@@ -1,8 +1,12 @@
-#include "test-framework/unity.h"
 #include "secret_handshake.h"
+#include "test-framework/unity.h"
+#include "test-framework/unity_internals.h"
```
ryanplusplus, line 3: I don't think this is necessary

ryanplusplus (approved):

ryanplusplus:
Thanks!

borderite:
@ryanplusplus I have just realized that I forgot to remove lines 7 - 8 in test_secret_handbreake.c. Can you delete the lines?

ryanplusplus:
> @ryanplusplus I have just realized that I forgot to remove lines 7 - 8 in test_secret_handbreake.c. Can you delete the lines?

Actually it will be easier if you create another PR since I can't approve my own PR.

borderite:
I see. I've submitted the change.

PR #1081, opened 2025-08-14
Pacman Rules concept exercise

colinleach:
Draft for now, until there is a policy decision about exemplar/example and header files.

This is a pretty straight ripoff of the C++ version, with a couple of exceptions.

The test file is, of course, new. This is my first experience of Unity, so I'l wait for your comments on what I did wrong.

The `introduction.md` starts with Bob's version in the Booleans concept, but I removed the section on comparison operators: not needed here, and will be covered in the Conditionals concept. I also added a bit on operator precedence from C++, but I'm not sure if it's needed.

This passes `configlet lint` and `configlet fmt`, but I still need to configure my fork to enable manual CI runs (something about `workflow_dispatch` that I need to read up on). CI is going to fail in any case.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1172,6 +1172,16 @@
           "control_flow_loops"
         ]
       }
+    ],
+    "concept": [
+      {
+        "slug": "pacman-rules",
+        "name": "Pacman Rules",
+        "uuid": "9f14f611-e783-49a5-904e-6f8f1c4d1f51",
+        "concepts": [],
+        "prerequisites": [],
+        "status": "wip"
+      }
     ]
   },
   "concepts": [
--- a/exercises/concept/pacman-rules/.docs/hints.md
+++ b/exercises/concept/pacman-rules/.docs/hints.md
@@ -1,0 +1,27 @@
+# Hints
+
+## General
+
+- Don't worry about how the arguments are derived, just focus on combining the arguments to return the intended result.
+
+## 1. Define if pac-man can eat a ghost
+
+- The function must return a boolean value.
+- You can use the logical operator and ([`&&`][logical operators]) to combine the arguments for a result.
+
+## 2. Define if pac-man scores
+
+- The function must return a boolean value.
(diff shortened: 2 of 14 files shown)


colinleach:
While I go away and learn about `clang-format` (overdue learning, apparently), I'll just note that `format.sh` only operates on practice exercises.

colinleach:
Now I'm confused. I let `clang-format -i <filename>` make a bunch of changes lcally, pushed them, and I'm still getting lots of errors. Any suggestions?

ryanplusplus:
You should be able to have GitHub do it for you by leaving a comment with this in the body:

/format

See:
https://github.com/exercism/c/blob/37a5a6e81ef008b6f76491de34a4d7cd6375a0bd/.github/workflows/format-code.yml#L11

ryanplusplus:
> You should be able to have GitHub do it for you by leaving a comment with this in the body:
> 
> /format
> 
> See:
> 
> https://github.com/exercism/c/blob/37a5a6e81ef008b6f76491de34a4d7cd6375a0bd/.github/workflows/format-code.yml#L11

Looks like this may not work for forks

ryanplusplus:
> Looks like this may not work for forks

For now just ignore it and I can format it on my Linux machine prior to merge.

colinleach:
Thanks. I'm obviously not doing everything right on my Linux machine, so it would be good to understand this at some later point.

ryanplusplus:
> Thanks. I'm obviously not doing everything right on my Linux machine, so it would be good to understand this at some later point.

More than likely it's a small difference related to the version of `clang-format`. Nothing is ever easy 😞

colinleach:
Better!

I installed the `clang-format` extension in VS Code, and that seemed to fix the problems.

colinleach:
> Nothing is ever easy 😞

Could be worse. I first ran a C compiler from 5¼" floppy disks (8086 processor, 640k RAM). That puts 2025 in perspective!

PR #1084, opened 2025-08-16
Draft concept: Headers and Includes

colinleach:
The Lasagna exercise surfaced some concerns in the team about when/how to introduce header files. I've drafted one suggestion for discussion. I aimed for it to be relatively short and gentle, suitable for quite early in the syllabus.

### DIFF
--- a/concepts/headers-includes/about.md
+++ b/concepts/headers-includes/about.md
@@ -1,0 +1,92 @@
+# About
+
+C requires that all functions are declared _before_ they are defined.
+This can be done at the top of the `*.c` file containing the definition, but there are advantages in splitting declarations into a separate header file with a `.h` extension.
+
+You can think of the header file as an API.
+The header will tell you _what_ a codebase has to offer, without going into the details of _how_.
+
+## Header files
+
+A header file can range from just a few lines to several hundred.
+
+For a simple example, we can look at the `Hello World` exercise (which you already completed). This is the [`hello_world.h` file][hello-h] which is part of that exercise:
+
+```c
+// This is called an include guard, which ensures that the header is only
+// included once. You could alternatively use '#pragma once'. See
+// https://en.wikipedia.org/wiki/Include_guard.
+#ifndef HELLO_WORLD_H
+#define HELLO_WORLD_H
+
+// Declare the 'hello()' function, which takes no arguments and returns a
+// 'const char *', i.e. a pointer to a character (in this case the first
+// character in a string). The function itself is defined in the hello_world.c
+// source file. This function is called by the test case(s) in the test source
+// file test_hello_world.c.
+const char *hello(void);
+
+#endif
+```
+
+Most of it is explanatory comments (lines starting with `//`).
+
+### Declarations
+
+Towards the bottom is the _declaration_ of the `hello()` function.
+This is the same as in the _definition_ in `hello_world.c`, except that the declaration ends with a semicolon `;` instead of the function body.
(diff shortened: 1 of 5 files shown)


siebenschlaefer:
Hi Colin, are you open to feedback?

colinleach:
Sure, I'm happy with feedback. Though a bit surprised - this syllabus didn't seem to be progressing much, so I assumed there was limited interest.

Perhaps the bigger problem is that a variety of health problems over the past couple of months have meant I've been very unproductive. There's a big backlog of Julia and Kotlin stuff I should have been working on. Annoying...

concepts/headers-includes/about.md
```diff
@@ -0,0 +1,92 @@
+# About
+
+C requires that all functions are declared _before_ they are defined.
```
siebenschlaefer: That's not true. It requires functions to be declared before they are *called* or otherwise used in an expression.  
There is however the compiler option `-Wmissing-declarations` that warns about non-`static`(!) functions that are defined without being previously declared. The idea is that there are typically two kinds of functions, those declared in a `.h` file and meant to be called from other `.c` files, and local (`static`) helper functions that are effectively hidden from other translation units. The latter can be defined without a prior declaration and `-Wmissing-declarations` will not complain.

How about "called" instead of "defined"?

concepts/headers-includes/about.md
```diff
@@ -0,0 +1,92 @@
...
+This can be done at the top of the `*.c` file containing the definition, but there are advantages in splitting declarations into a separate header file with a `.h` extension.
+
+You can think of the header file as an API.
```
siebenschlaefer: I'd prefer "interface" instead of "API" because I believe it's easier to understand for people who don't know what an "API" is. But that's not a strong preference.

concepts/headers-includes/about.md
```diff
@@ -0,0 +1,92 @@
...
+The header will tell you _what_ a codebase has to offer, without going into the details of _how_.
+
+## Header files
+
+A header file can range from just a few lines to several hundred.
+
+For a simple example, we can look at the `Hello World` exercise (which you already completed). This is the [`hello_world.h` file][hello-h] which is part of that exercise:
```
siebenschlaefer: one line per sentence

concepts/headers-includes/about.md
```diff
@@ -0,0 +1,92 @@
...
+// source file. This function is called by the test case(s) in the test source
+// file test_hello_world.c.
+const char *hello(void);
+
+#endif
+```
+
+Most of it is explanatory comments (lines starting with `//`).
+
+### Declarations
```
siebenschlaefer: I'm not sure about this whole section. It talks about parameters but not about the name or return type.
If this is the main explanation of function declarations I would like it to be more extensive.
But is this really the place for that? I'd prefer a separate concept for functions that contains function declarations.

concepts/headers-includes/about.md
```diff
@@ -0,0 +1,92 @@
...
+
+Towards the bottom is the _declaration_ of the `hello()` function.
+This is the same as in the _definition_ in `hello_world.c`, except that the declaration ends with a semicolon `;` instead of the function body.
+
+This function takes no parameters (indicated by [`void`][void]).
+
+Most functions have parameters, so these:
```
siebenschlaefer: IMHO "these" is ambiguous here. How about "their declarations"?

PR #1086, opened 2025-08-19
Use snprintf instead of sprintf

ahans:
`sprintf` has been deprecated on macOS and a warning is issued when using it. `-Werror` turns the warning into an error, consequently tests don't pass on macOS. This fixes this by using `snprintf` instead.

### DIFF
--- a/exercises/practice/beer-song/.meta/example.c
+++ b/exercises/practice/beer-song/.meta/example.c
@@ -2,29 +2,33 @@
 #include <stdint.h>
 #include <stdio.h>
 
+static const size_t MAX_LINE_LENGTH = 1024;
+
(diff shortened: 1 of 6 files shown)


ryanplusplus (approved):
Looks good once CI is happy

ahans:
> Looks good once CI is happy

Not sure why this works on macOS, it didn't make much sense. One of the rare cases where I tried having Copilot do all the editing. It figured out which files needed some modificiations and also interpreted the error message correctly (we need to replace `sprintf` with `snprintf`), but it did a terrible job at figuring out the actual max lengths. Doing it myself from the start would have been quicker ...

PR #1087, opened 2025-08-20
Add macos-15 CI job

ahans:
gcc on arm64 macOS does not support sanitizers yet, so we only build with Apple Clang there.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,11 +8,15 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-24.04
-
     strategy:
       matrix:
+        os: [ubuntu-24.04, macos-15]
         compiler: [gcc, clang]
+        exclude:
+          - os: macos-15
+            compiler: gcc
+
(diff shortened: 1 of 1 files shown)


ryanplusplus:
On macOS `gcc` is usually an alias for `clang` -- do you know if `gcc` is really `gcc` on macOS runners?

ahans:
> On macOS `gcc` is usually an alias for `clang` -- do you know if `gcc` is really `gcc` on macOS runners?

You're right, it is not. For gcc there are aliases such as `gcc-12`, see [here](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md#language-and-runtime). So things would probably work with `gcc` as well, but would do the exact same thing as the clang version. That seems wasteful. Configuring this via the matrix doesn't seem to work. I think I will just add an `if` to skip the macOS/gcc configuration.

ahans:
Turns out matrix has an `exclude` key. I'm happy with this now.

ryanplusplus:
Thanks!

PR #1154, opened 2025-10-04
WIP: Retry link checking

wolf99:
I saw another repo that similary uses lychee had some fails around the same time but since started working again.
Using this PR to check if the issue remains or is resolved in the lychee action.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -38,7 +38,7 @@ jobs:
         run: ./bin/check-include-guards ./exercises/concept ./exercises/practice
         
       - name: Check links
-        uses: lycheeverse/lychee-action@82202e5e9c2f4ef1a55a3d02563e1cb6041e5332 # 2.4.1
+        uses: lycheeverse/lychee-action@885c65f3dc543b57c898c8099f4e08c8afd178a2
         id: lychee
         with:
           args: >- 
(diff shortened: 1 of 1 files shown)


ryanplusplus:
I took a quick look and the failing links may actually be down

wolf99:
@ryanplusplus looked like some temporary issue with gnu.org and drdobbs is actually gone.
Trying if some things for the the gnu.org errors for now

wolf99:
So --insecure fixed the certificate errors.
Not yet sure if the issue is with the certificate on gnu.org for this useragent, or if there is some issue with the worker env that causes the certificate to, for example, appear to be out of date.

wolf99:
Using the info here to help debug the issues: https://lychee.cli.rs/troubleshooting/network-errors/

Looks like the 403 from stackoverflow is due to that site using Cloudflare Bot Management.
Per: https://builtwith.com/detailed/stackoverflow.com.
So we may just have to ignore stackoverflow URLs :/

wolf99:
Trying some of the suggestions at: https://lychee.cli.rs/troubleshooting/rate-limits/#_top

Looks  like default concurrent network requests is 128.

wolf99:
Still not sure if I'm using the correct token.
@ErikSchierboom do you know if there is a token with this name configured for Exercism? `secrets.GITHUB_TOKEN`

ErikSchierboom:
`${{ secrets.GITHUB_TOKEN }}` is the right token

PR #1465, opened 2026-02-03
Update description to match tested properties in nucleotide task

dadarakt:
I found that the description was misleading, as it implied a specific structure which was not the one tested for.

This PR simply updates the description to match these tests, hope it helps :)


wolf99:
Hi @dadarakt ,

Thank you for the suggestion and PR!
However this file is specified by the centralised `problem-specifications` repository.

You can see the relevant file here: https://github.com/exercism/problem-specifications/blob/main/exercises%2Fnucleotide-count%2Fdescription.md

Can you raise this for discussion on the forums and then, if agreed there, raise the PR on that file?

PR #1468, opened 2026-02-18
Changed wrong/confusing prime factors instructions

ananaft:
Issue: The current instructions consider 4 a possible prime factor, despite it not being a prime number.

Fix: Remove lines mentioning 4 as a possible factor.

### DIFF
--- a/exercises/practice/prime-factors/.docs/instructions.md
+++ b/exercises/practice/prime-factors/.docs/instructions.md
@@ -17,8 +17,6 @@ What are the prime factors of 60?
     So let's move on to our next divisor, 3.
 - 3 goes cleanly into 15, leaving 5.
   - 3 does not go cleanly into 5.
-    The next possible factor is 4.
-  - 4 does not go cleanly into 5.
     The next possible factor is 5.
 - 5 does go cleanly into 5.
 - We're left only with 1, so now, we're done.


wolf99:
I think perhaps you have misunderstood. This sentence is explaining that 4 is a possible factor of 60 but that it is _not_ a prime number so 4 should be discounted.

ananaft:
The current wording makes it seem like the exclusion criterion for the number 4 is the fact that it does not go cleanly into 5. This at least made me double check the exact definition of prime factors elsewhere before continuing with the exercise.

ananaft:
I now realize that the current wording serves as a good hint on how to possibly design the solution algorithm. Checking whether a possible factor is actually a prime number would be redundant. I sincerely apologize for being nitpicky and possibly bothersome. Thank you for your time, nonetheless!

siebenschlaefer:
@ananaft No worries.  

BTW the description and tests cases of the practice exercises come from the language-agnostic [problem-specifications repository](https://github.com/exercism/problem-specifications/).  

If you want to report an error or suggest an improvement that is not specific to the C track please post on the [forum](https://forum.exercism.org/c/exercism/bugs-and-features/126) first before creating an issue or pull request. That's where the maintainers want to keep these discussions.

ananaft:
Thanks for the information!

PR #1470, opened 2026-02-21
Add missing function declarations to resistor_color.h

### LINKED ISSUE #1095: Issue in resistor-color problem
I think the definitions of colors() and color_code() are missing in the resistor_colors.h file.
It is really confusing to see it in local machine that all you have to implement is an Enum and then from tests file find out there are more  that to solve it.
My experience, at the first glance i thought it was my internet connection and started printing the requests and responses until i found out the header file was not complete at all.
It would be awesome if you fixed it for future experiences.

user1-prajwal:
The header file currently contains an empty enum definition,
which may be confusing for contributors working locally.

Fixes #1095

This change adds:
- Enum values for resistor colors
- Function declarations for color_code() and colors()

This improves clarity and aligns the header with the expected API.

### DIFF
--- a/exercises/practice/resistor-color/resistor_color.h
+++ b/exercises/practice/resistor-color/resistor_color.h
@@ -2,7 +2,19 @@
 #define RESISTOR_COLOR_H
 
 typedef enum {
-
+    BLACK = 0,
+    BROWN,
+    RED,
+    ORANGE,
+    YELLOW,
+    GREEN,
+    BLUE,
+    VIOLET,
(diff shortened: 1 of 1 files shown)


ryanplusplus:
Hey @user1-prajwal, the intent is for the user to determine what should be in the header as part of the exercise.

user1-prajwal:
Thanks for the Clarification! I misunderstood the intent of the exercise earlier. I appreciate you taking the time to explain it.

